PR #12, opened 2018-01-17
Add wallet selector static components

banciur:


app/components/walletSelector/WalletLedger.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import { WalletLedgerChooser } from "./WalletLedgerChooser";
+import { WalletLedgerInitComponent } from "./WalletLedgerInitComponent";
+
+export const WalletLedger = () => {
+  // It's just mock-up
+  const check = true;
+
+  if (check) {
+    return <WalletLedgerChooser />;
```
krzkaczor: This component should be called `WalletLedgerAddressChooser`

app/components/walletSelector/WalletLedgerChooserComponent.module.scss
```diff
@@ -0,0 +1,73 @@
+@import "../../styles/neufund-theme";
+
+$blue-hover: rgb(224, 242, 255);
+$blue-accent: #d9edf7;
+
+.chooser-table {
+  margin: .9375rem 0;
```
krzkaczor: as discussed: I wouldnt use hardcoded values like these but use `mt-1` etc classes from bootstrap or use bootstrap variables directly like: `margin: $spacer * 1`. The same apples to padding. For border width please create new variable like `$border-width-min: .0625rem`. We can merge it and fix it in separate PR

PR #14, opened 2018-01-19
Refactor and add tests for wallet selector

banciur:


app/components/walletSelector/WalletLedgerChooserComponent.tsx
```diff
@@ -7,9 +7,42 @@ import { ILedgerAccount } from "../../../typings/typings";
...
+
+  handleClick = () => {
+    this.props.handleAddressChosen(this.props.ledgerAccount);
+  };
+
+  render(): React.ReactNode {
+    return (
+      <tr
+        onClick={this.handleClick}
+        className={cn(styles.useColumn, {
```
krzkaczor: could you change `useColumn` to some more meaningful name? like i dunno: 'addressRow' or whatever

test/components/walletSelector/WalletLedgerChooserComponent.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+  };
+
+  it("should render LoadingIndicator for loading attribute", () => {
+    const props = {
+      ...defaultProps,
+      loading: true,
+    };
+    const component = shallow(<WalletLedgerChooserComponent {...props} />);
+
+    expect(component.contains(<LoadingIndicator />)).to.be.true;
```
krzkaczor: there is a missing test that would assert that these accounts are rendered

test/components/walletSelector/WalletLedgerChooserComponent.spec.tsx
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  it("should render all provided accounts", () => {
+    const component = shallow(<WalletLedgerChooserComponent {...defaultProps} />);
+    expect(component.find(AccountRow).length).to.be.eq(defaultProps.accounts.length);
+  });
+
+  it("should show correct prev / next buttons regarding hasPreviousAddress property ", () => {
+    const componentWithPrevAddr = shallow(<WalletLedgerChooserComponent {...defaultProps} />);
```
krzkaczor: split it into two test cases please

test/components/walletSelector/WalletLedgerChooserComponent.spec.tsx
```diff
@@ -0,0 +1,103 @@
...
+  it("should call correct click handlers for prev / next buttons", () => {
+    const component = shallow(<WalletLedgerChooserComponent {...defaultProps} />);
+
+    component.find(tid("btn-previous")).simulate("click");
+    component.find(tid("btn-next")).simulate("click");
+    expect(defaultProps.showPrevAddresses).to.be.calledOnce;
+    expect(defaultProps.showNextAddresses).to.be.calledOnce;
+  });
+
+  it("<AccountRow /> should render correct account data and handle click ", () => {
```
krzkaczor: plase add describe block for `<AccountRow />`

test/components/walletSelector/WalletLedgerChooserComponent.spec.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const componentWithoutPrevAddr = shallow(<WalletLedgerChooserComponent {...props} />);
+
+    expect(componentWithPrevAddr.find(tid("btn-previous")).length).to.be.eq(1);
+    expect(componentWithPrevAddr.find(tid("btn-next")).length).to.be.eq(1);
+
+    expect(componentWithoutPrevAddr.find(tid("btn-previous")).length).to.be.eq(0);
+    expect(componentWithoutPrevAddr.find(tid("btn-next")).length).to.be.eq(1);
+  });
+
+  it("should call correct click handlers for prev / next buttons", () => {
```
krzkaczor: i would also split it in two separate test cases like: `it should call prev button callback on click ` 
``it should call next button callback on click `

test/components/walletSelector/WalletBrowser.spec.tsx
```diff
@@ -2,6 +2,7 @@ import { expect } from "chai";
 import { shallow } from "enzyme";
 import * as React from "react";
 import { Alert } from "reactstrap";
+
```
Moejoe90: Is this a style rule absolute imports and relative are separated with a line?

banciur: In commit.neufund.org we tried to separate library imports from project imports by new line. There is no linter rule for it so I'm not sure should we follow this pattern. 
@krzkaczor

krzkaczor: Yeah, there is no rule for it but I try to fix it manually when I see it because its cleaner to split them. I couldn't find a ready rule to enforce it so it's not obligatory for us to follow. In future, i hope we will write rule for it (with autofixer of course :D)

PR #15, opened 2018-01-21
Adds responsive header

Moejoe90:
Large
big screen
Medium
medium
Small
mobile


app/components/Header.tsx
```diff
@@ -1,26 +1,182 @@
...
+export class Header extends React.Component<IHeaderProps, IHeaderState> {
+  constructor(props: IHeaderProps) {
+    super(props);
+
+    this.state = {
+      isOpen: false,
+    };
+  }
+
+  toggle(): any {
```
Moejoe90: How would you deal with this situation what would type this?

app/components/Header.tsx
```diff
@@ -1,26 +1,182 @@
...
+    this.setState({
+      isOpen: !this.state.isOpen,
+    });
+  }
+
+  render(): any {
```
Moejoe90: same here

app/components/Header.tsx
```diff
@@ -1,26 +1,182 @@
...
+  name: string;
+  toggle: () => {};
+  isOpen: boolean;
+}
+
+export const NeufundBrand: React.SFC<INeufundBrand> = ({ classStyle }) => {
+  return (
+    <Row>
+      <img
+        src="https://neufund.org/img/social_logo.png"
```
Moejoe90: I took logo as a link for now until an assets folder is set, I know there is another branch with assets, so I didn't want to create conflict

app/components/Header.tsx
krzkaczor: it does not return anything so - `void`

app/components/Header.tsx
krzkaczor: You can read it from class that you extending - React.component (just click on it and your IDE should navigate) `React.ReactNode` is correct type

app/components/Header.tsx
```diff
@@ -1,26 +1,180 @@
...
-    </NavLink>
-
-    <NavLink exact to="/dashboard" className={styles.link} activeClassName={styles.active}>
-      Dashboard
-    </NavLink>
-  </Navbar>
-);
+type IHeaderProps = IunAuthProps | IAuthProps;
+
+interface IunAuthProps {
```
krzkaczor: IUnAuthProps

app/components/Header.tsx
```diff
@@ -1,26 +1,180 @@
...
+  name: string;
+  balanceEuro: number;
+  balanceNeu: number;
+}
+
+interface IHeaderState {
+  isOpen: boolean;
+}
+
+interface INeufundBrand {
```
krzkaczor: please delete this interface and use className which is part of props for all components

app/components/Header.tsx
```diff
@@ -1,26 +1,180 @@
...
+  balanceNeu: number;
+  name: string;
+  toggle: () => {};
+  isOpen: boolean;
+}
+
+export const NeufundBrand: React.SFC<INeufundBrand> = ({ classStyle }) => {
+  return (
+    <Row>
+      <img src="https://neufund.org/img/social_logo.png" className={styles.brand} alt="" />
```
krzkaczor: please leave a todo note here that it needs to be taken from assets instead.

Moejoe90: This will be changed when assets are added to repo

test/components/Header.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <Authorized
+          name={name}
+          balanceEuro={balanceEuro}
+          balanceNeu={balanceNeu}
+          toggle={toggle}
+          isOpen={false}
+        />,
+      );
+      component.find(tid("button-toggle")).simulate("click");
+      expect(toggle.calledOnce).to.equal(true);
```
krzkaczor: use `expect(toggle)to.be.calledOnce`

krzkaczor: https://github.com/domenic/sinon-chai

app/components/Header.tsx
Moejoe90: done

app/components/Header.tsx
Moejoe90: done

Moejoe90: className is not part of `React.SFC` so we will need an interface for it, however, i will change the name to `IReducedHTML` what do you think?

app/components/Header.tsx
Moejoe90: done image taken from asset and a `TODO` note is added

test/components/Header.spec.tsx
Moejoe90: changed, It was first implemented as example available in  
http://airbnb.io/enzyme/docs/api/shallow.html

app/components/Header.module.scss
```diff
@@ -1,16 +1,65 @@
...
+    border-bottom: 1.1rem;
+    border-bottom-style: solid;
+    border-bottom-color: $white;
+  }
+}
+
+.brand {
+  width: 3.75rem;
+  height: 1.875rem;
+  border-right: .09rem;
```
krzkaczor: all borders should have minimal width I think. According to https://offroadcode.com/rem-calculator/ it's: `1px = 0.0625rem`

app/components/Header.module.scss
```diff
@@ -1,16 +1,65 @@
...
+}
+
+.active {
+  font-weight: bold;
+}
+
+@include media-breakpoint-up(sm) {
+  .active {
+    padding-bottom: .5rem;
+    border-bottom: 1.1rem;
```
krzkaczor: look: border note

app/components/Header.module.scss
```diff
@@ -1,16 +1,65 @@
...
-.active {
-  font-weight: bold;
+.mini-header {
+  font-size: .75rem;
+  color: $white;
+}
+
+.title {
+  font-size: 1rem;
+  font-weight: 600;
```
krzkaczor: use font-weight predefined variables from bootstrap

app/components/Header.module.scss
```diff
@@ -1,16 +1,65 @@
...
   &:hover {
     color: $white;
+    text-decoration: none;
   }
 }
 
-.active {
-  font-weight: bold;
+.mini-header {
+  font-size: .75rem;
```
krzkaczor: use predefined variables from bootstrap

test/components/Header.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+    it("should render Authorized Navbar", () => {
+      const component = shallow(
+        <Header
+          isAuthorized={true}
+          name={name}
+          balanceEuro={balanceEuro}
+          balanceNeu={balanceNeu}
+        />,
+      );
+      expect(component.find(Authorized)).to.have.length(1);
```
krzkaczor: please assert here something like `<Authorized and all props here/>` so make sure that props are passed correctly

test/components/Header.spec.tsx
```diff
@@ -0,0 +1,45 @@
+import { expect } from "chai";
+import { shallow } from "enzyme";
+import * as React from "react";
+import * as sinon from "sinon";
+import { Authorized, Header, UnAuthorized } from "../../app/components/Header";
+import { tid } from "../testUtils";
+
+describe("<Header />", () => {
+  const name = "moe";
```
krzkaczor: consider them putting into one object called defaultProps as we do it other tests

test/components/Header.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const balanceEuro = 0;
+  const balanceNeu = 0;
+  describe("<Header />", () => {
+    it("should render UnAuthorized Navbar", () => {
+      const component = shallow(<Header isAuthorized={false} />);
+      expect(component.find(UnAuthorized)).to.have.length(1);
```
krzkaczor: please assert here something like <UnAuthorized /> which makes sure that no props were passed to that component

krzkaczor:
image
 it's broken on certain width

krzkaczor:
image
what is the point of having a separate row with a spacer? couldn't it be solved by having margin-top and some border-top?

app/components/Header.module.scss
Moejoe90: How would you style 
border

app/components/Header.module.scss
Moejoe90: good note

app/components/Header.module.scss
krzkaczor: i am sorry — this one's fine. I was referring to borders with minimal width. I think this is what you wanted to have with  `border-right: .09rem;` below

app/components/Header.module.scss
Moejoe90: done, taken out of CSS and used `font-weight-bold` by bootstrap

app/components/Header.module.scss
Moejoe90: Yeah for the one below definitely you on spot

app/components/Header.module.scss
Moejoe90: Changed from `600` to `font-weight-bold` but take into note bootstrap has only four weights I think 
bold `700`
normal `500` 
if we want to implement weight 600 then we will have to manually add it
https://getbootstrap.com/docs/4.0/utilities/text/#font-weight-and-italics

Moejoe90:
@krzkaczor 
The idea was for the separator to act as a different component that you can take off the view and grid if you wanted. I think this makes it simpler to change the spacer in the future to whatever we want. Based on designs without indulging too much into the complexities of defining many CSS styles and `@media` queries. 
If this is not the best approach. Then it can be amended to what you think is best.

Moejoe90:
fixed layout
image

PR #18, opened 2018-01-23
Browser wallet integration

krzkaczor:
- browser wallet integration
- simple dev logger


app/modules/wallet-selector/browser-wizard/reducer.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export const browserWalletWizardInitialState: IBrowserWalletWizardState = {};
+
+export const browserWalletWizardReducer: AppReducer<IBrowserWalletWizardState> = (
+  state = browserWalletWizardInitialState,
+  action,
+): IBrowserWalletWizardState => {
+  switch (action.type) {
+    case "BROWSER_WALLET_CONNECTION_ERROR":
```
Moejoe90: What case would this error happen? Failure to connect with web3

app/modules/web3/BrowserWallet.ts
```diff
@@ -0,0 +1,51 @@
+import { injectable } from "inversify";
+import * as Web3 from "web3";
+import { BrowserWalletSubType, IPersonalWallet, WalletType } from "./PersonalWeb3";
+import { Web3Adapter } from "./Web3Adapter";
+
+export class BrowserWalletError extends Error {}
+export class BrowserWalletMissingError extends BrowserWalletError {}
+export class BrowserWalletMismatchedNetworkError extends BrowserWalletError {
+  constructor(public readonly desiredNetworkId: string, public readonly actualNetworkId: string) {
+    super("MismatchedNetworkError");
```
Moejoe90: what's the idea behind this or the use case?

test/modules/wallet-selector/browser-wallet/actions.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+      expect(browserWalletMock.connect).to.be.calledWithExactly(expectedNetworkId);
+      expect(dispatchMock).to.not.be.called;
+    });
+
+    it("should dispatch error action on error", async () => {
+      const expectedNetworkId = "5";
+
+      const dispatchMock = spy();
+      const navigateToMock = spy();
+      const browserWalletMock = createMock(BrowserWallet, {
```
Moejoe90: So for every test case you generate the mock you need, instead of doing a general mocked model that test cases can use?

app/components/walletSelector/WalletSelector.tsx
```diff
@@ -84,19 +85,40 @@ export const WalletSelectorComponent: React.SFC<IWalletSelectorProps> = ({
...
+  LIGHT_WALLET = "LIGHT_WALLET",
+  WALLET_IN_BROWSER = "WALLET_IN_BROWSER",
+  LEDGER_WALLET = "LEDGER_WALLET",
+}
+
+export class WalletSelector extends React.Component<{}, IWalletSelectorState> {
+  constructor(props: {}) {
+    super(props);
+    this.state = {
+      selectedIndex: WalletSelectorTab.LIGHT_WALLET,
```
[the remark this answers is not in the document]
krzkaczor: for now? i just did it because it's first from the left

app/modules/wallet-selector/browser-wizard/reducer.ts
krzkaczor: yeah so it can happen in multiple cases, like you dont have metamask at all or it's locked etc.

app/modules/web3/BrowserWallet.ts
krzkaczor: actually i think that we should always follow this pattern — if you extend error with custom object make sure that there is a name of this error somehere in the error object. Can be custom property like name or whatever. Otherwise its hard to debug. console.log(JSON.stringify(err)) of such error will return "{}" which makes it hard to debug.

test/modules/wallet-selector/browser-wallet/actions.spec.ts
krzkaczor: Each time they are configured in a different way.

PR #23, opened 2018-01-25
tweak app routes, show authorized header after signin

krzkaczor:
image
image

### DIFF
--- a/app/components/Header.tsx
+++ b/app/components/Header.tsx
@@ -1,6 +1,17 @@
 import * as React from "react";
 import { NavLink } from "react-router-dom";
-import { Col, Collapse, Navbar, NavbarBrand, NavbarToggler, Row } from "reactstrap";
+import {
+  Col,
+  Collapse,
+  Container,
+  Nav,
+  Navbar,
+  NavbarBrand,
+  NavbarToggler,
+  NavItem,
+  Row,
+} from "reactstrap";
+import { appConnect } from "../store";
 import * as styles from "./Header.module.scss";
 
 type IHeaderProps = IUnAuthProps | IAuthProps;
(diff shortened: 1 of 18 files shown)


app/components/Header.tsx
```diff
@@ -154,16 +166,23 @@ export const Authorized: React.SFC<IAuthorized> = ({
...
+            </NavItem>
+          </Nav>
+        </Navbar>
+      </Container>
     </div>
   );
 };
 
-export class Header extends React.Component<IHeaderProps, IHeaderState> {
+export class HeaderComponent extends React.Component<IHeaderProps, IHeaderState> {
```
Moejoe90: Will this be a convention for example `walletLightComponent` ... etc

app/components/Header.tsx
```diff
@@ -192,3 +211,24 @@ export class Header extends React.Component<IHeaderProps, IHeaderState> {
     );
   }
 }
+
+export const Header = appConnect<IHeaderProps>({
+  stateToProps: state => {
+    // relying on web3state connected is temporary
+    if (state.web3State.connected) {
+      return {
+        isAuthorized: true,
```
Moejoe90: from where will you get these values in the future?

app/components/Header.tsx
krzkaczor: Yeah, that was my plan — most important component in given file should be simple named for example: `Wallet` if you have some functional components you can export them as `WalletComponent` to easily write tests.

app/components/Header.tsx
krzkaczor: We will have a state subtree for user/authentication — after successfully obtaining JWT from backend and saving some user-related data. You could have disconnected wallet but still logged in.

app/components/walletSelector/WalletRouter.tsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export const WalletRouter: React.SFC = () => (
+  <Switch>
+    <Route path={walletRoutes.light} component={WalletLight} exact />
+    <Route path={walletRoutes.browser} component={WalletBrowser} exact />
+    <Route path={walletRoutes.ledger} component={WalletLedger} exact />
+    <Redirect to={walletRoutes.light} />
+  </Switch>
+);
```
Moejoe90: I assume lightwallet should have a similar router file structure that takes care of all routes related with light wallet

krzkaczor: Yeah, I would imagine that as well. We could do the same for ledger "wizard" to simplify it. Also, we need to remember about adding enter guards — something like if user is already logged it then it should be redirected from `/login` to `/dashboard` for example. But we can do it later since we dont have all data required.

PR #24, opened 2018-01-28
Fix publicPath to avoid 404 when reloading page on nested routes

krzkaczor:


krzkaczor:
btw. it seems like codcov doesnt work correctly :(

PR #27, opened 2018-01-29
Adds light-wallet Visuals and routes

Moejoe90:
image

mediumlight
lightsmall


package.json
```diff
@@ -86,6 +86,7 @@
     "classnames": "^2.2.5",
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
+    "eth-lightwallet": "^3.0.1",
```
krzkaczor: You dont need this dependency in this PR but I assume that you will use it in the next one, is that correct?

krzkaczor:
https://www.dropbox.com/s/z5skba3gxijs5ma/Screenshot%202018-01-29%2015.11.07.png?dl=0 tabs in wallet selector are not highlighted properly anymore.  Plase use WalletSelector line 36 should use `active={location.pathname.startsWith(href)} ` instead.

app/components/walletSelector/walletLight/RecoverWallet.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <FormGroup>
+            <Label>Password</Label>
+            <Input type="password" name="password" id="recover-password" />
+          </FormGroup>
+          <FormGroup>
+            <Label>Repeat password</Label>
+            <Input type="password" name="repeated" id="recov-repeated-pass" />
+          </FormGroup>
+          <Link className="btn btn-secondary" to={walletLightRoutes.validate}>
+            I saved it
```
krzkaczor: i think that this label should be different. Something like "I want to recover my wallet"

package.json
Moejoe90: Yes, this is correct

app/components/walletSelector/walletLight/RecoverWallet.tsx
Moejoe90: Changed to `Recover my wallet`

PR #29, opened 2018-01-29
Inversify context

krzkaczor:


package.json
```diff
@@ -96,6 +96,7 @@
     "ledger-wallet-provider": "2.0.0-beta.2",
     "lodash": "^4.17.4",
     "material-ui": "^0.20.0",
+    "prop-types": "^15.6.0",
```
Moejoe90: Is there a reason for us to start using PropTypes? i assumed typescript should solve all those problems

krzkaczor: We just need it to make react context work. Whole react context api will be changed in next react major and then we hopefully get rid of them.

PR #44, opened 2018-01-31
Implement working choosing of derivation path in ledger adress selector

banciur:
screenshot


test/modules/wallet-selector/ledger-wizard/actions.spec.ts
```diff
@@ -156,6 +158,21 @@ describe("Wallet selector > Ledger wizard > actions", () => {
...
 
+  describe("setDerivationPathPrefixAction", () => {
+    it("should work", async () => {
+      const newDP = "test";
+      const mockDispatch = spy();
+
+      await setDerivationPathPrefixAction(newDP)(mockDispatch);
+
+      expect(mockDispatch).to.be.calledTwice;
+      expect(mockDispatch).to.be.calledWith(
```
krzkaczor: prefer `calledWithExactly` — it makes sure that no other parameters were passed.

test/modules/wallet-selector/ledger-wizard/actions.spec.ts
```diff
@@ -156,6 +158,21 @@ describe("Wallet selector > Ledger wizard > actions", () => {
...
+        setLedgerWizardDerivationPathPrefix({ derivationPathPrefix: newDP }),
+      );
+      expect(mockDispatch).to.be.calledWith(loadLedgerAccountsAction);
```
krzkaczor: same here

test/utils/Validators.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+    it("should return null for correct prefix path", () => {
+      expect(derivationPathPrefixValidator(correctPrefix)).to.be.null;
+    });
+
+    it("should return string message for incorrect prefix path", () => {
+      for (const dp of incorrectPrefix) {
+        const validationResult = derivationPathPrefixValidator(dp);
+        expect(validationResult).to.be.not.null;
+        expect(validationResult).to.be.a("string");
+        expect((validationResult as string).length).to.be.greaterThan(0);
```
krzkaczor: use `validationResult!` instead of `validationResult as string` to force unwrap optional value

krzkaczor: I am not sure about this assertion. If I am getting you right you're not interested in the error message but rather only in the fact that error was reported (so truthy (not empty string for example here) value was returned) . I think you could merge all these asserts to something like `expect(validationResult).to.be.ok` which does exactly this. But on the other hand, it's less obiovus and works only because type system prohibits us from returning anything different than string or null. So use your best judgement. We could also export all error messages to separate file with errors and then assert that exact error was thrown but avoid having hardcoded string in muplitple places. It's up to you - we can always improve it later.

app/components/walletSelector/WalletLedgerDPChooser.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <Alert color="danger">
+          <span data-test-id="dpChooser-error-msg">{errorMessage}</span>
+        </Alert>
+      )}
+    </Col>
+  </Row>
+);
+
+interface IDPChooserProps {
+  onChange: any;
```
krzkaczor: use something like `Function` or `() => any` or `() => void` instead of `any`. We need to unify it as some point...

test/modules/wallet-selector/ledger-wizard/actions.spec.ts
banciur: fixed

test/modules/wallet-selector/ledger-wizard/actions.spec.ts
banciur: fixed

app/components/walletSelector/WalletLedgerDPChooser.tsx
banciur: fixed

test/utils/Validators.spec.tsx
banciur: As discussed I prefer more verbose cases less relaying on type inference and we will move hard coded validation messages into separate file later

PR #89, opened 2018-02-04
On derivation path prefix error hide accounts

banciur:
This is new feature for existing PR #44. To ease checking I created separate PR that would be merged into `derivation_path_fix` branch

screenshot

### DIFF
--- a/app/components/walletSelector/WalletLedgerChooser.tsx
+++ b/app/components/walletSelector/WalletLedgerChooser.tsx
@@ -5,7 +5,7 @@ import {
   finishSettingUpLedgerConnectorAction,
   goToNextPageAndLoadDataAction,
   goToPreviousPageAndLoadDataAction,
-  handleDerivationPathPrefixErrorAction,
+  ledgerWizardDerivationPathPrefixAction,
   loadLedgerAccountsAction,
   setDerivationPathPrefixAction,
   verifyIfLedgerStillConnected,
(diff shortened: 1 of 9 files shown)


app/modules/wallet-selector/ledger-wizard/actions.ts
```diff
@@ -143,6 +151,13 @@ export const setDerivationPathPrefixAction = (derivationPathPrefix: string) =>
     [DispatchSymbol, GetStateSymbol],
   );
 
+export const handleDerivationPathPrefixErrorAction = injectableFn(
```
krzkaczor: You don't need that. Just call plain actions directly in a dispatch to props like: `dispatch(ledgerWizardDerivationPathPrefixAction())`.

test/modules/wallet-selector/ledger-wizard/actions.spec.ts
```diff
@@ -199,6 +200,15 @@ describe("Wallet selector > Ledger wizard > actions", () => {
     });
   });
 
+  describe("handleDerivationPathPrefixErrorAction", () => {
```
krzkaczor: this would also go

app/components/walletSelector/WalletLedgerChooser.tsx
```diff
@@ -31,6 +32,7 @@ export const WalletLedgerChooser = compose<React.SFC>(
       hasPreviousAddress: selectHasPreviousPage(state.ledgerWizardState),
     }),
     dispatchToProps: dispatch => ({
+      onDerivationPathError: () => dispatch(handleDerivationPathPrefixErrorAction),
```
krzkaczor: here just use plain action creator

PR #90, opened 2018-02-05
Prevent race condition on derivation path selector

banciur:
This is new feature for existing PR #44. To ease checking I created separate PR that would be merged into `derivation_path_fix` branch

### DIFF
--- a/app/modules/wallet-selector/ledger-wizard/reducer.ts
+++ b/app/modules/wallet-selector/ledger-wizard/reducer.ts
@@ -1,7 +1,7 @@
 import { AppReducer } from "../../../store";
 
 export const DEFAULT_DERIVATION_PATH_PREFIX = "44'/60'/0'/";
-const DEFAULT_LEDGER_ACCOUNTS_PER_PAGE = 10;
+export const DEFAULT_LEDGER_ACCOUNTS_PER_PAGE = 10;
 
 export interface ILedgerAccount {
   address: string;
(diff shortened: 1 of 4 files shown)


app/modules/wallet-selector/ledger-wizard/reducer.ts
```diff
@@ -59,12 +59,19 @@ export const ledgerWizardReducer: AppReducer<ILedgerWizardState> = (
...
-        isLoadingAddresses: false,
-      };
+    case "SET_LEDGER_WIZARD_ACCOUNTS": {
+      if (state.derivationPathPrefix === action.payload.derivationPathPrefix) {
+        return {
+          ...state,
+          accounts: action.payload.accounts,
+          isLoadingAddresses: false,
+        };
+      } else {
```
krzkaczor: You don't need that else branch since returning old state should be the default behavior of any reducer.

app/modules/wallet-selector/ledger-wizard/reducer.ts
```diff
@@ -59,12 +59,19 @@ export const ledgerWizardReducer: AppReducer<ILedgerWizardState> = (
...
         isLoadingAddresses: true,
       };
-    case "SET_LEDGER_WIZARD_ACCOUNTS":
-      return {
-        ...state,
-        accounts: action.payload.accounts,
-        isLoadingAddresses: false,
-      };
+    case "SET_LEDGER_WIZARD_ACCOUNTS": {
+      if (state.derivationPathPrefix === action.payload.derivationPathPrefix) {
```
krzkaczor: what do you think about putting comment here or even better, extracting this conditional to const like: 
`const isLatestUpdate` or `const isMatchingDerivationPath` something along this lines i think

PR #91, opened 2018-02-05
Production config

banciur:
Split webpack config file into production and development with shared common file
Add JS minification
Add hash of js bundle to filename
Add favicon

What is missing css styles should be moved to separate file.


webpack/webpack.config.prod.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.exports = merge(configCommon, {
+  output: {
+    filename: "[name].[hash].min.js",
+  },
+  plugins: [
+    new CleanWebpackPlugin(paths.dist, {
+      root: paths.root,
+    }),
+    new ExtractTextPlugin("styles.css"), // TODO: its not working
```
krzkaczor: do you have any insights why it doesnt work? Please create separate issue for it. You can assign it to me.

webpack/webpack.config.prod.js
```diff
@@ -0,0 +1,37 @@
...
+    new UglifyJsPlugin(),
```
krzkaczor: please provide here some additional options like:
```
new UglifyJsPlugin({
      uglifyOptions: {
        ecma: 6,
      },
    }),
```
We want to ship exactly es6 code.

PR #109, opened 2018-02-11
Login layout

banciur:
screencapture-localhost-9090-login-light-create-1518354377297


app/components/LayoutRegisterLogin.module.scss
```diff
@@ -0,0 +1,12 @@
+@import "../styles/neufund-theme";
+
+.layout {
```
krzkaczor: please create separate directory for layouts (`/components/layouts`)

app/components/LayoutRegisterLogin.module.scss
```diff
@@ -0,0 +1,12 @@
...
+  background: url("../assets/img/logo.svg") no-repeat center 2.688rem;
+  padding-top: 13.13rem;
+}
+
+.main-container {
+  background-color: $white;
+  box-shadow: $shadow;
+  padding: 1.25rem;
```
krzkaczor: hmm use spacing value as a baseline as bootstrap does for `md-x` styles

app/components/LayoutRegisterLogin.module.scss
```diff
@@ -0,0 +1,12 @@
+@import "../styles/neufund-theme";
+
+.layout {
+  background: url("../assets/img/logo.svg") no-repeat center 2.688rem;
+  padding-top: 13.13rem;
```
krzkaczor: >hmm use spacing value as a baseline as bootstrap does for md-x styles

I would say to do here the very same thing but just round the number so idk something like $spacing * 6

app/styles/variables.scss
```diff
@@ -3,4 +3,8 @@ $loading-indicator-color: #333;
 $black-lighter: #262b2f;
 $blue-darker: #315568;
 $gray-almost-dark: #7d919b;
+$pale-grey-50: rgba(227, 234, 245, .5);
```
krzkaczor: can we stick with hex numbers?

app/components/layouts/LayoutRegisterLogin.module.scss
```diff
@@ -0,0 +1,11 @@
+@import "../../styles/neufund-theme";
+
+.layout {
+  background: url("../../assets/img/logo.svg") no-repeat center $spacer * 2.5;
```
krzkaczor: you could just make a separate element for logo. Like a div with background

app/styles/variables.scss
```diff
@@ -3,4 +3,8 @@ $loading-indicator-color: #333;
 $black-lighter: #262b2f;
 $blue-darker: #315568;
 $gray-almost-dark: #7d919b;
+$pale-grey-50: #e3eaf5;
 $off-white: #f5f5f5;
+$white: #fff;
+
+$shadow:  0 .125rem .25rem 0 rgba(0, 0, 0, .1);
```
krzkaczor: and color here as hex please

app/components/layouts/LayoutRegisterLogin.module.scss
krzkaczor: but this is just my opinion. let it be

app/styles/variables.scss
banciur: In this case wouldn't it give different outcome as here we are using transparency?

app/components/layouts/LayoutRegisterLogin.module.scss
banciur: I can but I think in this case it would introduce more complication than benefits. To use it in different places we would have to be restyled for every case or we would have to make it configurable through properties. So on the end I will boil to similar combination of html / css.

app/styles/variables.scss
krzkaczor: yeah it's fine then i guess

PR #110, opened 2018-02-11
Add typed data signer

krzkaczor:
- adds typed data signer
- send signer type to backend


app/modules/web3/BrowserWallet.ts
```diff
@@ -1,8 +1,10 @@
 import { promisify } from "bluebird";
+import hex2ascii = require("hex2ascii");
```
banciur: Why are we mixing different styles of import here?

krzkaczor: simply because otherwise it doesnt work (i wrote my own typings for this module). I read about this somewhere on SO but i can't find the thread now :/

PR #112, opened 2018-02-11
Adds local storage interface

Moejoe90:
Implements initial local storage interface


app/modules/storage/storage.ts
```diff
@@ -0,0 +1,12 @@
+export interface ILocalStorage {
```
krzkaczor: you dont use anywhere this interface

app/modules/storage/storage.ts
```diff
@@ -0,0 +1,12 @@
...
+  setKey(key: string, value: string): void;
+  getKey(key: string): string;
+}
+
+export const setKey = (key: string, value: string): void => {
```
krzkaczor: the idea for abstracting this localstorage with our interface is that you create class implementing this interface that accesses real localStorage. So you should put these methods into a class

app/modules/storage/storage.ts
```diff
@@ -0,0 +1,12 @@
...
+  (window as any).localStorage.setItem(key, value);
```
krzkaczor: you can declare once at the top of the file `const windowAny = window as any` and use windowAny everywhere in this file.

test/modules/storage/storage.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { expect } from "chai";
+import * as localStorage from "node-localstorage";
+import { getKey, setKey } from "../../../app/modules/storage/storage";
+
+beforeEach(() => {
+  (window as any).localStorage = new localStorage.LocalStorage("./scratch");
```
krzkaczor: what does `scratch` mean?

Moejoe90: this is just a placeholder for storage when using this library you need to specify where the node file is. But you have a point it should be changed to `./tempStorage` for example

app/modules/storage/storage.ts
Moejoe90: There was a discussion last time regarding this exact point, should we use an interface for storage abstraction or a class. In all cases i will implement a class in this case

krzkaczor: You can have either only class (skip interface) or have a class implementing the interface. Eiter way we need to have a class that implements the whole behavior. I think it's easier to manage one class then injecting separate functions.

test/modules/storage/storage.spec.ts
krzkaczor: hmm, it sucks that it does serialization to disk automatically — we don't need that behavior (it's slower and produces mess on disk). What about using this instead: https://www.npmjs.com/package/localStorage You would need to manually inject it to global object and later remove it but I think it's fine.

test/modules/storage/storage.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+afterEach(() => {
+  (window as any).localStorage._deleteLocation();
+  (window as any).localStorage = undefined;
+});
+
+describe("Browser Storage", () => {
+  it("should store an item and retrieve it from local storage", async () => {
+    const expectedPhrase = "dummyWallet";
+    const key = "lightWallet";
```
krzkaczor: separate please with new lines arrange, act and assert phases of test.

test/modules/storage/storage.spec.ts
Moejoe90: Actually your initial idea of changing to  `localstorage-mock` is the best way to go lightweight implements it very simply on memory if your fine with that i would go with ` localstorage-mock`

krzkaczor: sure. go ahead ;)

test/modules/storage/storage.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    const localStorageMock = {
+      setItem: spy(),
+      getItem: spy(() => expectedPhrase),
+    };
+
+    const storage = new Storage(localStorageMock);
+    storage.setKey(key, expectedPhrase);
```
krzkaczor: you could actually split it into 2 separate test cases since theese calls are not related in any way (they are mocked anyway).

Moejoe90: you have a point I will do that asap and ping you for a final review

PR #113, opened 2018-02-11
Add mock APIs for lightwallet

Moejoe90:
Adds mock VaultApI
Adds mock UsersApi


app/modules/networking/UsersApi.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export interface IStoreEndpointResponse {
+  vault: string;
+}
+
+export const SignatureAuthApiSymbol = "UsersApiSymbol";
+
+@injectable()
+export class UsersApi {
+  // tslint:disable-next-line
```
krzkaczor: why tslint:disable is needed here?

Moejoe90: Because currently `JsonHttpClientSymbol` is declared but not used anywhere I would suggest we keep it like this so that the structure is very clear on the structure of the mock abstraction

PR #123, opened 2018-02-13
implement new design for browser wallet

krzkaczor:
image
image


app/components/HorizontalLine.tsx
```diff
@@ -0,0 +1,12 @@
+import * as cn from "classnames";
```
banciur: Maybe here we should just restyle `<hr />` element? and put it as global style

app/components/walletSelector/WalletBrowser.module.scss
```diff
@@ -9,3 +9,9 @@
   margin-top: .9375rem;
```
banciur: we could change it to `$spacer`

app/components/walletSelector/WalletBrowser.module.scss
```diff
@@ -9,3 +9,9 @@
...
   margin-bottom: .9375rem;
 }
+
+.wallet-logos {
+  img {
+    height: 85px;
```
banciur: Aren't we are using `rem` units?

app/styles/variables.scss
```diff
@@ -8,7 +8,16 @@ $gray-almost-dark: #7d919b;
...
 $white: #fff;
+$bright-orange: #ff6d00;
+
 
 $smaller-font-size: .8125rem;
 
 $shadow:  0 .125rem .25rem 0 rgba(0, 0, 0, .1);
+
+
+$headings-font-family: "Lato", sans-serif;
```
banciur: If those are bootstrap variable overrides I would move them to `neufund-theme.scss` where already are some overrides. Or if this place is better I would move ones from there here,

app/components/walletSelector/WalletBrowser.module.scss
banciur: Or use (spacing utils)[https://getbootstrap.com/docs/4.0/utilities/spacing/] sadly those are not implemented in react-strap

app/components/HorizontalLine.tsx
krzkaczor: yeah, i thought about this but it seems like styling HR is not so straight forward.

app/styles/variables.scss
krzkaczor: I thought that `variables.scss` is the file that we put you know, all variables :D I hope we won't have many other variables then bootstrap overrides.

PR #124, opened 2018-02-13
WalletLedgerInitComponent styling

banciur:
I'm creating this early PR as I changed base font size that might affect many dimensions so we can work on same base.

Next one will be ledger address chooser

screencapture-localhost-9090-login-ledger-1518547413052
screencapture-localhost-9090-login-ledger-1518547444055


typings/modules.d.ts
```diff
@@ -8,3 +8,7 @@ declare module "web3-provider-engine/subproviders/hooked-wallet";
 
 declare module "ledger-wallet-provider";
 declare module "eth-lightwallet";
+
+declare module "*.png";
```
Moejoe90: Is this the best way to declare extentions

krzkaczor: Yes, this is just to trick TSC into compiling the code but it should be moved to file called `files.d.ts`. Where actually png is already defined...

app/components/walletSelector/WalletLedgerInitComponent.tsx
```diff
@@ -1,54 +1,75 @@
...
+      <Col>
+        <h1 className="text-center">Logging in with Nano Ledger</h1>
+      </Col>
+    </Row>
     {errorMessage && (
-      <Alert color="info">
-        <h4>Connection status:</h4>
-        <p data-test-id="ledger-wallet-error-msg">{errorMessage}</p>
-      </Alert>
+      <WarningAlert className="my-4">
```
krzkaczor: this should be centered

krzkaczor: what does `my-4` mean? is it typo?

app/components/walletSelector/WalletLedgerInitComponent.tsx
```diff
@@ -1,54 +1,75 @@
...
-        </li>
-      </ul>
-    </div>
+  <>
+    <Row>
+      <Col>
+        <h1 className="text-center">Logging in with Nano Ledger</h1>
+      </Col>
+    </Row>
     {errorMessage && (
```
krzkaczor: in separate PR i will get rid of this and always show some error message

typings/modules.d.ts
banciur: This is webpack magic. Then it treats file as module and you can import it in code.

app/components/walletSelector/WalletLedgerInitComponent.tsx
```diff
@@ -1,54 +1,75 @@
...
+        desc="Use left and right key to enter numbers and press 2 keys at the same time to confirm the code"
+      />
+      <InitStep
+        header="3. Pick Ethereum"
+        img={imgStep3}
+        desc="Click on arrows to scroll  apps and pick Ethereum icon. Press 2 keys at the same time to confirm"
+      />
+    </Row>
+    <Row className="mt-5">
+      <Col className={cn("text-center text-md-right", styles.contact)}>
```
krzkaczor: this should be close to the bottom border. Maybe you need to change padding settings

app/components/walletSelector/WalletLedgerInitComponent.tsx
banciur: It's spacing class from (bootstrap)[https://getbootstrap.com/docs/4.0/utilities/spacing/] `y - for classes that set both *-top and *-bottom`

app/components/walletSelector/WalletLedgerInitComponent.tsx
banciur: fixed

app/components/walletSelector/WalletLedgerInitComponent.tsx
banciur: fixed

typings/modules.d.ts
banciur: I removed those

app/components/walletSelector/WalletLedgerInitComponent.tsx
krzkaczor: this is neat i didn't know about this :D

app/styles/variables.scss
```diff
@@ -15,6 +15,11 @@ $smaller-font-size: .8125rem;
 
 $shadow:  0 .125rem .25rem 0 rgba(0, 0, 0, .1);
 
+// Bootstrap variable overrides
+$body-color: $denim;
+$body-bg: $pale-grey-50;
+$font-family-base: "Open Sans", sans-serif;
+$font-size-base: .9375rem !default;
```
krzkaczor: !default here is not needed

banciur: This is how bootstrap does it in their source code. - https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L235

PR #140, opened 2018-02-14
[wip] D#/kyc implementation

huesforalice:
I have done a "skeleton implementation" of the kyc screens, it feature are new, action-based way of routing and sagas. If you like we can aim to merge this to master, but for now I'm just looking for comments. If there is anything unclear, please let me know.


app/modules/kyc/sagas.ts
```diff
@@ -0,0 +1,75 @@
...
+    // simulate some kind of request
+    yield delay(1000);
+    yield effects.put(actions.goToKYCManualVerificationIDUpload());
+  }
+}
+
+function* uploadID(): Iterator<any> {
+  while (true) {
+    const action: TAction = yield neuTake("KYC_UPLOAD_ID");
+    if (action.type !== "KYC_UPLOAD_ID") {
```
krzkaczor: is this only to trick the compiler into narrowing the action type?

huesforalice: Yeah, in this case the type guard is not needed, as I'm doing nothing with the action, but in real life the action (thus the payload) would be fully types.

PR #159, opened 2018-02-16
D#/fix namings

huesforalice:


app/components/shared/progressStepper/ProgressStepper.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from "react";
 
 import * as cn from "classnames";
 import { times } from "lodash";
-import * as styles from "./styles.module.scss";
+import * as styles from "./ProgressStepper.module.scss";
```
krzkaczor: i thought you will delete progressStepper directory entirely and we put all components in long list in shared/

huesforalice: ah ok, also good, will do

PR #163, opened 2018-02-17
D#/refactor actions

krzkaczor:


app/modules/flows.ts
```diff
@@ -0,0 +1,5 @@
+import { walletFlows } from "./wallet-selector/flows";
+
+export const flows = {
```
krzkaczor: i really don't see any benefit in putting all flows in objects and then assembling it. First of all, it breaks with circular dependencies which happened a lot in test cases and I had bad time solving them. Second of all it's just not needed when every flow can be just directly exported. What do you think @huesforalice ?

krzkaczor: walletFlows

PR #189, opened 2018-02-21
Adds Password Reset Component

Moejoe90:
image

### DIFF
--- a/app/components/AppRouter.tsx
+++ b/app/components/AppRouter.tsx
@@ -12,7 +12,7 @@ import { Redirect, Route, Switch } from "react-router-dom";
 import { Dashboard } from "./Dashboard";
 import { Home } from "./Home";
 import { Kyc } from "./kyc/Kyc";
-import { WalletBrowserComponent } from "./WalletLoginHelp";
+import { WalletLoginHelp } from "./WalletLoginHelp";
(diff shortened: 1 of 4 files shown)


app/components/WalletResetHeader.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
+import { Col } from "reactstrap";
```
krzkaczor: this file should be placed in more specific place. With other help related components i guess?

app/components/WalletResetHeader.module.scss
```diff
@@ -0,0 +1,5 @@
+@import "../styles/neufund-theme";
+
+.wallet-reset-layout {
+  font-weight: 500;
```
krzkaczor: use variable here

app/components/WalletResetHeader.tsx
Moejoe90: Shared?

app/components/WalletResetHeader.module.scss
Moejoe90: Done

app/components/WalletResetHeader.tsx
banciur: I would put it in `app/components/walletSelector/walletLight` as it will be used only in context of light wallet

PR #191, opened 2018-02-21
Signing flow

krzkaczor:
What works:
- full registration flow for all types of wallets
- message signing popup after registration
- rebuilding wallet state on request (when you need to sign a message) works with ledger and browser wallets

What's missing (and coming in next prs) :
- more tests
- signing message flow with light wallet


app/components/signing/MessageSignPrompt.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from "react";
```
banciur: Name of file is misleading as its not generic component but one that is used to confirm you identity.
So maybe JWTSignPrompt.tsx or AccountConfirmationPrompt.tsx or something similar

app/components/walletSelector/WalletSelector.tsx
```diff
@@ -3,50 +3,76 @@ import * as React from "react";
...
-        </NavLink>
-      </Col>
-    </Row>
-    <Row>
-      <Col>
-        <WalletRouter />
-      </Col>
-    </Row>
+    {isMessageSigning ? (
+      <WalletMessageSigner />
```
banciur: As we spoke use routing not common component.

app/lib/web3/Web3Manager.ts
```diff
@@ -76,6 +76,14 @@ export class Web3Manager {
     this.web3ConnectionWatcher.start();
   }
 
+  public async sign(message: string): Promise<string> {
```
banciur: This method should not be part `Web3Manager` and only in personal wallets

PR #196, opened 2018-02-22
updates internal kyc api endpoints and moves interfaces and shape def…

huesforalice:
…initions for kyc into one file


huesforalice:
@krzkaczor: yeah, if the server sends us data we do not expect, we need a generic error for that, all the user needs to know in that case is, that there is some kind of problem. But he can't solve it anyway, so there is no use in tell him the server send the wrong shape :)

huesforalice:
As soon as your httpclient changes are merged, I make sure this is taken care of

PR #197, opened 2018-02-23
adds toplevel tests for all actions

huesforalice:


app/modules/actions.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+const ALLOWED_DUPLICATES: { [type: string]: boolean } = {
+  "@@router/CALL_HISTORY_METHOD": true,
+};
+
+describe("modules", () => {
+  describe("all actions", () => {
+    it("should have unique types", () => {
+      const type_map: { [key: string]: boolean } = {};
+
+      forEach(allActions, (action_creator: any, creator_name: string) => {
```
krzkaczor: is there any reason to use lodash forEach instead of Array.prototype.forEach?

app/modules/actions.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+import { allActions } from "./actions";
+
+const ALLOWED_DUPLICATES: { [type: string]: boolean } = {
+  "@@router/CALL_HISTORY_METHOD": true,
+};
+
+describe("modules", () => {
+  describe("all actions", () => {
+    it("should have unique types", () => {
+      const type_map: { [key: string]: boolean } = {};
```
krzkaczor: what about camel case for symbol names? :D

PR #202, opened 2018-02-23
adds transforming of object keys for apis

huesforalice:


app/lib/api/client/JsonHttpClient.ts
```diff
@@ -3,6 +3,7 @@ import { compact } from "lodash";
 import * as queryString from "query-string";
 import * as urlJoin from "url-join";
 import { Dictionary } from "../../../types";
+import { toCamelCase, toSnakeCase } from "../../../utils/transformObjectKeys";
```
krzkaczor: one can argue that adding these automatic transformations to JsonHttpClient makes it less universal and thus perhaps should be part of another, child class but I think we won't use it for anything else. Just please leave a comment describing this behavior on the class definition.

PR #204, opened 2018-02-23
Restyle Header and add new color palette

### LINKED ISSUE #195: Restyle header for newest designs
app zeplin io_project_5a8a92c89c1a166a6a6e8f37_screen_5a8bee330e09c9b970b4d9b0

banciur:
Added new colour palette 
Header for now just static without any connection to store (will be done as part of dashboard

screencapture-localhost-9090-register-light-seed-1519395117917

closes #195


app/components/Header.module.scss
```diff
@@ -1,60 +1,6 @@
...
-
-.mini-header {
-  font-size: .75rem;
-  color: $white;
-}
-
-.title {
-  font-size: 1rem;
+.logo {
+  height: rem-to-px(50px);
```
krzkaczor: i just realized it should be named `px-to-rem` xD but we can fix it later.

app/styles/variables.scss
```diff
@@ -8,8 +9,23 @@ $denim: #406483;
...
 $bright-orange: #ff6d00;
+// end of old color palette
+
+$white-pure: #fff; // this one was not defined in styleguide but used in designs
+
+// new color palette
+$bright-orange: #ff6d00;
+$white: #efefef;
+$seaweed: #20e06e;
+$piss-yellow: #d5e20f;
```
Moejoe90: :D

banciur: Those are ones from our style guide :)

PR #212, opened 2018-02-23
Connects Wallet Recovery Flow

Moejoe90:


app/modules/wallet-selector/light-wizard/flows.ts
```diff
@@ -58,7 +64,8 @@ export const lightWizardFlows = {
             password,
           );
           await web3Manager.plugPersonalWallet(lightWallet);
-          dispatch(actions.wallet.connected());
+          if (seed) dispatch(actions.routing.goToSuccessfulRecovery());
```
Moejoe90: What do you think here?

app/modules/wallet-selector/light-wizard/flows.ts
```diff
@@ -30,10 +30,16 @@ export const lightWizardFlows = {
         logger: ILogger,
       ) => {
         try {
-          const lightWalletVault = await lightWalletUtil.createLightWalletVault({
-            password,
-            hdPathString: "m/44'/60'/0'",
-          });
+          const lightWalletVault = seed
+            ? await lightWalletUtil.createLightWalletVault({
```
krzkaczor: just pass recoverSeed: seed always and make sure that `createLightWalletVault` checks if its really defined or not.

app/components/shared/StandardLink.tsx
```diff
@@ -0,0 +1,15 @@
+import * as cn from "classnames";
+import * as React from "react";
+import { Link } from "react-router-dom";
+import * as styles from "./StandardButton.module.scss";
+
+interface IButtonProps {
+  text?: string;
+  to?: string;
```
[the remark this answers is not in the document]
Moejoe90: Good idea with destination holder

app/components/shared/StandardLink.tsx
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const StandardLink: React.SFC<IButtonProps> = ({ text, to }) => (
+  <Link className={cn("rounded-0 btn btn-outline-dark w-100", styles.standardButton)} to={to!}>
```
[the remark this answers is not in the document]
Moejoe90: After taking off the optional parameters there is no need for correct

app/modules/wallet-selector/light-wizard/flows.ts
Moejoe90: Done

Moejoe90:
Regarding generating light wallet recovery @krzkaczor you will need to use a BIP39 generator. there are a couple of rules including some words following a specific checksum, in order to generate a correct seed. one of them indeed is you can't have one word repeating. 
https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed

app/components/walletSelector/walletRecover/RecoverWallet.tsx
```diff
@@ -0,0 +1,97 @@
...
+export const RecoveryProcesses: React.SFC<any> = props => {
+  return <RecoveryProcessesComponent {...props} />;
+};
+
+export const RecoverWallet = compose<React.SFC>(
+  appConnect<IProps>({
+    dispatchToProps: dispatch => ({
+      submitForm: (values: IFormValues) => {
+        dispatch(
+          flows.wallet.tryConnectingWithLightWallet(
```
krzkaczor: why all values in IFormVauls are optional?

app/components/walletSelector/walletRecover/RecoverWallet.tsx
```diff
@@ -0,0 +1,97 @@
...
+            values.email as string,
+            values.password as string,
```
krzkaczor: this is very ugly

app/components/walletSelector/walletRecover/RecoverWallet.tsx
krzkaczor: If we need them optional in some place use Optional<IFormValues> and then here, (after validation) you're sure that all of them are provided. I think that validation is simply missing.

app/components/walletSelector/walletRecover/RecoverySuccess.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </Col>
+    </Row>
+  </div>
+);
+
+export const RecoverySuccess = compose<React.SFC>(
+  appConnect<void>({
+    dispatchToProps: dispatch => ({
+      //There must be a guard here if we will continue like this
+      submitForm: () => dispatch(actions.wallet.connected()),
```
krzkaczor: it shouldn't be named `submitForm` then ;) also what do you mean by this comment about guard? Like routing guard? I think would remove this comment since type guard will be on routing level probably.

app/components/walletSelector/walletRegister/RegisterWallet.tsx
```diff
@@ -97,6 +97,6 @@ export const CreateWallet = compose<React.SFC>(
         ),
     }),
   }),
-)(CreateWalletComponent);
+)(RegisterWalletComponent);
 
-//Solve IFormValues complain when elements are not optional
+//Solve IFormValues complains when elements are not optional
```
krzkaczor: you can use Optional<IFormValues> when you are requeired to pass optional values.

app/modules/wallet-selector/light-wizard/flows.ts
krzkaczor: it's good enough for now i think. But it should be part of a different saga/flow later.

app/components/walletSelector/walletRecover/RecoverySuccess.tsx
Moejoe90: I'm more worried that if we dispatch the action `actions.wallet.connected()` it doesn't check if everything went as planned. I haven't had a closer look but from what i see all the important backend communication happens in this action so that's good. 
It was more of a question if something fails and the user still fires this action would there be some checks or not

app/components/walletSelector/walletRegister/RegisterWallet.tsx
Moejoe90: This is a problem related with some code done by Dave, if you make values optional you will notice a crazy bunch of errors, i will follow up on this

krzkaczor: are you talking about runtime errors? about connected/not connected components? if yes this has nothing to do with type definion but with initial values that you need to set to empty strings or whatever.

app/components/walletSelector/walletRecover/RecoverWallet.tsx
Moejoe90: Solved

app/components/walletSelector/walletRecover/RecoverWallet.tsx
Moejoe90: Solved

app/components/walletSelector/walletRecover/RecoverySuccess.tsx
Moejoe90: Name of function changed to `goToDashBoard`

app/components/walletSelector/walletRecover/RecoverWallet.tsx
```diff
@@ -0,0 +1,93 @@
...
+  IMainRecoveryProps,
+  IMainRecoveryState
+> {
+  constructor(props: IMainRecoveryProps) {
+    super(props);
+    this.state = {};
+    this.setSeed = this.setSeed.bind(this);
+  }
+
+  setSeed(words: string): any {
```
krzkaczor: if you use `setSeet = (words: string): void ` you dont need to bind it in constructor

app/components/walletSelector/walletRecover/RecoverySuccess.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { Link } from "react-router-dom";
+import { Col, Row } from "reactstrap";
+import { compose } from "redux";
+import { actions } from "../../../modules/actions";
+import { appConnect } from "../../../store";
+import { StandardButton } from "../../shared/StandardButton";
+import { recoverRoutes } from "./recoverRoutes";
+import { WalletResetHeader } from "./WalletResetHeader";
+
+export const RecoverySuccessComponent: React.SFC<{ goToDashboard: () => void }> = props => (
```
krzkaczor: this is not properly typed

app/components/walletSelector/walletRecover/RecoverySuccess.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </Col>
+    </Row>
+  </div>
+);
+
+export const RecoverySuccess = compose<React.SFC>(
+  appConnect<void>({
+    dispatchToProps: dispatch => ({
+      goToDashboard: () => dispatch(actions.wallet.connected()),
+    }),
```
krzkaczor: and here. please do not leave code like this when you create PR

PR #223, opened 2018-02-25
Derivation path chooser restyle

banciur:


app/styles/variables.scss
```diff
@@ -17,18 +17,20 @@ $white-pure: #fff;
...
-$white: #efefef;
-$seaweed: #20e06e;
-$piss-yellow: #d5e20f;
-$black: #2b2b2b;
-$slate-grey: #656b72;
-$pale-blue: #f6f9fe;
-$silver: #c4c5c6;
-$pale-grey: #e3eaf5;
-$charcoal-grey: #394651;
+$neu-bright-orange: #ff6d00;
```
krzkaczor: whats the reasoning behind renaming these variables?

krzkaczor: Did you do this to match style guide colors?

banciur: It was interfering with bootstrap variables. There is one named `$white` other one is `$black` and maybe others if our style guide change. It would cause mistake every time it would be used by devs.

PR #241, opened 2018-02-27
Add ButtonPrimary and ButtonSecondary components

banciur:
buttons


app/styles/variables.scss
```diff
@@ -57,10 +54,41 @@ $headings-font-family: "Lato", sans-serif;
...
 }
 
+// buttons and forms
+$input-btn-border-width: 2px;
+
+// buttons
+$btn-font-weight: $font-weight-bold;
+
+// sass-lint:disable no-important
+.btn {
```
krzkaczor: please move custom styles out of this file to new one idk called "global-overrides" or something like that.

PR #242, opened 2018-02-27
Light wallet sign

krzkaczor:
- signing with light wallet works
- redid password resetting mechanism to use sagas


app/modules/web3/sagas.ts
```diff
@@ -0,0 +1,64 @@
...
+  [symbols.web3Manager, symbols.logger],
+);
+export function* autoLockLightWalletWatcher(): Iterator<any> {
+  while (true) {
+    const action: TAction = yield take(["NEW_PERSONAL_WALLET_PLUGGED", "WEB3_WALLET_UNLOCKED"]);
+
+    if (action.type === "NEW_PERSONAL_WALLET_PLUGGED" && action.payload.type !== WalletType.LIGHT) {
+      continue;
+    }
+    lockWalletTask = yield* forkAndInject(autoLockLightWallet);
```
huesforalice: you should cancel the task, if there is one, here too, otherwise you might have to lock tasks running at the same time

PR #260, opened 2018-03-02
Login light wallet

krzkaczor:
- [x] Logging in with Light wallet
- [x] Error handling
- [x] normalize data in redux store
- [x] adjust wording for other logic views
- [x] logout should disconnect wallet
- [x] login via email link


app/styles/variables.scss
```diff
@@ -36,6 +36,8 @@ $neu-charcoal-grey: #394651;
 $bigger-font-size: rem-to-px(16px);
 $smaller-font-size: rem-to-px(13px);
 
+$neu-font-size-lg: rem-to-px(20px);
```
banciur: This name is not consistent with ones above.

app/components/walletSelector/WalletBrowser.tsx
```diff
@@ -95,6 +100,7 @@ export const WalletBrowser = compose<React.SFC>(
     stateToProps: state => ({
       errorMessage: state.browserWalletWizardState.errorMsg,
       isLoading: state.browserWalletWizardState.isLoading,
+      isLoginRoute: selectIsLoginRoute(state.router),
```
Moejoe90: Why not pass it as a prop?

app/modules/wallet-selector/light-wizard/sagas.ts
```diff
@@ -0,0 +1,127 @@
...
+    salt: string,
+    email: string,
+  ): Promise<ILightWalletMetadata> {
+    const vaultKey = await getVaultKey(lightWalletUtil, salt, password);
+    try {
+      const vault = await vaultApi.retrieve(vaultKey);
+
+      return {
+        walletType: WalletType.LIGHT,
+        salt,
```
Moejoe90: Is there a reason why this function is returning more than what it's saying it takes I assume 
`retrieveMetadataFromVaultAPI` means retrieving and returning only the vault
In that case there is no need for email

app/styles/variables.scss
krzkaczor: I know but fonts are inconsistent as well... I did it like this because it doesn't override any bootstrap variable.

app/components/walletSelector/WalletBrowser.tsx
krzkaczor: because it connected component anyway 🤷‍♀️ I wanted to avoid pushing down these props across big component tree.

app/modules/wallet-selector/light-wizard/sagas.ts
krzkaczor: it does what it said — it returns wallet metadata. In vault API we store only vault. To create complete wallet metadata we need email as well. I think we could discuss storing more data in the vault api but I worry that it could result in versioning issues.

app/components/walletSelector/WalletBrowser.tsx
Moejoe90: Fair enough

PR #289, opened 2018-03-06
Layout for authorized views

### LINKED ISSUE #281: Generic layout. For authorised routs.
3 stage layout:
- navbar
- sidebar (should be different on different breakpoints)
- content view (fluid)

krzkaczor:
image
image
image

Colors were just added temporarily for presentation purpose. Routing will be improved in separate PR (common layout for all authorized views).

Closes #281

### DIFF
--- a/app/components/layouts/LayoutAuthorized.module.scss
+++ b/app/components/layouts/LayoutAuthorized.module.scss
@@ -1,21 +1,20 @@
 @import "../../styles/neufund-theme";
 
-$xs-height: 60px;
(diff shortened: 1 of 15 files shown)


app/components/layouts/LayoutAuthorized.module.scss
```diff
@@ -0,0 +1,21 @@
+@import "../../styles/neufund-theme";
+
+$xs-height: 60px;
```
banciur: We should use `rem-to-px` function

PR #290, opened 2018-03-06
Adds Auth Header

Moejoe90:
image
image

This PR connectes Header with the state


app/components/Header.spec.tsx
```diff
@@ -0,0 +1,31 @@
...
+    expect(component.find(tid("Header-logout"))).to.have.length(0);
+  });
+
+  it("should render logout button when authorized", () => {
+    const component = shallow(<HeaderComponent isAuthorized={true} logout={mockFunction} />);
+
+    expect(component.find(tid("Header-logout"))).to.have.length(1);
+  });
+
+  it("should Simulate a Click and logout", () => {
```
krzkaczor: i noticed, man, that you really like to capitalize random words in the middle of a sentence 😆

app/components/Header.spec.tsx
```diff
@@ -0,0 +1,31 @@
+import { expect } from "chai";
+import { shallow } from "enzyme";
+import * as React from "react";
+import { spy } from "sinon";
+import { tid } from "../../test/testUtils";
+import { HeaderComponent } from "./Header";
+
+describe("<HeaderComponent />", () => {
+  const mockFunction = spy();
```
krzkaczor: never share spies / mocks across test cases because then you're not sure which test really called it.

app/components/Header.spec.tsx
Moejoe90: Yeah its a curse I have, it gets worse if I'm writing an essay

app/components/Header.spec.tsx
Moejoe90: Nice tip

PR #291, opened 2018-03-06
Add PanelDark component

### LINKED ISSUE #285: Generic "dark" panel component
image
props:
- headerText
- component on the right
- children
- ...props just pass everything to top level container

banciur:
screencapture-localhost-9090-demo-1520336455639

closes #285


app/components/shared/PanelDark.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const PanelDark: React.SFC<IPanelDarkProps & React.HTMLAttributes<HTMLDivElement>> = ({
+  headerText,
+  rightComponent,
+  className,
+  children,
+  ...props
+}) => (
+  <Container {...props} className={cn(styles.panel, className)}>
+    <Row>
+      <Col className={cn(styles.header, "d-flex justify-content-between align-items-center")}>
```
krzkaczor: do you need d-flex here? arent cols flex by default?

banciur: No, only rows

PR #292, opened 2018-03-06
Moe/white panel

Moejoe90:
image
image
image

### DIFF
--- a/app/components/Demo.tsx
+++ b/app/components/Demo.tsx
@@ -1,13 +1,90 @@
 import * as React from "react";
-import { Col, Container, Row } from "reactstrap";
+import { Col, Container, FormFeedback, FormGroup, Input, Row } from "reactstrap";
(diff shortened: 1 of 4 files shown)


app/components/Demo.tsx
```diff
@@ -11,65 +11,95 @@ import {
...
-            <dt>headerText: string</dt>
-            <dd>Title of panel it will be rendered in span element</dd>
-            <dt>rightComponent: ReactNode</dt>
-            <dd>Component that will be put in header on right side.</dd>
-          </dl>
-        </PanelDark>
-      </Col>
-    </Row>
+  <Container>
+    <Container className={styles.demo}>
```
krzkaczor: i don't think that nest containers is a good idea. Also instead of repeating this className just use selector like: `.containerWrapper > .div {...}`

PR #295, opened 2018-03-06
make PanelDark header height fixed, dashboard layout basics

krzkaczor:
image
image

### DIFF
--- a/app/components/dashboard/Dashboard.tsx
+++ b/app/components/dashboard/Dashboard.tsx
@@ -1,12 +1,22 @@
 import * as React from "react";
+import { Col, Row } from "reactstrap";
 import { LayoutAuthorized } from "../layouts/LayoutAuthorized";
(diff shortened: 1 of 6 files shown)


app/components/dashboard/Dashboard.tsx
```diff
@@ -1,12 +1,22 @@
...
 export const Dashboard = () => (
   <LayoutAuthorized>
     <MessageSignModal />
-    <h2>Dashboard</h2>
+    <Row className="p-3">
+      <Col md={9}>
+        <MyPortfolio />
+      </Col>
+    </Row>
+    <br />
```
Moejoe90: Why not just give margin `mb-5` on the Row  and `mt-5` on `<UserInfo />`

krzkaczor: because it's just temporary :D I expect we are gonna get rid of it very soon. I was inspired by @huesforalice to use `<br />` 😆

PR #297, opened 2018-03-07
Notifications

lizonp:
screen shot 2018-03-07 at 13 39 15
screen shot 2018-03-07 at 13 39 01

TO DO:
- ~~add svgs~~
- ~~add test~~


lizonp:
@krzkaczor I just fixed linter issues

app/components/dashboard/notification-widget/NotificationWidgetComponent.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+import { INotification, Notification } from "./Notification";
+
+interface INotificationWidgetComponentProps {
```
krzkaczor: protip: since it's not exported you can use shorter names like IProps but we currently use both in codebase 🤷‍♂️

app/styles/variables.scss
```diff
@@ -19,6 +19,7 @@ $gray2: #85a0b5;
 $gray3: #406483;
 $gray4: #979797;
 $gray5: #aabcca;
+$color-blue-light: #78d7dc;
```
krzkaczor: to make it consistent with others variable names just use `blue-light`.

app/styles/typography.scss
```diff
@@ -0,0 +1,4 @@
+@mixin font-default {
+  font-size: 14px;
```
krzkaczor: i think we already have variable for this default font size ($font-size-base). Also we use REM everywhere.

app/components/dashboard/notification-widget/Notification.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from "react";
+import { Link } from "react-router-dom";
+import * as styles from "./Notification.module.scss";
+
+import * as closeIcon from "../../../assets/img/close.svg";
+import * as infoIcon from "../../../assets/img/notfications/info.svg";
+import * as warningIcon from "../../../assets/img/notfications/warning.svg";
+
+export enum NotificationType {
```
krzkaczor: I think we should not use raw enums anywhere. Use string enums https://stackoverflow.com/a/15491832/580181 or even string literals.

app/components/dashboard/notification-widget/Notification.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const warningClassName = "warning";
+
+export const Notification: React.SFC<INotification> = ({
+  type,
+  text,
+  onClose,
+  actionLink,
+  actionLinkText,
+}) => {
+  const { INFO, WARNING } = NotificationType;
```
krzkaczor: you can move it outside render function

app/components/dashboard/notification-widget/Notification.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const types = {
```
krzkaczor: this and next declaration as well.

app/components/dashboard/Dashboard.tsx
```diff
@@ -1,10 +1,12 @@
 import * as React from "react";
 import { LayoutAuthorized } from "../layouts/LayoutAuthorized";
 import { MessageSignModal } from "../modals/SignMessageModal";
+import { NotificationWidget } from "./notification-widget/NotificationWidget";
 import { UserInfo } from "./UserInfo";
 
 export const Dashboard = () => (
   <LayoutAuthorized>
+    <NotificationWidget />
```
krzkaczor: either pass here className with margins or hardcode it in the component. (probably here makes more sense): like
`<NotificationWidget className="my-3"  />`

we need it because currently first notification is right after navbar.

app/components/dashboard/notification-widget/Notification.module.scss
```diff
@@ -0,0 +1,82 @@
+@import "../../../styles/neufund-theme";
+@import "../../../styles/typography";
```
krzkaczor: you can put this `import` inside `neufund-theme`. The trick is to put there only 0 overhead things like variables/mixins/functions which declaration won't be included in CSS multiple times.

app/components/dashboard/notification-widget/Notification.module.scss
```diff
@@ -0,0 +1,82 @@
...
+
+  border: 1px solid currentColor;
+  color: $neu-slate-grey;
+  display: flex;
+  flex-direction: column;
+  margin-bottom: 20px;
+  padding: 10px 12px;
+  position: relative;
+
+  &:global {
```
krzkaczor: is it the same as &:global(.info) etc?

app/components/dashboard/notification-widget/Notification.tsx
```diff
@@ -0,0 +1,65 @@
...
+    [INFO]: infoIcon,
+    [WARNING]: warningIcon,
+  };
+
+  return (
+    <div data-test-id="notification" className={`${styles.notification} ${types[type]}`}>
+      <i className={`${styles.iconNotificationType}`}>
+        <img src={icons[type]} />
+      </i>
+      <span data-test-id="notification-text" className={`${styles.text}`}>
```
krzkaczor: just use `className={styles.text}`

app/components/dashboard/notification-widget/Notification.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  it("should be of correct type", () => {
+    const componentInfo = shallow(<Notification {...props} type={NotificationType.INFO} />);
+    expect(componentInfo.hasClass(infoClassName)).to.be.true;
+    const componentWarning = shallow(<Notification {...props} type={NotificationType.WARNING} />);
+    expect(componentWarning.hasClass(warningClassName)).to.be.true;
+  });
+
+  it("should call correct click handlers for close button", () => {
```
krzkaczor: always remember about [3A](https://www.google.de/search?ei=-uefWu_6D4OSsAey0qO4DQ&q=arrange+act+assert+test&oq=arrange+act+assert+test&gs_l=psy-ab.3..0i19k1j0i22i30i19k1l8.1063.8586.0.8709.30.29.1.0.0.0.169.2910.2j24.27.0....0...1.1.64.psy-ab..2.27.2906.0..0j0i67k1j0i22i30k1j33i160k1j33i21k1.84.ItbyMwkcZXI) principle while writing tests. and use new line to split each phase. Here arrange and act is kinda merged because it's very simple test. Write it like this:
```
const component = shallow(<Notification {...props} />);

component.find(tid("notification-close")).simulate("click");

expect(props.onClose).to.be.calledOnce;
```

app/components/dashboard/notification-widget/Notification.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+});
+
+describe("<Notification />", () => {
+  let props: any;
+
+  beforeEach(() => {
+    props = defaultProps();
+  });
+
+  it("should be of correct type", () => {
```
krzkaczor: here also make a new line between first expect or even better just split it into two separate tests and wrap it with additional (nested) describe block.

app/styles/variables.scss
lizonp: Changed 👍 Could we consider using `color-` prefix for better ide suggestions?

app/styles/typography.scss
lizonp: Changed 👍

app/components/dashboard/notification-widget/Notification.tsx
lizonp: Changed 👍

app/components/dashboard/notification-widget/Notification.tsx
lizonp: Changed 👍

app/components/dashboard/notification-widget/Notification.module.scss
lizonp: Changed 👍

app/components/dashboard/notification-widget/Notification.module.scss
lizonp: Exactly the same.

app/components/dashboard/notification-widget/Notification.tsx
lizonp: Changed 👍

app/components/dashboard/Dashboard.tsx
lizonp: I am not sure how it's going to be layoutad, that's why I skipped margins on outer element.

app/components/dashboard/notification-widget/Notification.tsx
lizonp: I have changed it.

app/components/dashboard/notification-widget/Notification.spec.tsx
lizonp: I have added extra line there.

PR #298, opened 2018-03-07
adds generic ArrowLink

Moejoe90:


app/components/Demo.tsx
```diff
@@ -101,5 +103,18 @@ export const Demo: React.SFC = () => (
...
     </Container>
+
+    <Container>
+      <Row>
+        <Col>
+          <ArrowLink arrowDirection="right" to="#">
+            Right arrow link
+          </ArrowLink>
+          <ArrowLink arrowDirection="left" to="#">
+            Right arrow link
```
krzkaczor: i think it's "Left arrow link" then xD

app/components/shared/ArrowLink.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+    const component = shallow(
+      <ArrowLink arrowDirection="left" to="#">
+        {testChild}
+      </ArrowLink>,
+    );
+
+    expect(component.contains(testChild)).to.be.true;
+  });
+  it("should render right arrow", () => {
```
krzkaczor: please separate test cases with new lines

app/components/shared/ArrowLink.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react";
+import { LinkProps } from "react-router-dom";
+import { ButtonSecondaryLink } from "./Buttons";
+
+interface IProp extends LinkProps {
```
krzkaczor: I woudnt extend LinkProps since you dont pass every unmatched prop down in render function. I would just explicity type add here `to: string;`.

krzkaczor: `IProps`*

app/components/shared/ArrowLink.tsx
```diff
@@ -0,0 +1,22 @@
...
+  arrowDirection: "left" | "right";
```
krzkaczor: 👍  very good use case for string literals. I think that future MoneyComponent should also use this approach instead of enums CC: @banciur

app/components/Demo.tsx
Moejoe90: done

app/components/shared/ArrowLink.spec.tsx
Moejoe90: done

app/components/shared/ArrowLink.tsx
Moejoe90: done

PR #300, opened 2018-03-08
Money component

### LINKED ISSUE #172: Money component
Take different currencies and format them in a correct way.

krzkaczor:
image

- [x] money component with formatting
- [x] dummy store to hold decimal values (these should be load from blockchain first
- [x] tests

Closes #172

### DIFF
--- a/app/components/Demo.tsx
+++ b/app/components/Demo.tsx
@@ -11,6 +11,7 @@ import {
   ButtonSecondary,
   ButtonSecondaryLink,
 } from "./shared/Buttons";
+import { Money } from "./shared/Money";
 import { NavigationButton, NavigationLink } from "./shared/Navigation";
 import { PanelDark } from "./shared/PanelDark";
 import { PanelWhite } from "./shared/PanelWhite";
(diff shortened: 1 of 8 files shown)


app/components/shared/Money.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const decimalPlaces = selectDecimalPlaces(currency);
+  const currencySymbol = selectCurrencySymbol(currency);
+
+  const formattedMoney = doNotSeparateThousands
+    ? formatMoney(value, decimals, decimalPlaces)
+    : formatThousands(formatMoney(value, decimals, decimalPlaces));
+
+  return (
+    <span {...props}>
+      {formattedMoney}{" "}
```
Moejoe90: Do you need this `{" "}` for a specific reason?

app/store.ts
```diff
@@ -70,7 +72,7 @@ export type IAppState = typeof appStateInstance & {
 export const reducers = combineReducers<IAppState>(allReducers);
 
 interface IAppConnectOptions<S, D, O> {
-  stateToProps?: (state: IAppState) => S;
+  stateToProps?: (state: IAppState, ownProps: O) => S;
```
Moejoe90: Whats the purpose of this?

app/components/Demo.tsx
```diff
@@ -116,5 +117,34 @@ export const Demo: React.SFC = () => (
...
     </Container>
+
+    <Container>
+      <Row>
+        <Col>
+          <Money
+            currency="eth"
+            value={"1234567" + "0".repeat(18)}
+            style={{ fontSize: "25px" }}
+            currencyStyle={{ fontSize: "18px", fontWeight: "bold" }}
```
lizonp: Wouldn't it be better to create enum with classNames (ie. small, medum, large)? I am not sure how universal it has to be, though.

app/components/shared/Money.tsx
krzkaczor: yeah - it creates space between value an currency item.

app/store.ts
krzkaczor: This allows to access props passed directly to connected component. Actually i think Dave fixed it as well so i am gonna rebase on master

app/components/Demo.tsx
krzkaczor: Currently, I think designs are wildly different so i made it generic. If we see some similarities among views we can extract them later.

PR #310, opened 2018-03-12
Pl/layout fix

lizonp:
screen shot 2018-03-12 at 10 00 29
screen shot 2018-03-12 at 10 00 18


webpack/webpack.config.dev.js
```diff
@@ -19,7 +19,7 @@ module.exports = merge(configCommon, {
         pathRewrite: { "^/node": "" },
       },
       "/api/signature": {
-        target: "http://localhost:5000",
```
huesforalice: why is this changed? It should stay at 5000, afaik there have been no changes here.

app/styles/global.scss
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  display: flex;
+  flex-direction: column;
+  min-height: calc(100% - #{$header-height});
+}
+
+.layout-container {
+  @include media-breakpoint-up (md) {
+    padding-left: 30px;
```
huesforalice: if you use pixels, please use px-to-rem, so 30px becomes px-to-rem(30px)

huesforalice: In general you can use bootstrap classes for padding and spacing, please see this documentation: https://getbootstrap.com/docs/4.0/utilities/spacing/

app/components/dashboard/notification-widget/Notification.module.scss
```diff
@@ -36,6 +31,10 @@
     }
   }
 
+  + #{&} {
```
Moejoe90: why is `id` used here?

app/styles/global.scss
lizonp: It's changed.

app/components/dashboard/notification-widget/Notification.module.scss
lizonp: It's sass interpolation. So it's the same as `.notification + .notification`

webpack/webpack.config.dev.js
lizonp: I have commited it by mistake, it's changed now.

Moejoe90:
btw i think you can use `w-100` for the same exact thing

app/styles/global.scss
```diff
@@ -0,0 +1,34 @@
...
+  @include media-breakpoint-up (md) {
+    padding-left: px-to-rem(30px);
+    padding-right: px-to-rem(30px);
+    padding-top: px-to-rem(20px);
+  }
+
+  max-width: 100%;
+  min-height: 100%;
+  overflow-x: hidden;
+  overflow-y: scroll;
```
banciur: Do we need to be like this? Now I get double scroll bars one that is inactive. 
scroll

app/styles/global.scss
```diff
@@ -0,0 +1,34 @@
+html,
```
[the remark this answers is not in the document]
lizonp: I's changed 👍

app/components/layouts/LayoutAuthorized.tsx
```diff
@@ -6,12 +6,14 @@ import { LayoutAuthorizedMenu } from "./LayoutAuthorizedMenu";
...
-  <Row className="h-100">
-    <Col xs={12} sm="auto">
-      <div className={styles.menu}>
-        <LayoutAuthorizedMenu />
-      </div>
-    </Col>
-    <Col className={styles.content}>{children}</Col>
-  </Row>
+  <>
+    <div className={styles.menu}>
```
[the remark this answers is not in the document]
lizonp: I's changed 👍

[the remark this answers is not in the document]
banciur: It's still here - https://github.com/Neufund/platform-frontend/blob/pl/layout-fix/app/components/layouts/LayoutAuthorized.tsx#L10
This line instead `<div className={styles.menu}>` should be just `<div>`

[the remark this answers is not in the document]
lizonp: @banciur just fixed it

app/styles/global.scss
```diff
@@ -0,0 +1,32 @@
...
+body,
+[id="app"] {
+  height: 100%;
+  min-height: 100%;
+}
+
+.wrapper {
```
krzkaczor: okay we will figure today if we need this or we go with row/column or something else. for now it stays

PR #311, opened 2018-03-12
Add typography mixins

lizonp:
I have added mixins with fonts from zeplin.


lizonp:
This file is imported by `neufund-theme`, I've added it for future use.

PR #315, opened 2018-03-12
Adds MyWallet Widget

Moejoe90:
image
image
image

The Wallet link on Xs was changed from `Manage Wallet` to `Wallet` after discussions, I will keep like this for now until the layout is more clear 

Additionally, this PR will not look into the general layout, for example, myPortfolio being getting very close. 

My only point of concern is this 

image

There is simply no space to put this anywhere in at some pixel i think this is a good solution but please give me comments if there are any better ideas


app/components/dashboard/myWallet/MyWalletWidget.module.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.small-red-font {
+  font-size: px-to-rem(14px);
+  color: $neu-red;
+}
+
+@include media-breakpoint-up (lg) {
+  .border {
+    border-bottom: solid 1px;
```
huesforalice: I guess here we don't nee a px-to-rem?

Moejoe90: No we actually do need `px-to-rem`, this pr was still in WIP phase

PR #322, opened 2018-03-14
Adds light wallet seed backup saga logic 

Moejoe90:


app/components/settings/backupSeed/BackupSeedWidget.tsx
```diff
@@ -6,13 +6,27 @@ import * as successIcon from "../../../assets/img/notfications/Success_small.svg
...
+import { ButtonPrimary } from "../../shared/Buttons";
 import { PanelDark } from "../../shared/PanelDark";
 
-export const BackupSeedWidgetComponent: React.SFC<IUser> = ({ backupCodesVerified }) => {
+interface IStateProps {
+  backupCodesVerified?: boolean;
+}
+
+interface IDispatchProps {
+  verifyBackupPhrase?: () => void;
```
Moejoe90: This is done on purpose for testing etc, in reality there should be no dispatched actions here this is only for testing

app/modules/wallet-selector/light-wizard/sagas.ts
```diff
@@ -76,6 +79,23 @@ export const getWalletMetadata = injectableFn(
...
+export const lightWalletBackupWatch = injectableFn(
+  function*(getState: GetState): any {
+    while (true) {
+      const action: TAction = yield neuTake("LIGHT_WALLET_BACKUP");
+      if (action.type !== "LIGHT_WALLET_BACKUP") {
+        continue;
+      }
+      try {
+        const user = getState().auth.user;
+        yield effects.spawn(updateUser, { ...user, backupCodesVerified: true });
```
huesforalice: You will want to do effects.call here. If you spawn, the catch block won't execute on error, because it creates a new thread.

Moejoe90: I see what you mean

PR #324, opened 2018-03-15
adds default values for wallet values and a few selectors

huesforalice:


app/components/dashboard/Dashboard.tsx
```diff
@@ -17,12 +33,11 @@ export const Dashboard = () => (
...
-          euroTokenAmount={"36490" + "0".repeat(18)}
-          euroTokenEuroAmount={"36490" + "0".repeat(18)}
-          ethAmount={"66482" + "0".repeat(14)}
-          ethEuroAmount={"6004904646" + "0".repeat(16)}
-          percentage="-3.67"
-          totalAmount={"637238" + "0".repeat(18)}
+          euroTokenAmount={props.liquidEuroTokenBalance}
+          ethAmount={props.liquidEtherBalance}
+          ethEuroAmount={props.liquidEtherBalanceEuroAmount}
+          percentage="0"
```
Moejoe90: Maybe we should do a `todo` comment here for connecting `percentage`

huesforalice: done

PR #337, opened 2018-03-19
global dependency context

huesforalice:


app/di/setupBindings.ts
```diff
@@ -145,3 +154,18 @@ export function customizerContainerWithMiddlewareApi(
 
   return container;
 }
+
+export const createGlobalDependencies = (container: Container) => ({
```
krzkaczor: I would leave comment here saying why we do this. Something like: "We use plain object for injecting deps into sagas"

app/modules/kyc/sagas.ts
```diff
@@ -80,12 +81,11 @@ function* loadIndividualFiles(action: TAction): Iterator<any> {
   }
 }
 
-function* loadIndividualRequest(action: TAction): Iterator<any> {
+function* loadIndividualRequest(deps: TGlobalDependencies, action: TAction): Iterator<any> {
```
krzkaczor: i would prefer object destruction here: `{ apiKycService }: TGlobalDependencies` but i think it's up to the author ;)

app/di/setupBindings.ts
```diff
@@ -145,3 +154,39 @@ export function customizerContainerWithMiddlewareApi(
...
 }
+
+/**
+ * We use plain object for injecting deps into sagas"
+ */
+export const createGlobalDependencies = (container: Container) => ({
+  // misc
+  logger: container.get<ILogger>(symbols.logger),
+  notificationCenter: container.get<NotificationCenter>(symbols.notificationCenter),
+  getState: container.get<GetState>(symbols.getState),
```
krzkaczor: I would put TODO note here: it should be removed and we can use just `select` from redux-saga

app/index.tsx
```diff
@@ -66,6 +75,7 @@ function startupApp(history: any): { store: Store<IAppState>; container: Contain
   );
 
   const store = createStore(reducers, composeWithDevTools(middleware));
+  context.deps = createGlobalDependencies(container);
```
krzkaczor: could you leave a note here as well? like: "we do it b/c we to create dependencies object we need already dispatch and getState inside container"

PR #348, opened 2018-03-20
Adds change email flow

Moejoe90:
image
image
image

This what is missing is the general layout


app/lib/api/users/interfaces.ts
```diff
@@ -5,6 +5,7 @@ export interface IUser {
   language?: string;
   unverifiedEmail?: string;
   verifiedEmail?: string;
+  new_email?: string;
```
huesforalice: this should be newEmail, the props get automatically transformed to snake case on the network request layer

app/modules/settings/sagas.ts
```diff
@@ -0,0 +1,30 @@
...
+import { neuTakeEvery } from "../sagas";
+import { actions } from "./../actions";
+
+export function* addNewEmail(
+  { notificationCenter: NotificationCenter }: TGlobalDependencies,
+  action: TAction,
+): Iterator<any> {
+  if (action.type !== "SETTINGS_ADD_NEW_EMAIL") return;
+  const email = action.payload.email;
+  const user = yield select((s: IAppState) => selectUser(s.auth));
```
huesforalice: You should be able to write yield select(selectUser), the selectors always automatically operate on the main state object. Try it out,if it does not work let me know.

app/modules/settings/sagas.ts
```diff
@@ -0,0 +1,30 @@
...
+import { TGlobalDependencies } from "../../di/setupBindings";
+import { IAppState } from "../../store";
+import { TAction } from "../actions";
+import { selectUser } from "../auth/reducer";
+import { updateUser } from "../auth/sagas";
+import { neuTakeEvery } from "../sagas";
+import { actions } from "./../actions";
+
+export function* addNewEmail(
+  { notificationCenter: NotificationCenter }: TGlobalDependencies,
```
huesforalice: You don't need to type notificationcenter, it is autotyped.

app/modules/settings/sagas.ts
```diff
@@ -0,0 +1,30 @@
...
+  action: TAction,
+): Iterator<any> {
+  if (action.type !== "SETTINGS_ADD_NEW_EMAIL") return;
+  const email = action.payload.email;
+  const user = yield select((s: IAppState) => selectUser(s.auth));
+  try {
+    yield effects.call(updateUser, { ...user, new_email: email });
+    yield effects.put(actions.routing.goToSettings());
+    NotificationCenter.info("New Email added");
```
huesforalice: this looks like some kind of global scope variable? use the lowercase notificationCenter, as this is the var that is injected here.

app/modules/settings/sagas.ts
```diff
@@ -0,0 +1,30 @@
...
+  } catch {
+    NotificationCenter.error("Failed to change email");
```
huesforalice: Same as above

app/components/settings/changeEmail/ChangeEmail.tsx
```diff
@@ -0,0 +1,71 @@
...
+    <FormField
+      type={"email"}
+      placeholder="Enter new email"
+      touched={formikBag.touched}
+      errors={formikBag.errors}
+      name={EMAIL}
+    />
+    <div className="text-center">
+      <ButtonPrimary type="submit" disabled={!formikBag.values.email}>
+        Login
```
huesforalice: Maybe call this "submit" or something like that.

app/components/settings/changeEmail/ChangeEmail.tsx
```diff
@@ -0,0 +1,71 @@
...
+})(ChangeEmailForm);
+
+export const ChangeEmailComponent: React.SFC<IDispatchProps> = props => {
+  return (
+    <>
+      <PanelDark headerText="CHANGE MY E-MAIL ADDRESS">
+        <Row>
+          <Col xs={6} className="mt-3">
+            <p data-test-id="paragraph-section">
+              Lorem ipsum dolor sit amet, consetetur, sadpcsi and then that is how you get the email
```
huesforalice: Maybe add some remotely sensible text here :) Not super important but at least we don't release garbage text in case we overlook to change this.

app/lib/api/users/interfaces.ts
Moejoe90: Done

app/modules/settings/sagas.ts
Moejoe90: cool didn't know this

app/modules/settings/sagas.ts
Moejoe90: nice

app/modules/settings/sagas.ts
Moejoe90: done

app/modules/settings/sagas.ts
Moejoe90: done

app/components/settings/changeEmail/ChangeEmail.tsx
Moejoe90: true

app/components/settings/changeEmail/ChangeEmail.tsx
Moejoe90: Hmm, this will require some creativity I suggest we keep as is so that it's clear that it needs to be changed. if I write something that is presumably making sense during demo there might be an assumption that there was already wording work done here.

app/modules/settings/sagas.ts
krzkaczor: I think this is not true. I think selectors opearate always on subtree (so they are dont bound to whole app state but only subtree that they care about).

app/modules/settings/sagas.ts
```diff
@@ -1,27 +1,26 @@
...
   try {
     yield effects.call(updateUser, { ...user, new_email: email });
     yield effects.put(actions.routing.goToSettings());
-    NotificationCenter.info("New Email added");
+    notificationCenter.info("New Email added");
   } catch {
-    NotificationCenter.error("Failed to change email");
-    throw new Error();
+    notificationCenter.error("Failed to change email");
+    throw new Error("Email change failed");
```
huesforalice: Sorry, missed this one. Don't rethrow this in the saga unless you catch it somewhere, this will destroy all the sagas (I have built a recovery mechanism in the root saga for this,  to be safe though)

Moejoe90: Ok good that you pointed this out

app/components/settings/changeEmail/ChangeEmail.spec.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { shallow } from "enzyme";
+import * as React from "react";
+import { spy } from "sinon";
+import { tid } from "../../../../test/testUtils";
+import { ChangeEmailComponent } from "./ChangeEmail";
+
+describe("<ChangeEmailComponent />", () => {
+  it("Should render Component", () => {
+    const submitFormMock = spy();
+    const MyNeuWidgetComponent = shallow(<ChangeEmailComponent submitForm={submitFormMock} />);
```
krzkaczor: remember what i like about new lines between stages in tests? ;d This is one of very few style releated things that i want to enforce. (for other stuff we have prettier)

app/components/settings/changeEmail/ChangeEmail.tsx
```diff
@@ -0,0 +1,71 @@
...
+  mapPropsToValues: props => props.currentValues as IFormValues,
+  handleSubmit: (values, props) => {
+    props.props.submitForm(values);
+    values.email = "";
+  },
+})(ChangeEmailForm);
+
+export const ChangeEmailComponent: React.SFC<IDispatchProps> = props => {
+  return (
+    <>
```
krzkaczor: <> is not needed.

PR #352, opened 2018-03-21
Notifications

### LINKED ISSUE #346: Connect notifications widgets to store
Notification widget should have it's own redux state that will contain notifications list and actions to add and remove notifications.

Notification actions can hidden by clicking on x symbol in right corner. 
Error notifications dispatches actions on text visible on right side (instead x)

banciur:
closes #346


app/components/dashboard/Dashboard.tsx
```diff
@@ -19,13 +21,14 @@ interface IStateProps {
...
   liquidEuroTotalAmount: string;
+  notifications?: INotification[];
 }
 
 type IProps = IStateProps;
 
 export const DashboardComponent = (props: IProps) => (
   <LayoutAuthorized>
-    <NotificationWidget />
+    <NotificationWidget notifications={props.notifications} />
```
krzkaczor: these notifications should be renamed to `additionalNotifications` (since component is connected anyway)

krzkaczor: and whole widget should be a part of layout as we discussed (so it's visible on every screen).

app/modules/notifications/reducer.ts
```diff
@@ -0,0 +1,69 @@
...
+    case "NOTIFICATIONS_ADD": {
+      const pNotification = action.payload.notification;
+
+      const newState = {
+        notifications: Array.from(state.notifications),
+      };
+
+      const id = pNotification.id ? pNotification.id : Date.now();
+
+      newState.notifications.push({
```
krzkaczor: dont use push in reducers!

PR #353, opened 2018-03-21
Unify ulpoad component for business and personal upload

lizonp:
I have unified it. Now it has two 'modes'. personal and business. They have different list and personal have extra image with passport.


app/components/kyc/shared/KycFileUploadList.module.scss
```diff
@@ -3,36 +3,32 @@
...
-      .title {
-        display: none;
-      }
-    }
-
-    .upload-zone {
-      padding-left: px-to-rem(15px);
-    }
-  }
+  // &:global(.business) {
```
krzkaczor: just remove commented code?

lizonp: Sure, strange that linter didn't catch that.

PR #357, opened 2018-03-21
Smartcontracts integration

krzkaczor:
- [x] initialize smart contract structure
- [x] add ethereum address to the wallet metadata
- [x] get all wallet data


app/modules/contracts/ContractsService.ts
```diff
@@ -0,0 +1,68 @@
+import { inject, injectable } from "inversify";
```
huesforalice: ContractsService.ts should be in the lib folder

app/components/dashboard/myWallet/MyWalletWidget.tsx
```diff
@@ -41,50 +118,32 @@ export const MyWalletWidget: React.SFC<IMyWalletWidget> = ({
...
   );
 };
+
+export const MyWalletWidget = compose<React.SFC>(
+  onEnterAction({ actionCreator: d => d(actions.wallet.startLoadingWalletData()) }),
+  appConnect<StateProps>({
+    stateToProps: s => {
+      const isLoading = s.wallet.loading;
+      if (!isLoading) {
+        const walletData = s.wallet as IWalletStateLoaded; // TS doesnt realize here that we are sure that its wallet already loaded
```
huesforalice: Won't this fail in case there is an error loading?

app/modules/init/sagas.ts
```diff
@@ -4,22 +4,27 @@ import { TGlobalDependencies } from "../../di/setupBindings";
...
   try {
+    yield neuCall(detectUserAgent);
     yield web3Manager.initialize();
 
     const jwt = yield neuCall(loadJwt);
     if (jwt) {
       yield loadUser();
     }
     yield neuCall(loadPreviousWallet);
+    yield neuCall(initializeContracts);
```
huesforalice: This seems to me built in a way, that when trying to initialize the contracts fails, the whole app won't start. There will be quite a few requests to ethereum to set up all the contracts no? Maybe this should be a bit more robust.

krzkaczor: I think this is what we are going to do — if app won't connect to blockchain it's not usable. It seems that there will be a lot of views with data coming both from API and blockchain.

app/components/dashboard/myWallet/MyWalletWidget.tsx
krzkaczor: yeah... error handling is missing. Let me fix that

krzkaczor:
@huesforalice i fixed your comments. And now it should be clear how to run it ;) please let me know what do you think

huesforalice:
please add NF_UNIVERSE_CONTRACT_ADDRESS to .env file

PR #367, opened 2018-03-22
removes global selector in Settings Saga

Moejoe90:


app/modules/settings/sagas.ts
```diff
@@ -13,9 +14,9 @@ export function* addNewEmail(
 ): Iterator<any> {
   if (action.type !== "SETTINGS_ADD_NEW_EMAIL") return;
   const email = action.payload.email;
-  const user = yield select(selectUser);
+  const user = yield select((s: IAppState) => selectUser(s.auth));
   try {
-    yield effects.call(updateUser, { ...user, new_email: email });
+    yield effects.call(updateUser, { ...user, newEmail: email });
```
krzkaczor: ehhh this should be caught by type system :/

Moejoe90: Why was it not caught?

PR #369, opened 2018-03-22
refactor signing model, so it now is used for message signing, jwt permissions and seed retrieval

huesforalice:
…re versatile


app/components/modals/AccessWalletModal.tsx
```diff
@@ -21,38 +23,40 @@ interface IDispatchProps {
...
-      )}
-
-      <p>{props.errorMsg}</p>
-      <LoadingIndicator />
+      <div className="text-md-center">
+        <h1>{props.title}</h1>
+        <p>{props.message}</p>
+        <img src={lockIcon} className="mb-3" />
+        {props.isLightWallet ? (
+          <LightWalletSignPrompt {...props} />
```
krzkaczor: shouldnt these be renamed as well? to something like `LightWalletAccessPrompt`?

app/components/modals/AccessWalletModal.tsx
```diff
@@ -21,38 +23,40 @@ interface IDispatchProps {
...
+        ) : (
+          <GenericSignPrompt onCancel={props.onCancel} />
```
krzkaczor: here as well

app/components/settings/backupSeed/BackupSeed.tsx
```diff
@@ -1,97 +1,59 @@
...
+  componentWillMount(): void {
+    this.props.getSeed();
+  }
+
   componentWillUnmount(): void {
     this.props.clearSeed();
   }
 
-  renderUnlockedWalletComponent(): React.ReactNode {
+  render(): React.ReactNode {
```
krzkaczor: good refactoring in this component 👍 finally we got rid of this ugly fetchSeed in random place in render method.

PR #372, opened 2018-03-23
fixes visuals on kyc after buttons update

huesforalice:


app/components/kyc/business/BusinessData.tsx
```diff
@@ -57,10 +57,11 @@ const KYCForm = (formikBag: FormikProps<IKycBusinessData> & IProps) => (
       </Col>
     </Row>
     <FormSelectCountryField label="Country" name="country" />
-    <br />
-    <Button type="submit" disabled={!formikBag.isValid || formikBag.loadingData}>
-      Save
-    </Button>
+    <div className="p-4 text-center">
```
krzkaczor: woudnt' adding className directly to button have the same effect?

PR #384, opened 2018-03-27
fix light wallet reconnection bug

krzkaczor:


huesforalice:
Why is walletInstance on IVault not properly typed? This is what made this error possible in the first place.

krzkaczor:
@huesforalice bc it's coming from 3rd party lib that doesnt have typings

PR #392, opened 2018-03-27
Multiple layout fixes

krzkaczor:
Unify spacings on multiple pages. 
Fixed react warning while displaying unlock wallet popup.

image
image


app/components/wallet/wallet-balance/WalletBalance.module.scss
```diff
@@ -1,17 +1,5 @@
 @import "../../../styles/neufund-theme";
 
-.wallet-balance {
-  &:global {
```
krzkaczor: It doesnt have any effect because it's hidden my PanelDark background anyway. Am I missing somethign @lizonp

PR #393, opened 2018-03-27
Deposit flow visuals

lizonp:
screen shot 2018-03-27 at 22 07 56
screen shot 2018-03-27 at 22 08 19
screen shot 2018-03-27 at 22 08 24


huesforalice:
Looks nice. In terms of project structure, we had decided that deposting is a subview of the wallet, and  we want individual routers for both pages:
/wallet/deposit/eth for ethereum deposit
/wallet/deposit/neur for euro deposit
So these should be added in the wallet router, and the tab buttons should be issuing routing actions to these pages. @lizonp do you know how to do this or do you need help with this?

app/components/wallet/Router.tsx
```diff
@@ -0,0 +1,25 @@
...
+    {/* Manage Wallet */}
+    <Route path={walletRoutes.manageWallet} component={ManageWallet} exact />
+
+    {/* Deposit Funds */}
+    <Route
+      path={walletRoutes.euroToken}
+      render={() => <DepositEuroToken path={walletRoutes.euroToken} />}
+      exact
+    />
+    <Route path={walletRoutes.eth} render={() => <DepositEth path={walletRoutes.eth} />} exact />
```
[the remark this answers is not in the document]
lizonp: I ahve added redirection to default wallet view.

app/components/AppRouter.tsx
```diff
@@ -1,6 +1,6 @@
 export const appRoutes = {
+  depositFunds: "/deposit-funds",
```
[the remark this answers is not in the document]
lizonp: I have removed it.

app/components/Demo.tsx
```diff
@@ -335,5 +336,25 @@ export const Demo: React.SFC = () => (
...
+    <Container>
+      <Tabs
+        theme="t-dark"
+        tabs={[
+          { text: "tab 1", isSelected: false },
+          { text: "tab 2", isSelected: true },
+          { text: "tab 3", isSelected: false },
+        ]}
+      />
+      <Tabs
```
krzkaczor: do we need theme prop on this component at all? We don't use it anywhere else then demo page and it looks kinda ugly... Also can you normalize these theme properties? There are multiple places when "light"|"dark" union type is used. I would love to have one theme type that we reuse across components. (but still i think here is simply not needed ;) ) Another thing that i want to simplify usaing component as much as we can so i go with simpler strings like light | dark (instead of t-light and t-dark) and have `t-` prefix added inside the component when it's really needed.

app/components/depositFunds/DepositFunds.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { LayoutAuthorized } from "../layouts/LayoutAuthorized";
+import { BreadCrumb } from "../shared/BreadCrumb";
+import { Button } from "../shared/Buttons";
+import { PanelWhite } from "../shared/PanelWhite";
+import { Tabs } from "../shared/Tabs";
+import * as styles from "./DepositFunds.module.scss";
+import { MethodEth } from "./methods/MethodEth";
+import { MethodEuroToken } from "./methods/MethodEuroToken";
+
+export const DepositFunds = () => {
```
krzkaczor: this file is not used anywhere.

app/components/depositFunds/methods/MethodEth.module.scss
```diff
@@ -0,0 +1,94 @@
+@import "../../../styles/neufund-theme";
+
+.method-eth {
```
krzkaczor: aren't these files duplicated?

krzkaczor: all of them in `app/components/depositFunds/` dir

lizonp: ok, I have removed all of them.

app/components/Demo.tsx
lizonp: I have added it, because I have spotted different version somewhere in other screens designs. I agree that we should unify naming for themes. I have been using t- prefix so I could quickly spootlight theme class in stylesheet, but if you prefer to remove it then I don't see any problems.

app/components/shared/Tabs.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import * as styles from "./Tabs.module.scss";
+
+type TTheme = "t-dark" | "t-light";
+
+interface ITab {
```
krzkaczor: there is already `WalletSelectorNavigation` component that does somehing similar. Please unify them. You can use NavLink to automatically select currently selected tab. This simplifies parent component (no need to hold path)

app/components/Demo.tsx
krzkaczor: Okay so keeping it around makes sense - that's fine. Reusing this prop for class name bounds them together which is not ideal since you want to make sure that unique (thus the prefix right?). So please take a look at https://github.com/Neufund/platform-frontend/blob/master/app/components/wallet/wallet-balance/WalletBalance.tsx#L35 you can use same approach or extract function like `mapThemeIdToClassName` please unify usage.

app/components/shared/Tabs.tsx
lizonp: I have modified theme property to accept only 'dark' and light' theme. Also I have hanged it to use NavLink.

PR #396, opened 2018-03-28
D#/idnow page

huesforalice:
adds instant id page to kyc
includes network requests


app/modules/kyc/sagas.ts
```diff
@@ -120,6 +120,19 @@ function* submitIndividualRequest(
   }
 }
 
+function* startIndividualInstantId({
+  apiKycService,
+  notificationCenter,
+}: TGlobalDependencies): Iterator<any> {
+  try {
+    const result: IHttpResponse<IKycRequestState> = yield apiKycService.startInstantId();
+    if (result.body.redirectUrl) window.location.replace(result.body.redirectUrl);
```
krzkaczor: if there is no redirectUrl should we rais excetion or is it normal?

app/components/kyc/personal/InstantId.tsx
```diff
@@ -0,0 +1,71 @@
...
+      </Button>
+    </div>
+  </KycPanel>
+);
+
+export const KycPersonalInstantId = compose<React.SFC>(
+  appConnect<IStateProps, IDispatchProps>({
+    dispatchToProps: dispatch => ({
+      onStartInstantId: () => dispatch(actions.kyc.kycStartInstantId()),
+      onContinue: () => dispatch(actions.routing.goToKYCIndividualUpload()),
```
krzkaczor: should be renamed to onManualVerification?

app/modules/kyc/sagas.ts
```diff
@@ -120,6 +120,19 @@ function* submitIndividualRequest(
...
+    yield put(actions.kyc.kycUpdateIndividualRequestState(false, result.body));
```
krzkaczor: i think here result.body should be passed without redirectUrl b/c otherwise it will be saved to reducer but not typed properly.

app/components/kyc/personal/InstantId.tsx
huesforalice: done

app/modules/kyc/sagas.ts
huesforalice: there will always be one, if this request can't be executed, there will be an error (400 or 500) which will land in the try catch block.

krzkaczor: so why do we need if?

app/modules/kyc/sagas.ts
huesforalice: it should be typed properly and this value is used elsewhere in the app, so it should stay

app/modules/kyc/sagas.ts
huesforalice: otherwise tslint complains otherwise. Redirecturl will always be there when hitting this endpoint, but will not be when putting a normal kyc request. I can also create a subtype of kycrequest with this value marked as non optional if you like.

app/modules/kyc/sagas.ts
krzkaczor: oh it's fine i didnt saw it first time. nvm :D

PR #400, opened 2018-03-28
Basic handling of expired JWT token

### LINKED ISSUE #380: Remove expired JWTs during app init
During app init it can happen that JWT from local storage is already expired. We should logout user then. Currently app goes into broken state.

banciur:
closes #380

### DIFF
--- a/app/modules/init/sagas.ts
+++ b/app/modules/init/sagas.ts
@@ -1,9 +1,10 @@
 import { effects } from "redux-saga";
 import { fork, put } from "redux-saga/effects";
+
 import { TGlobalDependencies } from "../../di/setupBindings";
+import { isJwtExpiringLateEnough } from "../../utils/JWTUtils";
 import { actions } from "../actions";
 import { loadJwt, loadUser } from "../auth/sagas";
-
 import { initializeContracts } from "../contracts/sagas";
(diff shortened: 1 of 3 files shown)


app/modules/init/sagas.ts
```diff
@@ -15,9 +16,17 @@ function* setup({ web3Manager, logger }: TGlobalDependencies): Iterator<any> {
...
     const jwt = yield neuCall(loadJwt);
+
     if (jwt) {
-      yield loadUser();
+      if (isJwtExpiringLateEnough(jwt)) {
+        yield loadUser();
+      } else {
+        yield put(actions.auth.logout());
+        yield put(actions.routing.goToLogin());
+        logger.error("JTW expiring to soon.");
```
krzkaczor: too

app/utils/JWTUtils.ts
```diff
@@ -1,9 +1,29 @@
+import * as Moment from "moment";
+
 interface IJwt {
+  exp: number;
   permissions?: {
     [name: string]: number;
   };
 }
 
+const MAX_EXPIRATION_DIFF_MINUTES = 10;
```
krzkaczor: there is a file with constants

PR #409, opened 2018-04-03
Add missing key attribute for Tabs component

lizonp:


app/components/shared/Tabs.tsx
```diff
@@ -21,8 +21,8 @@ interface IProps {
 
 export const Tabs: React.SFC<IProps> = ({ tabs, theme, className, ...props }) => (
   <div className={cn(styles.tabs, className)} {...props}>
-    {tabs.map(({ path, text, handleClick, dataTestId }) => (
-      <NavLink to={path} className={cn(styles.tab, theme)} data-test-id={dataTestId}>
+    {tabs.map(({ path, text, handleClick, dataTestId }, index) => (
+      <NavLink to={path} className={cn(styles.tab, theme)} data-test-id={dataTestId} key={index}>
```
krzkaczor: probably in this case you could use path or text (they will be unique) as key but i am fine with this solution.

PR #411, opened 2018-04-03
clears whole state during logout

Moejoe90:
fixes #411

### DIFF
--- a/app/modules/auth/reducer.ts
+++ b/app/modules/auth/reducer.ts
@@ -24,8 +24,7 @@ export const authReducer: AppReducer<IAuthState> = (
         ...state,
         jwt: action.payload.jwt,
       };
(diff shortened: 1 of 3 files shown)


app/modules/auth/reducer.ts
```diff
@@ -24,8 +24,7 @@ export const authReducer: AppReducer<IAuthState> = (
         ...state,
         jwt: action.payload.jwt,
       };
-    case "AUTH_LOGOUT":
-      return authInitialState;
+    //Log out is done on whole initial state instead of just AUTH reducer
```
krzkaczor: i think you wanted here to write `...on whole state instead of...` (no initial)

Moejoe90: True good catch

PR #421, opened 2018-04-06
Moe/email activation

Moejoe90:


app/modules/auth/sagas.ts
```diff
@@ -162,5 +195,6 @@ export function* ensurePermissionsArePresent(
 
 export const authSagas = function*(): Iterator<effects.Effect> {
   yield fork(neuTakeEvery, "AUTH_LOGOUT", logoutWatcher);
+  yield fork(neuTakeEvery, "AUTH_VERIFY_EMAIL", VerifyUserEmail);
```
krzkaczor: @Moejoe90 general rule -  only upercase classes names. This should be `verifyUserEmail `

app/modules/auth/sagas.ts
```diff
@@ -78,14 +97,28 @@ function* signInUser(): Iterator<any> {
...
+    }
   } catch (e) {
     yield effects.put(
       actions.walletSelector.messageSigningError("Error while connecting with server!"),
     );
   }
 }
 
+function* VerifyUserEmail(): Iterator<any> {
+  yield neuCall(VerifyUserEmailPromise);
```
krzkaczor: same here — lowercase

PR #440, opened 2018-04-10
disable KYC button on unverified email

krzkaczor:
image


app/components/settings/kycStates/KycStatusWidget.tsx
```diff
@@ -37,6 +39,18 @@ const statusTextMap: { [status: string]: string } = {
     "Your instant identification is being processed. You will be notified by e-mail once this is completed.",
 };
 
+const getStatus = (selectIsUserEmailVerified: boolean, requestStatus?: TRequestStatus): string => {
+  if (!selectIsUserEmailVerified) {
+    return "You need to verify email before starting KYC";
+  }
+
+  if (!requestStatus) {
```
Moejoe90: To be honest i think requestStatus should never be undefined, but that's just my opinion because like this its having a complete different state from the UNION type

app/components/settings/kycStates/KycStatusWidget.tsx
```diff
@@ -37,6 +39,18 @@ const statusTextMap: { [status: string]: string } = {
...
+    return "";
+  }
+
+  return statusTextMap[requestStatus];
+};
+
 export const KycStatusWidgetComponent: React.SFC<IProps> = props => {
```
Moejoe90: Again this is a matter of style but I would suggest we add to the guidelines either 
1 - spreading all props into there respective vars 
2 - Always use `props.something`

krzkaczor: Uhh i think it depends on a case. Basically, if there are not so many props then probably spreading it makes more sense. I will try to add something about it to our guidelines.

app/components/settings/kycStates/KycStatusWidget.tsx
krzkaczor: But i agree its not perfect. But trying to type it properly with union types (whole state) tends to be super difficult — i tried it already somewhere.

PR #443, opened 2018-04-10
Fix wallet calcs

### LINKED ISSUE #442: incorrect wallet balances, ETH/EUR ICBM probably mixed up
I have 1000 EURT and 10 ETH in LockedAccount.
I also have 136 ETH in simple accounts

It seems you add EURT to ETH, see above

krzkaczor:
closes #442

### DIFF
--- a/app/modules/wallet/selectors.ts
+++ b/app/modules/wallet/selectors.ts
@@ -1,22 +1,5 @@
-import BigNumber from "bignumber.js";
+import { addBigNumbers, multiplyBigNumbers } from "../../utils/BigNumberUtils";
 import { IWalletState, IWalletStateData } from "./reducer";
-
-/**
- * helper for adding up array of big numbers
- */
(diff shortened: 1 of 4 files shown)


app/modules/wallet/selectors.ts
```diff
@@ -29,13 +29,16 @@ export const selectLockedWalletHasFunds = (state: IWalletStateData): boolean =>
  * ICBM Wallet Assets
  */
 export const selectICBMLockedEtherBalance = (state: IWalletStateData) =>
-  state.euroTokenICBMLockedBalance;
+  state.etherTokenICBMLockedBalance;
```
Moejoe90: :(

krzkaczor:
@Moejoe90 it uses WEIs in calculations. Data is kept as strings in WEI (no precision lost)  and it's converted to bignumbers when modified and back to strings.

PR #447, opened 2018-04-10
splits userAPI interfaces

Moejoe90:


app/lib/api/users/UsersApi.ts
```diff
@@ -17,8 +17,8 @@ export class UsersApi {
     @inject(symbols.logger) private logger: ILogger,
   ) {}
 
-  public async createAccount(newUser?: INewUser): Promise<IUser> {
-    this.logger.info("Creating account for email: ", newUser && newUser.unverifiedEmail);
+  public async createAccount(newUser?: IUserInput): Promise<IUser> {
+    this.logger.info("Creating account for email: ", newUser && newUser.newEmail);
```
krzkaczor: good that we fixed typings here

app/modules/auth/sagas.ts
```diff
@@ -75,8 +75,8 @@ export function* loadUser(): Iterator<any> {
   yield effects.put(actions.auth.loadUser(user));
 }
 
-export function* updateUser(updatedUser: IUser): Iterator<any> {
-  const user: IUser = yield neuCall(updateUserPromise, updatedUser);
+export function* updateUser(updatedUser: IUserInput): Iterator<any> {
+  const user: IUserInput = yield neuCall(updateUserPromise, updatedUser);
```
krzkaczor: it return IUser (take a look at updateUserPromise definition)

PR #449, opened 2018-04-11
Eto social profiles editor

lizonp:
screen shot 2018-04-11 at 10 43 05


app/components/Demo.tsx
```diff
@@ -362,5 +369,35 @@ export const Demo: React.SFC = () => (
         ]}
       />
     </Container>
+
+    <Container>
+      <SocialProfilesEditor
```
krzkaczor: i am under the impression that this will be reused always with the same set of services. Consider wraping it in another component like: `StandardSocialProfilesEditor`. We can as well postpone it till we need it again

PR #451, opened 2018-04-11
Tags editor

lizonp:
screen shot 2018-04-11 at 10 54 27

I have changed compiler target option to es2016 to get rid of `Property 'includes' does not exist on type 'string[]'.` warning.


app/components/eto/shared/TagsEditor.tsx
```diff
@@ -0,0 +1,128 @@
+import * as React from "react";
+
+import { Input } from "reactstrap";
```
krzkaczor: regroup imports according to guideliens please

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2015",
+    "target": "es2016",
```
krzkaczor: I think this was changed by accident

app/components/shared/Tag.tsx
```diff
@@ -12,21 +13,36 @@ export interface ITag {
   to?: string;
   layout?: TLayout;
   size?: TSize;
-  color?: TColor;
+  theme?: TTheme;
   className?: string;
+  onClick?: ((e: any) => void) | undefined;
```
krzkaczor: I think  `| undefined` is not needed since its optional

app/components/shared/Tag.tsx
```diff
@@ -12,21 +13,36 @@ export interface ITag {
...
     <>
       {to ? (
         <Link to={to} className={classes}>
+          {Boolean(svgIcon) && <InlineIcon svgIcon={svgIcon || ""} />}
           {text}
         </Link>
       ) : (
-        <span className={classes}>{text}</span>
+        <span onClick={onClick} className={classes}>
+          {Boolean(svgIcon) && <InlineIcon svgIcon={svgIcon || ""} />}
```
krzkaczor: why there is `|| ""`? since you do check for svgIcon to be thruthy anyway?

krzkaczor: I would propose using `!!value` instead of `Boolean(value)` I think it's more idiomatic in JS (and we already use it across the codebase)

app/components/eto/shared/TagsEditor.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    this.setState({ tags: [tag, ...tags] });
+    this.inputRef!.value = "";
+  }
+
+  private selectTag(e: { target: HTMLElement }): void {
+    const { selectedTagsLimit } = this.props;
+    const { tags, selectedTags } = this.state;
+    const tagElement: any = e.target.closest(this.tabComponentClassName);
```
krzkaczor: just pass selected tag as an argument to callback while rendering it

app/components/shared/Tag.tsx
Moejoe90: :+1:

tsconfig.json
lizonp: I have changed compiler target option to es2016. That was the only way I've figured out to eliminate `Property 'includes' does not exist on type 'string[]'.` warning.

krzkaczor: its not part of es2015. Please, just use includes from lodash

app/components/eto/shared/TagsEditor.tsx
```diff
@@ -0,0 +1,150 @@
...
+  state = {
+    selectedTags: [...this.props.selectedTags],
+    tags: [...this.props.availiableTags],
+    inputValue: "",
+  };
+
+  createTag = () => {
+    const { tags, selectedTags } = this.state;
+    const tag: string = this.state.inputValue.trim().replace(/\s\s+/g, " ");
+    const tagExist: boolean = includes(tags, tag) || includes(selectedTags, tag);
```
krzkaczor: these type annotations are not needed

app/components/eto/shared/TagsEditor.tsx
```diff
@@ -0,0 +1,150 @@
...
+  availiableTags: string[];
+}
+
+interface IStateWrapper {
+  selectedTags: string[];
+  tags: string[];
+  inputValue: string;
+}
+
+class TagsEditor extends React.Component<IProps> {
```
krzkaczor: this could be functional component

PR #452, opened 2018-04-11
Plug NEU value to NEU widget

krzkaczor:


app/components/dashboard/myPortfolio/MyPortfolioWidget.tsx
```diff
@@ -1,38 +1,83 @@
...
+  };
+}
+
+interface IStateProps extends IBodyProps {
+  isLoading: boolean;
+}
+
+type IProps = TOwnProps & IStateProps;
+
+export const MyPortfolioWidgetComponentBody: React.SFC<IBodyProps> = ({ error, data }) => {
```
Moejoe90: It would be cool maybe to add a new noncritical issue that covers the use of for example DarkPanel with an already loading and error handler like here

krzkaczor: Yeah, that's what i wrote on slack xD

PR #464, opened 2018-04-11
Adds sub headers to Settings

Moejoe90:


app/components/shared/SubHeader.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import * as styles from "./SubHeader.module.scss";
+
+interface ISubHeader {
+  view: string;
+}
+
+export const SubHeader: React.SFC<ISubHeader & React.HTMLAttributes<HTMLDivElement>> = ({
+  view,
+  ...props
```
Moejoe90: I would prefer to maybe take the props and ` React.HTMLAttributes<HTMLDivElement>> ` off as I think this component has a very specific case

app/components/shared/SubHeader.tsx
```diff
@@ -0,0 +1,20 @@
+import * as cn from "classnames";
+import * as React from "react";
+import { Col, Row } from "reactstrap";
+
+import * as styles from "./SubHeader.module.scss";
```
krzkaczor: component that does it should be done somewhere alread

Moejoe90: I just found from Patrik that there is indeed a component

PR #467, opened 2018-04-12
ETO users

krzkaczor:
routing for both investors and issuers, tweak interfaces to contain user type property


app/modules/auth/reducer.spec.ts
```diff
@@ -7,7 +7,9 @@ describe("auth > selectors", () => {
     it("should return true for authorized users", () => {
```
Moejoe90: Shouldn't the test somehow reflect that we are only dealing with investors no users anymore?

app/modules/auth/sagas.ts
```diff
@@ -39,10 +39,12 @@ export async function loadUserPromise({
       newEmail: walletMetadata.email,
       salt: walletMetadata.salt,
       backupCodesVerified: false,
+      type: "investor",
```
Moejoe90: Are you planning to register `issuers` in the same saga? or create a complete different saga for them

app/lib/api/users/UsersApi.ts
```diff
@@ -42,11 +42,13 @@ export class UsersApi {
...
     }
 
+    response.body.type = "investor"; // @todo remove it when api is done
+
     return response.body;
   }
 
   public async verifyUserEmail(userCode: IVerifyEmailUser): Promise<IUser> {
-    const response = await this.httpClient.put<IVerifyEmailUser>({
+    const response = await this.httpClient.put<IUser>({
```
Moejoe90: :+1:

app/modules/auth/reducer.spec.ts
krzkaczor: this selector doesnt use investor type at all sooo i guess its fine

app/modules/auth/sagas.ts
krzkaczor: probably the same saga — just paramettrized. i am gonna take a closer look when working on login

app/components/AppRouter.tsx
```diff
@@ -1,11 +1,17 @@
...
+
   root: "/",
+
   register: "/register",
+  etoRegister: "/register-eto",
   login: "/login",
-  kyc: "/kyc",
+  etoLogin: "/login-eto",
   recover: "/recover",
+  etoRecover: "/recover-eto",
```
huesforalice: Instinctively I would have "namespaced" this routes, so /eto/login and /investors/login. I think it looks nicer. Just a suggestion though.

app/components/AppRouter.tsx
```diff
@@ -38,18 +45,40 @@ export const AppRouter: React.SFC = () => (
...
-    <OnlyAuthorizedRoute path={appRoutes.wallet} component={Wallet} />
-    <OnlyAuthorizedRoute path={appRoutes.kyc} component={Kyc} />
-    <OnlyAuthorizedRoute path={appRoutes.dashboard} component={Dashboard} exact />
-    <OnlyAuthorizedRoute path={appRoutes.settings} component={Settings} exact />
-    <OnlyAuthorizedRoute path={settingsRoutes.seedBackup} component={BackupSeed} exact />
-    <OnlyAuthorizedRoute path={appRoutes.eto} component={Eto} />
+    {/* only investors routes */}
+    <OnlyAuthorizedRoute path={appRoutes.wallet} investorComponent={Wallet} />
+    <OnlyAuthorizedRoute path={appRoutes.kyc} investorComponent={Kyc} />
+
```
huesforalice: This is quite nice.

PR #475, opened 2018-04-16
Fixes Settings notifiaction

Moejoe90:


app/modules/notifications/selectors.ts
```diff
@@ -0,0 +1,14 @@
+import { IAppState } from "../../store";
+import { selectBackupCodesVerified, selectIsUserEmailVerified } from "../auth/selectors";
+import { selectKycRequestStatus } from "./../kyc/selectors";
+import { settingsNotification } from "./reducer";
+
+export const selectIsActionRequiredSettings = (state: IAppState): boolean =>
```
krzkaczor:
```
? true
   : false;
```
isn't this not needed? It will be a boolean expression anyway

Moejoe90: i see your point

PR #482, opened 2018-04-17
Issuer login

krzkaczor:
What's missing:
- [ ] recover routes for Issuers doesnt work
- [ ] both users use the same local storage
It will be done in separate PR.


app/components/walletSelector/WalletSelector.tsx
```diff
@@ -81,16 +90,14 @@ export const WalletSelectorComponent: React.SFC<IStateProps> = ({
...
     actionCreator: dispatch => dispatch(actions.walletSelector.reset()),
-    pure: false,
   }),
   appConnect<IStateProps>({
     stateToProps: s => ({
       isMessageSigning: s.walletSelector.isMessageSigning,
-      rootPath: selectIsLoginRoute(s.router) ? appRoutes.login : appRoutes.register,
+      rootPath: selectRootPath(s.router),
       isLoginRoute: selectIsLoginRoute(s.router),
+      userType: selectUrlUserType(s.router),
```
Moejoe90: :+1:

app/modules/auth/sagas.ts
```diff
@@ -90,7 +99,7 @@ function* logoutWatcher({ web3Manager, jwtStorage }: TGlobalDependencies): Itera
   yield effects.put(actions.init.start());
 }
 
-function* signInUser(): Iterator<any> {
+function* signInUser(_: TGlobalDependencies, { payload: { userType } }: any): Iterator<any> {
```
Moejoe90: What was `_` used for

app/modules/wallet-selector/selectors.ts
```diff
@@ -0,0 +1,25 @@
+import { RouterState } from "react-router-redux";
+import { appRoutes } from "../../components/AppRouter";
+import { TUserType } from "../../lib/api/users/interfaces";
+
+export const selectUrlUserType = (router: RouterState): TUserType =>
+  router.location && router.location.pathname.includes("eto") ? "issuer" : "investor";
```
Moejoe90: maybe this should be added as a note somewhere

app/modules/auth/sagas.ts
Moejoe90: Ah i had a similar problem before good to know

PR #495, opened 2018-04-18
Adds EtoProgressBar

Moejoe90:
image
Missing mobile designs


app/components/shared/InlineIcon.tsx
```diff
@@ -6,13 +6,20 @@ interface IProps {
...
+  svgIcon,
+  width,
+  height,
+  className = cn("inline-icon", styles.inlineIcon),
+  ...props
+}) => {
   return (
     <span
-      className={cn("inline-icon", styles.inlineIcon)}
+      className={className}
```
Moejoe90: Any suggestions how to better do this, its ugly like this

lizonp: Why not to add set span's className to `cn("inline-icon", styles.inlineIcon, className)`?

app/components/eto/shared/EtoProgressStepper.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const EtoProgressStepper: React.SFC<IEtoProgressStepper> = ({ currentStep, stepProps }) => (
+  <div className={styles.progressStepper}>
+    {range(stepProps.length).map(number => (
+      <div key={number} className="text-center" data-test-id="EtoProgressStepper-header-text">
+        {stepProps[number].isDone ? (
+          <InlineIcon
+            svgIcon={checkIcon}
+            height="11px"
+            width="15px"
+            className=""
```
lizonp: You have added className to InlineIcon component, but it's empty

app/components/shared/InlineIcon.tsx
Moejoe90: Because this style is breaking the style for this component I had to do a solution to clear this style while being backwards compatible
image

app/components/eto/shared/EtoProgressStepper.tsx
Moejoe90: Like explained above this is on purpose to clear the style for this component

app/components/Demo.tsx
```diff
@@ -438,5 +415,26 @@ export const Demo: React.SFC = () => (
...
       />
     </Container>
+    <Container>
+      <EtoProgressStepper
+        currentStep={2}
+        stepProps={[
+          {
+            name: "test",
+            isDone: true,
+            onClick: () => {
```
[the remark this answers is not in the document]
Moejoe90: fair enough, now its changed and uses index a parameter

app/components/shared/InlineIcon.tsx
lizonp: this shoud do the job
```:global(.text-center) {
  display: flex;
  flex-direction: column;
  justify-content: flex-end;
  align-items: center;
}```
then just add proper `margin-bottom` for `inline-icon` class

File: app/components/eto/shared/EtoProgressStepper.tsx (hunk 2 above)
Comment (krzkaczor) [reply app/components/eto/shared/EtoProgressStepper.tsx]: Yeah we really need to refactor InlineIcon — i think @Moejoe90 found the source of our problems. One of the selectors there is `&, svg` when it should be only `svg`.

Comment (krzkaczor) [reply app/components/eto/shared/EtoProgressStepper.tsx]: But lets wait with refactoring unless we have some nicer tests etc.

File: app/components/eto/shared/EtoProgressStepper.spec.tsx [hunk 4]
```diff
@@ -0,0 +1,50 @@
...
+      <EtoProgressStepper currentStep={1} stepProps={[mockData[0]]} onClick={spyFunction} />,
+    );
+    const component2 = shallow(
+      <EtoProgressStepper currentStep={1} stepProps={[mockData[1]]} onClick={spyFunction} />,
+    );
+
+    expect(component1.find(tid("check-icon")).length).to.be.eq(1);
+    expect(component2.find(tid("check-icon")).length).to.be.eq(0);
+  });
+  it("Should render name of step when isDone is false", () => {
```
krzkaczor: new line pls

app/components/shared/InlineIcon.tsx
Moejoe90: To be honest i don't think adding more styling to the component is a solution, as i think this will just put a band aid on the real problem, and it will bite us again in the future. 

I think the best way to go is refactor this component in such a way that makes it compatible using as less styling elements as possible

app/components/eto/shared/EtoProgressStepper.spec.tsx
Moejoe90: done

PR #504, opened 2018-04-20
Change Locked Wallet  to ICBM Walled  and add isLocked prop

lizonp:


app/components/wallet/views/start/Start.tsx
```diff
@@ -87,8 +88,9 @@ const WalletStartComponent: React.SFC<TProps> = props => (
       props.lockedWalletData!.hasFunds && (
         <Col lg={6} xs={12}>
           <WalletBalance
+            isLocked={true}
             className="h-100"
-            headerText="Locked Wallet"
+            headerText="ICBM Wallet"
```
krzkaczor: this one should stay named  as "Locked wallet"

PR #527, opened 2018-04-24
Add error label to select field

lizonp:
screen shot 2018-04-24 at 12 48 48


krzkaczor:
Could you tweak the error message to: "US citizens are not allowed" i think it makes more sense.

PR #533, opened 2018-04-24
Changes email activation logic

Moejoe90:


app/components/emailVerify.tsx
```diff
@@ -1,40 +1,24 @@
...
 }
 
-interface IEmailVerifyStateProps {
-  isVerified?: boolean;
-}
-
-export const emailVerifyComponent: React.SFC<
-  IEmailVerifyDispatchProps & IEmailVerifyStateProps
-> = ({ verifyEmail, isVerified, goHome }) => {
-  if (isVerified) {
```
krzkaczor: this kind of logic should not be in react component anyway 👍

app/modules/auth/sagas.ts
```diff
@@ -55,7 +58,14 @@ export async function verifyUserEmailPromise({
   getState,
   notificationCenter,
 }: TGlobalDependencies): Promise<void> {
-  const userCode = selectActivationCodeFromQueryString(getState().router);
+  const state = getState();
+  const userCode = selectActivationCodeFromQueryString(state.router);
+  const urlEmail = selectLightWalletEmailFromQueryString(state.router);
+  const verifiedEmail = selectVerifiedUserEmail(state.auth);
+  if (urlEmail === verifiedEmail) {
+    notificationCenter.error("Your email is already verified");
```
krzkaczor: i think it should be just info notification

app/modules/auth/sagas.ts
```diff
@@ -55,7 +58,14 @@ export async function verifyUserEmailPromise({
   getState,
   notificationCenter,
 }: TGlobalDependencies): Promise<void> {
-  const userCode = selectActivationCodeFromQueryString(getState().router);
+  const state = getState();
```
krzkaczor: you should never use getState injected function. Get data first in saga using select effect and pass it to promise returing function as argument. In this case extract fi

app/components/emailVerify.tsx
Moejoe90: True that's why i took it off

app/modules/auth/sagas.ts
Moejoe90: done

PR #534, opened 2018-04-24
Kk/intl setup

krzkaczor:
- add intl with react intl
- add `yarn intl:extract` task to extract all ids from react codebase and put them into `en-en.json`
- add `yarn start:remote` task to connect with remote backend


webpack/webpack.config.dev-remote.js
```diff
@@ -0,0 +1,44 @@
+const merge = require("webpack-merge");
```
Moejoe90: It would have been better i think if this was a separate PR but good idea in all cases

webpack/webpack.config.dev.js
```diff
@@ -112,12 +112,6 @@ module.exports = merge(configCommon, {
           {
             test: /\.(tsx?)$/,
             use: [
-              {
```
Moejoe90: I thought this was already taken off in the last PR?

krzkaczor: yeah but its not merged yet :/

Moejoe90: true hence why separate PRs are better ;)

PR #539, opened 2018-04-25
KYC translations

lizonp:


app/components/kyc/personal/Start.tsx
```diff
@@ -94,16 +92,16 @@ const KYCForm = (formikBag: FormikProps<IKycIndividualData> & IProps) => (
     <FormSelectField values={US_CITIZEN_VALUES} label="Are you a US citizen?" name="isUsCitizen" />
     <FormSelectField
       values={HIGH_INCOME_VALUES}
-      label="Is your income above 5.000 EUR monthly, or do you have more than 100k EUR of assets available?"
+      label={formatIntlMessage("kyc.personal.high-income")}
       name="isHighIncome"
     />
     <div className="p-4 text-center">
-      <Button type="submit" disabled={!formikBag.isValid || formikBag.loadingData}>
```
lizonp: @krzkaczor I am not sure how to properly pass formikBag object

app/components/kyc/personal/Start.tsx
```diff
@@ -19,33 +18,34 @@ import {
...
 import { Button } from "../../shared/Buttons";
 import { FormFieldDate } from "../../shared/forms/formField/FormFieldDate";
 import { Tooltip } from "../../shared/Tooltip";
 import { KycPanel } from "../KycPanel";
 
 const PEP_VALUES = {
-  [NONE_KEY]: "-please select-",
-  [BOOL_TRUE_KEY]: "Yes I am",
-  [BOOL_FALSE_KEY]: "No I am not",
+  [NONE_KEY]: <FormattedMessage id="form.select.please-select" />,
```
lizonp: FormattedMessage always returns swring wrapped in `<span>` tag. Because of this chagne `<span> cannot appear as a child of <option>` error apeared in console.

I beleive that setting `textComponent` property to React.Fragment on FormattedMessage shoud fix ithttps://github.com/yahoo/react-intl/wiki/Components#intlprovider

krzkaczor: hmm this is a interesting propsal but it can break some things so lets do it now but with a little bit coordination. let me read about these textComponent prop more

app/components/kyc/personal/Start.tsx
krzkaczor: You have renamed them to props above from what i see L62

app/components/shared/forms/formField/FormSelectField.tsx
```diff
@@ -35,7 +35,7 @@ export const unboolify = <T extends {}>(values: T): T => {
 interface IFieldGroup {
   label?: string | React.ReactNode;
   values: {
-    [key: string]: string;
+    [key: string]: any;
```
krzkaczor: could we make it something like string | React.Component? or React.Element i am not sure which is which now

lizonp: It's changed 👍

PR #542, opened 2018-04-25
Intl scripts improvement

krzkaczor:
- clears previous messages which could couse a buggy behaviour *sometimes*
- add check option to run on CI to fail a build with not extracted messages
- add info about addded/ deleated messages


app/components/walletSelector/WalletSelector.tsx
```diff
@@ -72,7 +79,10 @@ export const WalletSelectorComponent: React.SFC<IStateProps> = ({
           <Row className="mt-5">
             <Col xs={12} sm={6}>
               <span>
-                Having troubles with login? <Link to={appRoutes.recover}>Help</Link>
+                <FormattedHTMLMessage id="wallet-selector.help-link" />
```
Moejoe90: You should use `tagName="span"`

package.json
```diff
@@ -104,8 +104,8 @@
     "tslint:fix": "tslint --fix --project . --format stylish && tslint --fix --project . --format stylish --config ./tslint.spec.json '**/*.spec.ts?(x)' '**/*.spec.node.ts?(x)'",
     "tsc": "tsc --noEmit --project .",
     "sasslint": "sass-lint -c .sass-lint.yml -v -q",
-    "lint": "yarn format && yarn tslint && yarn sasslint && yarn tsc",
-    "lint:fix": "yarn tslint:fix && yarn format:fix && yarn sasslint && yarn tsc",
+    "lint": "yarn format && yarn tslint && yarn sasslint && yarn tsc && yarn intl:extract:check",
+    "lint:fix": "yarn tslint:fix && yarn format:fix && yarn sasslint && yarn tsc && yarn intl:extract",
```
Moejoe90: I'm a bit worried about adding `yarn intl:extract` to lint for as you might overwrite by mistake something you wanted to keep but that's just my opinion

scripts/mergeMessagesWithLocales.js
```diff
@@ -19,8 +19,11 @@ function stringifyAndSort(obj) {
 }
 
 function main() {
-  const defaultLocaleFileContents = readFileOrDefault(defaultLocalePath, "{}");
-  const defaultLocale = JSON.parse(defaultLocaleFileContents);
+  const checkMode = process.argv[2] === "--check";
```
lizonp: Wouldn't it be better to use contains here?

app/components/walletSelector/WalletSelector.tsx
krzkaczor: only last comment is part of this PR.

krzkaczor: but you're right :D i will fix it in another pr. acutally there should be just FormattedMessage used

package.json
krzkaczor: I am not sure about this either. Lets see how it performs and we can change it ofc in future.

PR #560, opened 2018-04-26
Fix bug that was preventing login

krzkaczor:


app/modules/init/sagas.ts
```diff
@@ -19,10 +19,16 @@ function* setup({ web3Manager, logger }: TGlobalDependencies): Iterator<any> {
 
     if (jwt) {
       if (isJwtExpiringLateEnough(jwt)) {
-        yield loadUser();
+        try {
+          yield loadUser();
+        } catch (e) {
+          yield cleanupAndLogoutSaga();
+          logger.error(
+            "Cannot retrieve account. This could happen b/c account was deleted on backend",
```
Moejoe90: Should there be information for the user

krzkaczor: nah. This is a internal thingy i guess...

PR #575, opened 2018-04-27
adds email to widget

Moejoe90:
image


app/components/settings/verifyEmail/VerifyEmailWidget.tsx
```diff
@@ -15,27 +16,115 @@ import { IIntlProps, injectIntlHelpers } from "../../../utils/injectIntlHelpers"
 import { Button } from "../../shared/Buttons";
 import { PanelDark } from "../../shared/PanelDark";
 
+import { Form, FormikProps, withFormik } from "formik";
+import * as Yup from "yup";
```
krzkaczor: imports order

app/components/settings/verifyEmail/VerifyEmailWidget.tsx
```diff
@@ -15,27 +16,115 @@ import { IIntlProps, injectIntlHelpers } from "../../../utils/injectIntlHelpers"
...
+}
+
+interface INoEMailUser {
+  addNewEmail: (values: { email: string }) => void;
+  isLocked?: boolean;
 }
 
+const SetEmailForm: React.SFC<IEnhancedFormProps & FormikProps<IFormValues>> = props => (
+  <Form className={cn(styles.content, "mt-0 pt-0 mb-0")}>
+    <FormField placeholder="Email address" name="email" />
```
krzkaczor: raw untranslated string

app/modules/settings/sagas.ts
```diff
@@ -30,8 +31,10 @@ export function* addNewEmail(
     yield effects.call(updateUser, { ...user, new_email: email });
     notificationCenter.info("New Email added");
   } catch {
-    notificationCenter.error("Failed to change email");
+    yield effects.call(loadUser);
+    notificationCenter.error("Failed to send email");
   }
+  yield effects.put(actions.verifyEmail.freeConnectedButton());
```
krzkaczor: i would prefer to put this into finally block so it's harder to accidentally omit it. (by using return for example in try or catch block

app/components/settings/verifyEmail/VerifyEmailWidget.tsx
```diff
@@ -4,38 +4,131 @@ import { FormattedMessage } from "react-intl";
...
+  isInitialValid: () => EmailFormSchema.isValidSync(false),
+  handleSubmit: (values, props) => props.props.handleSubmit(values),
+})(SetEmailForm);
+
+const NoEmailUser: React.SFC<INoEMailUser> = ({ addNewEmail, isLocked }) => (
+  <div className={styles.content}>
+    <p className={styles.customPaddingTop}>
+      <FormattedMessage id="settings.verify-email-widget.enter-email" />
+    </p>
+    <SetEmailEnhancedForm handleSubmit={addNewEmail} isLocked={isLocked} />
```
krzkaczor: i would just follow current naming convention and name it as isLoading (in state as well).

app/modules/verifyEmailWidget/reducer.ts
```diff
@@ -0,0 +1,31 @@
...
+        ...state,
+        isButtonLocked: false,
+      };
+  }
+
+  return state;
+};
+
+export const selectIsConnectedButtonLocked = (state: IVerifyEmailWidgetState): boolean =>
+  state && state.isButtonLocked;
```
krzkaczor: `state &&` is not needed

PR #577, opened 2018-04-27
kyc back buttons hotfix

lizonp:


app/components/kyc/start/Start.tsx
```diff
@@ -21,7 +21,7 @@ export const KYCStartComponent = injectIntlHelpers<IProps>(
       steps={4}
       currentStep={1}
       title={formatIntlMessage("kyc.start.title")}
-      hasBackButton={false}
+      backLink={""}
```
krzkaczor: just omit it entirely, right? its optional prop. I think we should rely on falsy values where its avoidable.

PR #587, opened 2018-05-02
KYC Widget fixes

krzkaczor:


app/components/settings/kycStates/KycStatusWidget.tsx
```diff
@@ -57,48 +65,61 @@ export const KycStatusWidgetComponent: React.SFC<IProps> = ({
...
   onGoToKycHome,
+  isLoading,
+  error,
 }) => {
   return (
     <PanelDark
       headerText="KYC PROCESS"
       rightComponent={
-        requestStatus === "Accepted" ? (
+        !isLoading &&
```
Moejoe90: So you don't want to display an icon during loading?

krzkaczor: Yup. We don't know which icon it should be.

Moejoe90: I assumed always red until proven otherwise

PR #589, opened 2018-05-02
Add translations for modals

lizonp:


app/components/modals/AccessLightWalletPrompt.tsx
```diff
@@ -30,10 +32,10 @@ const AccessLightWalletForm = (formikBag: FormikProps<IFormValues> & IProps) =>
     <div className="mt-3">
       <Button type="submit" disabled={!formikBag.values.password}>
         {" "}
-        Accept
```
krzkaczor: i wouldnt put every button translation into from.button. Please keep it in separate namespace as it is specific to given view.

Moejoe90: I thought we agreed to reuse these and call them general components i did the same thing in my PR

intl/locales/en-en.json
```diff
@@ -128,6 +128,11 @@
   "kyc.start.go-to-company": "I represent a company",
   "kyc.start.go-to-personal": "I represent myself",
   "kyc.start.title": "Start your KYC",
+  "modal.button.cancel": "Cancel",
+  "modal.button.dismiss": "Ok",
```
krzkaczor: use form.button.ok

intl/locales/en-en.json
```diff
@@ -128,6 +128,11 @@
   "kyc.start.go-to-company": "I represent a company",
   "kyc.start.go-to-personal": "I represent myself",
   "kyc.start.title": "Start your KYC",
+  "modal.button.cancel": "Cancel",
```
krzkaczor: use form.button.cancel

PR #593, opened 2018-05-02
Add Menu translations && help tab

lizonp:


app/components/layouts/LayoutAuthorizedMenu.tsx
```diff
@@ -31,17 +33,25 @@ interface IStateProps {
 
 const MenuEntry: React.SFC<IMenuEntry & NavLinkProps> = ({
   actionRequired,
-  title,
+  menuName,
   to,
   svgString,
 }) => {
-  return (
+  return /^https?:\/\//.test(to) ? (
```
krzkaczor: no inline regexes please. Extract it outside component as `const isAbsoluteLink = ...`

app/components/layouts/LayoutAuthorizedMenu.tsx
```diff
@@ -31,17 +33,25 @@ interface IStateProps {
...
+    <a href={to} target="_blank" className={cn(styles.menuItem, svgString)}>
+      <span className={styles.icon}>
```
krzkaczor: i think body of these two branches is the same — please extract it to another component

PR #616, opened 2018-05-03
adds wording fixes

Moejoe90:


.babelrc
```diff
@@ -5,7 +5,9 @@
       {
         "messagesDir": "./intl/messages/"
       }
-    ]
+    ],
+    ["@babel/plugin-proposal-decorators", { "legacy": true }],
```
krzkaczor: this is solved better by @lizonp PR. if he merges his PR first i will fix it.

app/config/.env.example
```diff
@@ -1,6 +1,6 @@
 NF_RPC_PROVIDER=https://localhost:9090/node
 NF_UNIVERSE_CONTRACT_ADDRESS="0xbe84036c11964e9743f056f4e780a99d302a77c4"
 
-NF_FEATURE_EMAIL_CHANGE_ENABLED=0
+NF_FEATURE_EMAIL_CHANGE_ENABLED=1
```
krzkaczor: i think we dont need that change upstream. Right? Just change it locally if you want to see this feature.

.babelrc
Moejoe90: This is from your pr, how do you recommend going forward with this?
@krzkaczor

app/config/.env.example
Moejoe90: You are correct

PR #637, opened 2018-05-05
Storybook integration

lizonp:
- [x] move all static components from demo page
- [x] enable image and fonts import
- [x] add react router integration (https://www.npmjs.com/package/storybook-react-router)


.babelrc
```diff
@@ -9,5 +9,5 @@
     ["@babel/plugin-proposal-decorators", { "legacy": true }],
     "@babel/plugin-proposal-optional-catch-binding"
```
krzkaczor: remove these plugins since (I THINK) they are part of preset typescript. Remove them with yarn remove as well.

app/components/eto/shared/TagsEditor.stories.tsx
```diff
@@ -0,0 +1,15 @@
+// tslint:disable-next-line:no-implicit-dependencies
```
krzkaczor: can you turn this rule globally and remove these comments. I did the same on some other branch that is still WIP. This rule is not really useful IMO.

PR #638, opened 2018-05-06
Improve ledger error handling [WIP]

banciur:


app/lib/web3/LightWallet.ts
```diff
@@ -207,6 +207,7 @@ export class LightWallet implements IPersonalWallet {
       // @see https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/index.md#torpcsig
       return ethSig.concatSig(rawSignedMsg.v, rawSignedMsg.r, rawSignedMsg.s);
     } catch (e) {
+      //TODO return SignerRejectConfirmationError which is more generic
```
banciur: Actually I'm not sure if we should return `SignerRejectConfirmationError` here. I should discuss it with @Moejoe90 or @krzkaczor

krzkaczor: I don't think so. Since it's gonna be rejected by clicking huge "reject" button. Here is more like internal problem during signing.

app/lib/web3/Web3Adapter.ts
```diff
@@ -54,6 +54,10 @@ export class Web3Adapter {
       from: address as string,
     });
 
+    if (resultData.error !== undefined) {
+      throw resultData.error;
```
krzkaczor: is it gonna be a error instance or what? if not we should wrap it like `throw new Error(resultData.error)`

app/modules/accessWallet/sagas.ts
```diff
@@ -117,7 +117,7 @@ export function* connectWalletAndRunEffect(effect: Effect | Iterator<Effect>): a
       if (isLightWallet) {
         yield call(unlockLightWallet);
       }
-      break;
+      return yield effect;
```
krzkaczor: soo this change shouldn't change behaviour right?

PR #641, opened 2018-05-07
Additional wording fixes

Moejoe90:


app/components/shared/Tooltip.tsx
```diff
@@ -9,13 +9,14 @@ import * as icon from "../../assets/img/inline_icons/icon_questionmark.svg";
 import * as styles from "./Tooltip.module.scss";
 
 interface IProps {
-  text: string;
+  text: string | React.ReactNode;
```
krzkaczor: if it doesnt work with HTML do not change the prop type then

Moejoe90: indeed

PR #654, opened 2018-05-09
Wording fixes

Moejoe90:


app/components/shared/InvestmentPreview.tsx
```diff
@@ -112,7 +112,7 @@ export const InvestmentPreview: React.SFC<IProps> = ({
                   {endInDays === 1 ? (
```
krzkaczor: this is not a correct way to pluralize translations. Please read on in docs: https://github.com/yahoo/react-intl

PR #685, opened 2018-05-14
Change default login for ETO page

lizonp:


app/components/walletSelector/WalletRouter.tsx
```diff
@@ -15,6 +15,6 @@ export const WalletRouter: React.SFC<IProps> = ({ rootPath }) => (
     <Route path={`${rootPath}/light`} component={WalletLight} />
     <Route path={`${rootPath}/browser`} component={WalletBrowser} exact />
     <Route path={`${rootPath}/ledger`} component={WalletLedger} exact />
-    <Redirect to={`${rootPath}/light`} />
+    <Redirect to={`${rootPath}/ledger`} />
```
krzkaczor: this should change only for issuers

app/components/walletSelector/WalletRouter.tsx
```diff
@@ -15,6 +17,11 @@ export const WalletRouter: React.SFC<IProps> = ({ rootPath }) => (
     <Route path={`${rootPath}/light`} component={WalletLight} />
     <Route path={`${rootPath}/browser`} component={WalletBrowser} exact />
     <Route path={`${rootPath}/ledger`} component={WalletLedger} exact />
+    <Route
+      path={`${appRoutes.loginEto}`}
```
krzkaczor: what about register ETO route? I think better solution is just to intoduce helper function like:
`getRedirectionUrl(rootPath) which based on rootPath will detect if its eto or investor route and then return either light wallet path or neufund wallet path.

app/config/.env.example
```diff
@@ -5,5 +5,6 @@ NF_CONTRACTS_NEW=0
 
 NF_FEATURE_EMAIL_CHANGE_ENABLED=0
 NF_USER_INFO_COMPONENT_ENABLED=0
+NF_ISSUERS_CAN_LOGIN_WITH_NEUFUND_WALLET=0
```
krzkaczor: Here it should be 1 so developers can easily run it in default env.

app/components/walletSelector/WalletRouter.tsx
```diff
@@ -2,9 +2,11 @@ import * as React from "react";
 import { Redirect, Route } from "react-router-dom";
 
 import { SwitchConnected } from "../shared/connectedRouting";
+
```
krzkaczor: delete empty lines to align it with guidelines

PR #693, opened 2018-05-15
Limit charts length @ datefield && update packages

lizonp:


app/components/shared/forms/formField/FormFieldDate.tsx
```diff
@@ -40,13 +40,16 @@ export class FormFieldDate extends React.Component<IProps> {
     year: "",
   };
 
+  monthInputRef = React.createRef();
```
krzkaczor: i guess its not used anywhere

PR #696, opened 2018-05-15
Improve process of signing message with ledger

banciur:


app/lib/web3/LedgerWallet.ts
```diff
@@ -261,6 +277,13 @@ export function parseLedgerError(error: any): LedgerError {
     return new LedgerConfirmationRejectedError();
   } else if (error.message !== undefined && error.message === "Invalid status 6a80") {
     return new LedgerContractsDisabledError();
+  } else if (
+    error.message !== undefined &&
```
krzkaczor: this is unnecassary.

Issue #718, opened 2018-05-16
If your country is a High Risk Country


meneramandarin:
this sentence should pop up:

"Unfortunately, we do not accept investors or companies coming from your country due to regulatory restrictions."

PR #731, opened 2018-05-17
[#718] Risk country error message

lizonp:
screen shot 2018-05-17 at 13 04 40


krzkaczor:
Yeah, I agree with @Moejoe90 that this should not be part of generic FormSelectField or at least there should be an option to select behavior.

app/components/shared/forms/formField/FormSelectField.tsx
```diff
@@ -97,7 +110,9 @@ export class FormSelectField extends React.Component<FieldGroupProps & IOwnProps
               onFocus={() => setFieldTouched(name, true)}
               type="select"
               value={field.value}
-              valid={isValid(touched, errors, name)}
+              valid={
+                includes(DISABLED_COUNTRIES, field.value) ? false : isValid(touched, errors, name)
```
Moejoe90: Do you still need this after your changes?

lizonp: You are right. I have removed it.

PR #734, opened 2018-05-17
fixes kyc progress stepper

Moejoe90:
image


krzkaczor:
Why business steps is not reused across all components 😱

PR #735, opened 2018-05-17
changes Nano Ledger to Ledger Nano

Moejoe90:


intl/locales/en-en.json
```diff
@@ -417,8 +417,8 @@
   "wallet-selector.register.help-link": "Having trouble with registering?",
   "wallet-selector.tabs.browser-wallet-login": "USE EXTERNAL WALLET",
   "wallet-selector.tabs.browser-wallet-register": "REGISTER WITH EXTERNAL WALLET",
-  "wallet-selector.tabs.ledger-login": "USE NANO LEDGER",
-  "wallet-selector.tabs.ledger-register": "REGISTER WITH NANO LEDGER",
+  "wallet-selector.tabs.ledger-login": "USE Ledger Nano",
+  "wallet-selector.tabs.ledger-register": "REGISTER WITH Ledger Nano",
```
banciur: I think in two above string you should use big letters

Issue #741, opened 2018-05-17
ICBM investors note


krzkaczor:
![image](https://user-images.githubusercontent.com/1814312/40186674-f95930de-59f5-11e8-9456-d54d4af1c574.png)

PR #745, opened 2018-05-17
Add info for ICBM investors

lizonp:
#741 

screen shot 2018-05-17 at 18 04 21


app/components/walletSelector/WalletSelector.tsx
```diff
@@ -79,6 +83,18 @@ export const WalletSelectorComponent: React.SFC<IStateProps & IIntlProps> = ({
               ])}
             />
           </Row>
+          {location.pathname === `${rootPath}/light` && (
```
krzkaczor: why not to just make it part of light wallet wallet component? (the one defined in routing).

PR #783, opened 2018-05-22
adds direct change to content management

Moejoe90:


app/index.tsx
```diff
@@ -41,6 +42,16 @@ function forceRerenderInDevMode(): number {
   }
 }
 
+if (process.env.NODE_ENV === "translate") {
+  const config = {
+    projectId: "6c987746f39a4c910f601940bba9b912",
```
krzkaczor: should we make it env variable?

typings/modules.d.ts
```diff
@@ -13,4 +13,6 @@ declare module "bitcore-mnemonic";
 declare module "eth-sig-util";
 declare module "tweetnacl";
 
+declare module "react-intl-phraseapp";
```
krzkaczor: We can't loose all type information. Remember problem with `FormattedHtmlMessage`? Please copy `react-intl` typings and put into `./@types/` dir.

webpack/webpack.config.translate.js
```diff
@@ -0,0 +1,19 @@
+const merge = require("webpack-merge");
+const devRemoteConfig = require("./webpack.config.dev-remote");
+
+const loadAppEnv = require("./loadAppEnv");
+const applicationEnv = loadAppEnv(process.env);
+
+if (!applicationEnv.NF_REMOTE_BACKEND_PROXY_ROOT) {
```
krzkaczor: i think all of this is not used

webpack/webpack.config.translate.js
```diff
@@ -0,0 +1,19 @@
...
+  throw new Error("Missing NF_REMOTE_BACKEND_PROXY_ROOT env variable. Add it to your .env file");
+}
+
+const backendUrl = JSON.parse(applicationEnv.NF_REMOTE_BACKEND_PROXY_ROOT);
```
krzkaczor: as well

.storybook/webpack.config.js
```diff
@@ -1,12 +1,19 @@
...
 const devConfig = require("../webpack/webpack.config.dev");
+const webpack = require("webpack");
 
 module.exports = (baseConfig, env, config) => {
   const pathToStyleLoader = path.join(__dirname, "./setup-styles.ts");
 
   config.entry.preview.push(pathToStyleLoader);
 
+  config.plugins.push(
+    new webpack.NormalModuleReplacementPlugin(/(.*)react-intl-phraseapp(\.*)/, data => {
```
krzkaczor: i would sleep better if this regex match only inside node modules

Moejoe90: It shouldn't match node_modules at all. this is used wherever `import "react-intl-phraseapp"` is called in the `./app`

typings/modules.d.ts
Moejoe90: The problem is it's not exactly the same interface some things are missing I can do some manual work and copy only what is there. 
but for sure the interface is different

app/index.tsx
Moejoe90: this would be better if it was a `env` variable indeed

.storybook/webpack.config.js
krzkaczor: soo why `(.*)`? I think it resolves to full path with node_modules in front, i am wrong then make regex more concise.

webpack/webpack.config.translate.js
Moejoe90: because it already taken from `devRemoteConfig` correct

webpack/webpack.config.translate.js
Moejoe90: done

typings/modules.d.ts
krzkaczor: Copy then only things that we use. Id should be required and id and tagName should be required for Html version of the component

Moejoe90: done

.storybook/webpack.config.js
Moejoe90: in documentation 
https://webpack.js.org/plugins/normal-module-replacement-plugin/
this is how they use the tool to change imports in files from how I see it it's covering the app directory

app/config/.env.example
```diff
@@ -2,6 +2,7 @@ NF_RPC_PROVIDER=https://localhost:9090/node
 NF_UNIVERSE_CONTRACT_ADDRESS="0xbe84036c11964e9743f056f4e780a99d302a77c4"
 NF_UNIVERSE_CONTRACT_ADDRESS_NEW="0x2795e3ac05f49db68e5b51a2643d99f00c1af6af"
 NF_CONTRACTS_NEW=0
+NF_TRANSLATION_ID="6c987746f39a4c910f601940bba9b912"
```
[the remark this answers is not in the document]
Moejoe90: This is a application ID number, 
Without having access to the project there is nothing you can do with this number. In their example they provided it publicly but it can be removed just to be safe

Issue #748, opened 2018-05-17
please change "web3 disconnected" popup via info that ledger nano was disconnected


glupiawymowka:
user disconnects his nano, this is what he can see

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/b80ee762-551b-4049-9f1a-3a7d9c7d9e86)

glupiawymowka:
@meneramandarin please suggest wording for this info

meneramandarin:
Your Nano Ledger was disconnected

meneramandarin:
For all others:
Your Browser Wallet was disconnected

glupiawymowka:
works on dev, doesn't work on staging

PR #784, opened 2018-05-23
[#748] Change "web3 disconnected" translation

lizonp:
#748


app/modules/web3/selectors.ts
```diff
@@ -49,6 +49,10 @@ export const selectIsLightWallet = (state: IWeb3State): boolean => {
   );
 };
 
+export const selectIsLedgerWallet = (state: IWeb3State): boolean => {
+  return state.connected && state.wallet.walletType === WalletType.LEDGER;
```
krzkaczor: this won't work correctly if wallet is disconnected.

krzkaczor: all such selectors need to work in both cases. When it's disconnected use: `previousConnectedWallet` data (take a look at: `isLightWalletReadyToLogin`) and please write tests for this selector!

app/modules/web3/flows.ts
```diff
@@ -21,10 +21,13 @@ export const web3Flows = {
 
       const state = getState();
       const isLightWallet = selectIsLightWallet(state.web3);
+      const isLedgerWallet = selectIsLedgerWallet(state.web3);
 
       if (!isLightWallet) {
```
krzkaczor: refactor it please into something more redable: behaviour should be as follows:
wallet is light wallet - no message
wallet is browser wallet - message `#1`
wallet is ledger wallet - message `#2`

Use `switch case ` or just not nested ifs.

app/modules/web3/selectors.ts
lizonp: I've removed it

app/modules/web3/flows.ts
```diff
@@ -20,12 +21,23 @@ export const web3Flows = {
...
 
       const state = getState();
-      const isLightWallet = selectIsLightWallet(state.web3);
+      const walletType = selectWalletType(state.web3);
 
-      if (!isLightWallet) {
-        notificationCenter.error(
-          intlWrapper.intl.formatIntlMessage("modules.web3.flows.web3-error"),
-        );
+      if (!state.web3.connected) {
```
krzkaczor: what is the resoning for this if statement? this function is called: `personalWalletDisconnected` so i guess it's always true. Before there wasn't any condition like this.

lizonp: You are right, it's redundant. I have removed it

app/modules/web3/flows.ts
```diff
@@ -22,23 +22,20 @@ export const web3Flows = {
...
+          error = intlWrapper.intl.formatIntlMessage("modules.web3.flows.web3-error.browser");
+          break;
+        case WalletType.LEDGER:
+          error = intlWrapper.intl.formatIntlMessage("modules.web3.flows.web3-error.ledger");
+          break;
+        default:
+          return;
       }
+
+      notificationCenter.error(error);
```
krzkaczor: i think this would post a message when `error === ""` please add an if statement.

app/modules/web3/flows.ts
```diff
@@ -22,23 +22,20 @@ export const web3Flows = {
 
       const state = getState();
       const walletType = selectWalletType(state.web3);
+      let error = "";
```
krzkaczor: it would be cool to export `switch` statement  to another function and then you can use return instead of `let`. We should use it very sparingly. The function should return string | undefined; and then you can only do `notificationCenter.error` when `if(error)`

app/modules/web3/flows.ts
krzkaczor: we shall have test for this....

app/modules/web3/flows.ts
```diff
@@ -20,12 +21,22 @@ export const web3Flows = {
...
-          intlWrapper.intl.formatIntlMessage("modules.web3.flows.web3-error"),
-        );
+      const disconnectedWalletErrorMessage = () => {
+        switch (selectWalletType(state.web3)) {
+          case WalletType.BROWSER:
+            return intlWrapper.intl.formatIntlMessage("modules.web3.flows.web3-error.browser");
+          case WalletType.LEDGER:
+            return intlWrapper.intl.formatIntlMessage("modules.web3.flows.web3-error.ledger");
+          default:
+            return "";
```
krzkaczor: return undefined

app/modules/web3/flows.ts
```diff
@@ -20,12 +21,22 @@ export const web3Flows = {
...
+        }
+      };
+
+      const message = disconnectedWalletErrorMessage();
+
+      if (message.length) {
```
krzkaczor: if (message)

app/modules/web3/__tests__/flows.spec.ts
```diff
@@ -4,25 +4,51 @@ import { spy } from "sinon";
...
 import { dummyIntl } from "./../../../utils/injectIntlHelpers.fixtures";
 import { getDummyLightWalletMetadata } from "./fixtures";
 
 describe("web3 > flows", () => {
   describe("personalWalletDisconnected", () => {
-    it("should send notification if it's not light wallet", () => {
+    it("should send notification if it's ledger wallet", () => {
       const dummyDispatch = spy();
-      const state: Partial<IAppState> = {
+      const state: Partial<any> = {
```
krzkaczor: `Partial<any>` doesnt make any sense. Either use any or type it properly as Partial<IAppState> — this is what i would prefer. just add missing data to previousConnectedWallet structure to match it with interface. The same problem is repeated in next test cases.

app/modules/web3/__tests__/fixtures.ts
```diff
@@ -1,11 +1,37 @@
...
-export const getDummyLightWalletMetadata = (): ILightWalletMetadata => ({
-  walletType: WalletType.LIGHT,
+const walletsCommon = {
   address: dummyEthereumAddress,
   vault: "vault",
   email: "test@example.com",
   salt: "salt",
+};
+
+interface IBrowserWalletMetadata extends ICommonWalletMetadata {
```
krzkaczor: this is not currect. Browser wallet metada requires just address. So it's simply ICommonWalletMetadata

app/modules/web3/__tests__/fixtures.ts
```diff
@@ -1,11 +1,37 @@
...
+import {
+  ICommonWalletMetadata,
+  ILedgerWalletMetadata,
+  ILightWalletMetadata,
+} from "../../../lib/persistence/WalletMetadataObjectStorage";
 import { WalletType } from "../types";
 
-export const getDummyLightWalletMetadata = (): ILightWalletMetadata => ({
-  walletType: WalletType.LIGHT,
+const walletsCommon = {
```
krzkaczor: this should be called something like lightWalletMetadata

PR #803, opened 2018-05-24
Moe/ Add plurals to Investment Preview

Moejoe90:


app/components/shared/InvestmentPreview.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  },
+  hasStarted: false,
+  startingOn: "19/10/2020",
+  moneyGoal: "5.000.000",
+  currentValuation: "10.000.000",
+  tokenPrice: "0.50",
+  linkToDetails: "http://localhost:9001",
+  handleEmailSend: () => {},
+};
+storiesOf("InvestmentPreview", module)
```
krzkaczor: new line bitte

app/components/shared/InvestmentPreview.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  .add("ETO didn't start", () => <InvestmentPreview {...defaultProps} />)
+  .add("ETO ends in 1 day", () => <InvestmentPreview {...defaultProps} {...{ hasStarted: true }} />)
```
krzkaczor: ` {...{ hasStarted: true }}` ??? use just `hasStarted={true}` or hasStarted`

app/components/shared/InvestmentPreview.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+  .add("ETO ends in more day", () => (
+    <InvestmentPreview {...defaultProps} {...{ hasStarted: true, endInDays: 20 }} />
```
krzkaczor: same

app/components/shared/InvestmentPreview.tsx
```diff
@@ -6,6 +6,7 @@ import { Money } from "./Money";
 import { PercentageIndicatorBar } from "./PercentageIndicatorBar";
 import { ITag, Tag } from "./Tag";
 
+import { FormattedPlural } from "react-intl";
```
krzkaczor: what about using `react-intl-phraseapp`

app/components/shared/InvestmentPreview.tsx
```diff
@@ -109,26 +110,15 @@ export const InvestmentPreview: React.SFC<IProps> = ({
...
-                  )}
+                  <FormattedHTMLMessage
+                    tagName="span"
+                    id="shared-component.investment-preview.investment-timeline"
+                    values={{
+                      endInDays: endInDays,
+                      neuInvestors: neuInvestorsNum,
+                    }}
+                  />{" "}
+                  <FormattedPlural value={endInDays} one="day" other="days" />
```
krzkaczor: `one="day" other="days"` i think this is not a correct way to use it since it's nor part of translations file. Instead of passing strings you need to pass <FormattedMessage /> elements

intl/locales/en-en.json
```diff
@@ -333,8 +333,7 @@
   "shared-component.investment-preview.details": "Details",
   "shared-component.investment-preview.get-notification": "Get notification email",
   "shared-component.investment-preview.goal": "Hard cap",
-  "shared-component.investment-preview.investment-timeline": "NEU Investors <strong>{neuInvestors}</strong> | Ends in <strong>{endInDays}</strong> days",
-  "shared-component.investment-preview.investment-timeline-one-day": "NEU Investors <strong>{neuInvestors}</strong> | Ends in <strong>{endInDays}</strong> day",
+  "shared-component.investment-preview.investment-timeline": "NEU Investors <strong>{neuInvestors}</strong> | Ends in <strong>{endInDays}</strong>",
```
krzkaczor: i am not sure if we should have partial translations like this. I would imagine it that `days` or `day` translation is a parameter. So whole string would be:   

`NEU Investors <strong>{neuInvestors}</strong> | Ends in <strong>{endInDays}</strong> {daysTranslation}` and then you inject formattedPlurarl as a value

app/components/shared/InvestmentPreview.tsx
Moejoe90: We can't use `react-intl-phraseapp` they don't have an implementation of that

krzkaczor: That's fine actually 👍  We will be fine just by using FormattedMessage as args

app/components/shared/InvestmentPreview.stories.tsx
Moejoe90: done

app/components/shared/InvestmentPreview.stories.tsx
Moejoe90: :+1:

app/components/shared/InvestmentPreview.stories.tsx
Moejoe90: :+1:

Issue #814, opened 2018-05-28
Slider Component


SergejSoft:
pls. add to story board.
![Screen Shot 2018-05-28 at 12.25.28.png](https://images.zenhubusercontent.com/57dd639be40e5714b16f0753/bcb35209-4a1b-40ad-b347-44881721a25f)

PR #816, opened 2018-05-29
#814 Range component

lizonp:
issue: #814 
storybook: http://localhost:9001/?selectedKind=Range&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=intl%2Fpanel


app/components/shared/Range.stories.tsx
```diff
@@ -0,0 +1,13 @@
+import { storiesOf } from "@storybook/react";
+import * as React from "react";
+
+import { Range } from "./Range";
+
+storiesOf("Range", module)
+  .add("default", () => <Range name="name" min={10} max={20} unit={"%"} />)
```
krzkaczor: you dont need `{` around strings. so it should be just: `unit="%"`
the same applies to next lines

lizonp: It's updated

Issue #505, opened 2018-04-20
Issuer dashboard | two tabs. Book building, Investment


SergejSoft:
We will use Overview page for both Book Building stats and Investment stats. 
In the future, we will have some charts that show estimate vs real investment stats. 
<img width="1111" alt="tabs" src="https://user-images.githubusercontent.com/2290474/39047748-51329982-449b-11e8-8885-b736bbe14e60.png">

PR #821, opened 2018-05-30
#505 Add extra size && stories for Tabs component

lizonp:
issue: #505 
stories: http://localhost:9001/?selectedKind=Tabs&selectedStory=large&full=0&addons=1&stories=1&panelRight=0&addonPanel=intl%2Fpanel


app/components/shared/Tabs.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+const tabs = [
+  { text: "first tab active by default", path: "/" },
+  { text: "second tab", path: "/rand" },
+  { text: "third tab", path: "/om" },
+];
+
+storiesOf("Tabs", module)
+  .addDecorator(withStore())
+  .add("default", () => <Tabs tabs={tabs} />)
+  .add("theme dark", () => <Tabs theme="dark" tabs={tabs} />)
```
krzkaczor: i cant spot any difference between default theme and dark theme....

lizonp: You are right, it should be 'light'

PR #831, opened 2018-05-31
#826 Add form section && checkbox

lizonp:
issue: #826 
checkbox stories: http://localhost:9001/?selectedKind=FormSection&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=intl%2Fpanel
form section stories: http://localhost:9001/?selectedKind=FormSection&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=intl%2Fpanel


app/components/shared/forms/formField/FormCheckbox.tsx
```diff
@@ -0,0 +1,52 @@
...
+export class FormCheckbox extends React.Component<IProps> {
+  static contextTypes = {
+    formik: PropTypes.object,
+  };
+
+  render(): React.ReactNode {
+    const { name } = this.props;
+
+    return (
+      // TODO: add form validation if needed
```
krzkaczor: if designs are consitent in terms of how checkbox's label should be shown then here instead of leaving TODO render a label

app/components/shared/forms/formField/FormCheckbox.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import { storiesOf } from "@storybook/react";
+import { Form, Formik } from "formik";
+import * as React from "react";
+
+import { FormCheckbox } from "./FormCheckbox";
+
+const formWrapper = (formState: any) => (Component: React.SFC) => () => (
```
krzkaczor: i think this is the same as the other form wrapper. Please do not copy code! Just extract it to shared file and reuse it.

app/components/shared/forms/formField/FormCheckbox.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+        <Component />
+      </Form>
+    )}
+  </Formik>
+);
+
+storiesOf("FormCheckbox", module)
+  .add(
+    "checkbox",
+    formWrapper({ name: "checkbox" })(() => (
```
krzkaczor: here the idea is that you pass initial state. "name" should be a name of the field so for example test things like what happens when one of the fields is sellected:
`{ checkbox: true }`

checkbox is the current name of the field because: `name="checkbox"`

app/components/shared/forms/formField/FormCheckbox.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <FormCheckbox type="checkbox" label="checkbox's label" name="checkbox" />
+    )),
+  )
+  .add(
+    "radio",
+    formWrapper({ name: "checkbox" })(() => (
+      <>
+        <FormCheckbox type="radio" label="checkbox's label" name="radio" />
```
krzkaczor: please check how does it look in formik's state b/c i have no idea if this will work.

PR #895, opened 2018-06-12
Adds additional ETO Api

Moejoe90:


app/lib/api/EtoApi.interfaces.ts
```diff
@@ -87,5 +95,56 @@ export const EtoLegalInformationType = YupTS.object({
...
+  whitelistDurationDays: YupTS.number(),
+});
+
+export type TEtoSpecsData = YupTS.TypeOf<typeof EtoSpecsInformationType>;
+
+/*General Interfaces */
+export type TPartialEtoSpecData = DeepPartial<TEtoSpecsData>;
+export type TPartialCompanyEtoData = DeepPartial<TCompanyEtoData>;
+
+export type TFullEtoData = {
```
krzkaczor: this name is a little bit misleading since it's something like full but partial eto data

app/components/eto/registration/pages/EtoTerms.tsx
```diff
@@ -271,11 +271,11 @@ export const EtoRegistrationTerms = compose<React.SFC>(
     stateToProps: s => ({
       loadingData: s.etoFlow.loading,
       savingData: s.etoFlow.saving,
-      stateValues: s.etoFlow.data,
+      stateValues: s.etoFlow.companyData,
     }),
     dispatchToProps: dispatch => ({
       saveData: (data: any) => {
```
krzkaczor: we know here what kind of data we are gonna save so please do not use any here. Could you test it in other forms as well. It should be TPartialCompanyEtoData or even smaller type dervided from validator

app/lib/api/EtoApi.interfaces.ts
Moejoe90: renamed to `TGeneralEtoData`

app/components/eto/registration/pages/EtoTerms.tsx
Moejoe90: `TPartialCompanyEtoData` 
When using a small type derived from validator typescript complained i assume the whole data is sent always

Issue #889, opened 2018-06-11
[Neufund Wallet - recovery] salt e-mail is not sent after successful recovery


rudolfix:
after successful recovery from backup phrases I provided verified e-mail when changing password. at the end of the process, salt was not sent in the e-mail.

It seems that there is no call to /user/me to send a new salt. IMO recovery context is not passed to the JWT login and later and this is just lost in the flow

we should always send new salt to e-mail provided during recovery. there are following cases

1. new account was created see #864 
2. existing account was recovered - we always send salt to e-mail provided in recovery. in case active e-mail was provided, activation link should not be sent but salt link should always be sent.

all cases in (2) should be dealt by backend. you just need to send new unverified e-mail and salt. backend should do the rest.

I'm putting @Exef here to be sure we cover all those cases
1. trying to activate e-mail that is already activated (we just ignore activation but we send salt)
2. trying to activate new e-mail when other e-mail is already activated (we send salt and activation code, old activated e-mail will be removed when new one is activated)
3. trying to activate new e-mail when no e-mail is activated (we send salt and activation code)

Issue #864, opened 2018-06-06
[Neufund Wallet - recovery] email address input not working correctly 


glupiawymowka:
1. go to "Recover wallet" section on login page
2. put all the recovery phrases
3. you are asked to put email & 2x password -> put an email that was not used for this account before 
4. there is no error nor information that you should use the registered email 
5. go to settings
6. old email is shown
7. no email comes to the new email address you provided

ALSO

8. logoff
9. go to login page again 
10. NEW email is shown 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/46280764-b244-4564-9286-41c6b0e88f96)
11. login & go to settings
12. old email is shown 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/339cc249-fdf3-4ebd-b0a1-8ecbac3de164)

It might be very confusing for a user that would think he changed his email address here

PR #905, opened 2018-06-15
Moe/light recover bug fix

Moejoe90:
This solves the flow where the user recovers their wallet and enters another email 
#864 related
#889 related 
#760 related

An e2e test that checks for salted email and recovered eth address was added


Moejoe90:
@krzkaczor 
the reason why it's not translated is that translation is injected via `intl` instead of directly using `<Formmatedtext>`

app/components/walletSelector/walletRecover/__tests__/WalletRecover.spec.e2e.ts
```diff
@@ -28,7 +30,9 @@ describe("Wallet recover", () => {
     "strike",
   ];
 
-  it("should recover wallet from saved phrases", () => {
+  const generatedAddress = "0x429123b08df32b0006fd1f3b0ef893a8993802f3";
```
krzkaczor: please rename it to something like expectedGeneratedAddress to make it more obvious

PR #946, opened 2018-06-26
Revert color changes

lizonp:


app/styles/overrides.scss
```diff
@@ -41,9 +41,9 @@ textarea {
   padding: 0 px-to-rem(10px) px-to-rem(4px);
   background-color: transparent;
 
-  &::placeholder {
-    color: inherit;
-  }
+  // &::placeholder {
```
krzkaczor: remove if not needed

PR #1059, opened 2018-07-18
investment flow views

trival:


app/components/dashboard/investment-flow/Investment.tsx
```diff
@@ -0,0 +1,112 @@
+import { Form, FormikProps } from "formik";
```
lizonp: Please sort imports in all files as shown at `react-guidelines.md`

app/components/dashboard/investment-flow/Investment.module.scss
```diff
@@ -0,0 +1,42 @@
+.container {
+  :global(.row) {
```
lizonp: I am not sure if it wouldn't be better to use `my-<>` bootstrap class or one of `$spacing-vertical-<>` variable. We don't have a design system nor guidelines on that, so it's just an opinion  🤷🏻‍♂️

app/components/dashboard/investment-flow/Investment.module.scss
```diff
@@ -0,0 +1,42 @@
...
+    margin: 1.5rem 0;
+  }
+}
+
+.invest-input {
+  &:global(.form-group) {
+    margin-bottom: 0;
+  }
+
+  label {
```
lizonp: I think we should be styling elemnts basing on classes not tag. Again it's just an opinion.

app/components/shared/modals/Modals.module.scss
```diff
@@ -0,0 +1,18 @@
+.heading {
```
lizonp: it seems to be similar to `SectionHeader` component

app/styles/variables.scss
```diff
@@ -49,6 +49,7 @@ $neu-cloudy-blue: #aabcca;
 $neu-light-blue-grey: #d8e0eb;
 $neu-red: #ed0000;
 $neu-white-two: #fff;
+$neu-blue: #4a90e2;
```
lizonp: We should be unifying colors pallet instead of adding new colors.

app/components/shared/modals/TransactionState.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface IProps {
+  state: TransactionStateType;
+  txHash?: string;
+  blockNumber?: number;
+  title?: TTranslatedString;
+}
+
+export const TransactionState: React.SFC<IProps> = props => {
+  let animation: React.ReactNode;
```
lizonp: THis could be changed to the function

app/components/dashboard/investment-flow/Investment.tsx
trival: ok, thanks, I havn't seen these guidelines yet, and i thought that some lint or prettier tasks would do that automatically. I think these are mostly automatic imports from vscode. I will check that out

app/styles/variables.scss
trival: totally agree!

app/components/shared/modals/Modals.module.scss
trival: Oh, havn't seen that. The Panel has a similar styling, without a extracted component, so I thought It wasn't there yet

app/components/shared/modals/TransactionState.tsx
trival: don't understand. you mean the switch statement?

app/components/dashboard/investment-flow/Investment.module.scss
trival: fair enought!

app/components/dashboard/investment-flow/Investment.module.scss
trival: Yes, I havn't been able to recognize a unified design system, but would be happy to talk to you about it more next week! Will have a look at your classes in more detail in the future.

app/components/shared/modals/TransactionState.tsx
lizonp: Yes, instead of assingnig react elements to varaible. It's good enough, and it's matter of preference I think :)

Issue #1088, opened 2018-07-24
Replace 'Neufund' text with graphic in nav bar of platform 


falinebambi:
The ‘Neufund’ word needs to be updated to this graphic as it does’t render correctly on all screens![neufund-logo.svg](https://images.zenhubusercontent.com/57c01da2e40e5714b16dddf9/a39386e5-896e-4560-8e2b-dc2e7aff0a0e)

PR #1089, opened 2018-07-24
replace neufund logo in header

trival:
https://github.com/Neufund/platform-frontend/issues/1088


app/components/Header.module.scss
```diff
@@ -9,21 +9,20 @@
...
   position: relative;
-  top: 3px;
-  left: 8px;
+  left: 1rem;
 }
 
 .bar {
   height: px-to-rem(60px);
   background-color: $neu-charcoal-grey;
   z-index: 1000;
```
lizonp: please use one of values from `z-index.scss` file

PR #1090, opened 2018-07-24
progress calculation performance

trival:
I made some improvements on performance of the eto progress calculation.

The main performance issue with the forms is not the progress calculation, it is formik. After each input, it rerenders everything, then validates everything with YUP, then rerenders again.

see
https://github.com/jaredpalmer/formik/pull/422
and 
https://github.com/jaredpalmer/formik/issues/430

That issue is solved with formik version 1.0.0. The version that we use has no simple solution for this. shouldComponentUpdate has no effect, since form values are propagated by context, not by props...

To get better form performance, we should update the library, which would be a different PR, i guess


app/components/eto/registration/EtoFormBase.tsx
```diff
@@ -23,15 +28,19 @@ class PercentageFormDone extends React.Component<IFormPercentageDoneProps> {
...
-    const { values } = this.context.formik as FormikProps<any>;
+  calculate: ProgressCalculator;
 
-    const calculatedFraction = selectFormFractionDone(
-      this.props.validator,
-      values,
-      this.props.progressOptions,
+  constructor(props: IFormPercentageDoneProps) {
+    super(props);
+    this.calculate = throttle(
```
krzkaczor: please use class field. like `calculate = throttle(...` on the class level. Then you don't need constructor at all

trival: I need props.validator and props.progressOptions at this point, so I guess we cannot ommit the constructor here

PR #1095, opened 2018-07-25
fix remove button on key individuals section. Fix arrayHelpers usage

krzkaczor:
Fixes remove button on forms

### DIFF
--- a/app/components/eto/registration/pages/KeyIndividuals.tsx
+++ b/app/components/eto/registration/pages/KeyIndividuals.tsx
@@ -51,12 +51,12 @@ interface IKeyIndividualsGroup {
   title: TTranslatedString;
 }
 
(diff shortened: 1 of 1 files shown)


app/components/eto/registration/pages/KeyIndividuals.tsx
```diff
@@ -51,12 +51,12 @@ interface IKeyIndividualsGroup {
   title: TTranslatedString;
 }
 
-const blankMember = {
+const getBlankMember = () => ({
```
Moejoe90: Whats the difference between assigning blankMember directly or wrapping it around a function call?

krzkaczor: every time you will get new object (new reference)

Issue #1091, opened 2018-07-25
Madia links |Additional documents.


SergejSoft:
There one more input section at the end of the form to add aditional documents via link:
Design:
https://zpl.io/2y7xYpL
Section title: "optional campaigning links"

This links: Google doc links, Dropbox links will then be displayed in the Public view under document section: "Marketing documents"
Design:
https://zpl.io/bAqgW8O

Issue #933, opened 2018-06-21
Enable Twitter Feed. Public View


SergejSoft:
https://developer.twitter.com/en/docs/twitter-for-websites/timelines/overview.html

On the ETO Media link form, we have a checkbox "Disable twitter feed"
So by default it is always enabled.
twitter feed is visible in Public View Page: Right section.
https://zpl.io/bAqgW8O

This shell activates company twitter feed. 

TO DO:
- [ ] Move Twitter icon in Media form to the left. Make a first from the left.
- [ ] Make it open by default. See mock up.
- [ ] Enable twitter feed and display in Public View as soon as the user enter the twitter link.
- [ ] If YouTube link is not filled. The top section is taken by twitter feed with 2 Feeds. 
https://zpl.io/bWYnGXn

SergejSoft:
@rudolfix we need some clarification how to make this happened form your views.

PR #1108, opened 2018-07-26
Tg/twitter and marketing documents

trival:
* add twitter widget, inside an new iframe container https://github.com/Neufund/platform-frontend/issues/933
* add marketing links https://github.com/Neufund/platform-frontend/issues/1091
* add proper typing in EtoPublicComponent


app/components/eto/EtoPreview.tsx
```diff
@@ -42,7 +46,10 @@ class EtoPreviewComponent extends React.Component<IProps> {
 
     return (
       <LayoutAuthorized>
-        <EtoPublicComponent companyData={this.props.companyData} etoData={this.props.etoData} />
+        <EtoPublicComponent
+          companyData={this.props.companyData as TCompanyEtoData}
```
Moejoe90: Can you add a todo here to remove this casting?

app/components/eto/shared/EtoPublicComponent.tsx
```diff
@@ -142,7 +147,22 @@ export const EtoPublicComponent: React.SFC<IProps> = ({ companyData, etoData })
...
-  const isYouTubeVideoAvailable = companyVideo !== undefined ? companyVideo.url.length : false;
+  const isYouTubeVideoAvailable = companyVideo && companyVideo.url && companyVideo.url.length;
+  const twitterUrl =
+    isTwitterFeedEnabled && socialChannels
+      ? (socialChannels.find(c => c.type === "twitter") as any).url
+      : "";
+
+  const marketingLinks = companyData.marketingLinks && {
+    documents: companyData.marketingLinks.map(l => ({
+      url: l.url as string,
```
Moejoe90: also here todo for typings

app/components/eto/shared/EtoPublicComponent.tsx
```diff
@@ -142,7 +150,22 @@ export const EtoPublicComponent: React.SFC<IProps> = ({ companyData, etoData })
   const isTwitterFeedEnabled =
     some(socialChannels, (channel: any) => channel.type === "twitter" && channel.url.length) &&
     !disableTwitterFeed;
-  const isYouTubeVideoAvailable = companyVideo !== undefined ? companyVideo.url.length : false;
+  const isYouTubeVideoAvailable = companyVideo && companyVideo.url && companyVideo.url.length;
```
krzkaczor: This should be boolean, no? use `const isYouTubeVideoAvailable = companyVideo && companyVideo.url && companyVideo.url.length > 0;
`

app/twitter-iframe.html
```diff
@@ -0,0 +1,22 @@
+
```
krzkaczor: why leading new line?

app/twitter-iframe.html
```diff
@@ -0,0 +1,22 @@
...
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+        <title>Neufund Twitter Iframe</title>
+    </head>
+    <body>
+        <a id="twitter" class="twitter-timeline" data-chrome="noheader nofooter noborders transparent"></a>
+        <script>
+            var a = document.getElementById('twitter')
```
krzkaczor: use const everywhere or var

app/twitter-iframe.html
```diff
@@ -0,0 +1,22 @@
...
+            var search = location.search.substring(1);
+            var query = JSON.parse('{"' + decodeURI(search).replace(/"/g, '\\"').replace(/&/g, '","').replace(/=/g, '":"') + '"}')
+            a.href = query.url
+            a.innerText = "Tweets by " + query.user
+            const s = document.createElement('script')
```
krzkaczor: can't you just create script tag after that one? instead of creating it dynamically?

app/components/eto/shared/EtoPublicComponent.tsx
trival: good catch!

app/twitter-iframe.html
trival: you think this is ok? This code is not transpiled, so I wanted to use var to make it as compatible as possible. Actually I forgot the const at the script tag. But if you are ok, with const, I will happily do it, cause I generally prefer it as well.

PR #1137, opened 2018-07-31
Adds IPFS modal image

Moejoe90:
image
image


app/components/eto/shared/EtoFileIpfsModal.module.scss
```diff
@@ -7,3 +7,12 @@
   @include make-col-ready;
   text-align: center;
 }
+
+.image {
```
lizonp: @moe Couldn't you use <ResponsiveImage/> component with set `max-width`?

Moejoe90: `<ResponsiveImage/>` was exactly what i was looking for

Issue #1049, opened 2018-07-16
send correct wallet type and subtype to the backend


rudolfix:
on PUT and POST to /users/me:
send the following properties of the user "object"
```
      wallet_type:
        example: browser
        type: string
        enum: [ledger, browser, light, unknown]
      wallet_subtype:
        example: metamask
        type: string
        enum: [metamask, parity, unknown]
```

PR #1147, opened 2018-08-01
Adds WalletType to User

Moejoe90:
related https://github.com/Neufund/platform-frontend/issues/1049


app/lib/api/users/UsersApi.ts
```diff
@@ -27,17 +33,25 @@ export class UsersApi {
 
   public async createAccount(newUser?: IUserInput): Promise<IUser> {
     this.logger.info("Creating account for email: ", newUser && newUser.newEmail);
+    // Backend expects walletType and walletSubType values as lower case
+    const modifiedNewUser = newUser
```
krzkaczor: Have you tried asking backend team to normalize this? Maybe they dont' care too much and then we dont need to (de)serialization steps

Moejoe90: We discussed this yesterday Backend strongly pushed for this solution 
From my opinion I think it looks really ugly.

huesforalice: Backend enums are all lower type. I want to stay consistent there.

PR #1172, opened 2018-08-06
Kk/withdraw gas price

krzkaczor:


app/components/modals/txSender/shared/TxPending.tsx
```diff
@@ -0,0 +1,32 @@
...
+  txHash: string;
+}
+
+export const TxPending: React.SFC<ITxPendingProps> = ({ blockId, txHash }) => (
+  <div className="text-center">
+    <SpinningEthereum className="mb-3" />
+
+    <h3 className={styles.title}>Pending transaction</h3>
+
+    <p>
```
Moejoe90: Can we add this to translations?

webpack/proxy-urls.js
```diff
@@ -12,7 +12,7 @@ function generateProxyConfig(backendUrl) {
   const base = {
     "/node": {
       targetLocal: "http://localhost:8545",
-      targetRemote: "http://localhost:8545",
+      targetRemote: "https://platform.neufund.io/nodes/private",
```
Moejoe90: Does this mean there is no need for ganache with this change?

app/components/modals/txSender/withdraw-flow/Withdraw.tsx
```diff
@@ -57,3 +78,21 @@ export const Withdraw: React.SFC<IWithdrawProps> = ({ onAccept }) => (
     </Formik>
   </div>
 );
+
+export const GasComponent: React.SFC<IGasState> = ({ gasPrice, error }) => {
+  if (error) {
+    return <WarningAlert>Error occured while getting optimal GAS price.</WarningAlert>;
```
Moejoe90: Also this can you add it to translatiosn

webpack/proxy-urls.js
krzkaczor: ah fuck maybe i should not commit this. Let's talk about it during daily

Issue #1173, opened 2018-08-06
Change slider values| TERM Sheet


SergejSoft:
![Screen Shot 2018-08-06 at 14.59.35.png](https://images.zenhubusercontent.com/57dd639be40e5714b16f0753/c46e0dcf-a867-4218-b58e-5346b4329150) 
Public Offer duration 14 days min max 60 days.
Signing Duration min 14 days ma 30 days.

Issue #1170, opened 2018-08-06
Format the box with 4 digits numbe with EURO sign


SergejSoft:
<img width="699" alt="screen shot 2018-08-06 at 14 30 54" src="https://user-images.githubusercontent.com/2290474/43716703-64c9f67c-9985-11e8-9d48-b0d80d8bb4e4.png">

Issue #1169, opened 2018-08-06
Min new shares to issue. 


SergejSoft:
We need a positive number validator.

0 possible and all positive number. Make negative number not possible ot enter. 

![Screen Shot 2018-08-06 at 14.28.04.png](https://images.zenhubusercontent.com/57dd639be40e5714b16f0753/87bf9b9b-28c1-48c6-b46a-7650428a16c0)

Issue #1107, opened 2018-07-26
[settings] all users | incorrect box title 


glupiawymowka:
there is box "1. " and "2" 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/263cc9ff-5804-4871-98f5-b8c95f5a082b)

according to designs, both versions are wrong and the numbers should be gone 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/615a02aa-9f52-4478-b0db-a92c17b5b681)

PR #1180, opened 2018-08-07
UI fixes

lizonp:
Related: #1170 #1107 #1173 #1169 

Please note, that after changing minimum value of `minimumNewSharesToIssue` field to 0 as a response we have error.

```
{
  "detail": "0 is less than the minimum of 1",
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank"
}
```


SergejSoft:
No, we can't get 400 on "0". Please make it possible to have 0 and above. So all positive numbers.

lizonp:
@SergejSoft This has to be changed on the backend side (@rudolfix )

PR #1193, opened 2018-08-08
fixes backend pinning on CI

rudolfix:


app/lib/api/client/JsonHttpClient.ts
```diff
@@ -65,9 +65,11 @@ export class NetworkingError extends HttpClientError {
 //supports only JSON apis
 @injectable()
 export class JsonHttpClient implements IHttpClient {
+  // todo: JWT should be injected there if is is available in ObjectStorage, now just clear the header
```
Moejoe90: Have you checked AuthorizedJsonHttpClient?

.gitignore
```diff
@@ -14,3 +14,4 @@ __screenshots__
 e2e/platform-backend
 .reg
 .vscode
+.idea
```
Moejoe90: Is there a reason for adding this?

krzkaczor: This should go to your global gitingore not project specific.

Issue #1220, opened 2018-08-15
Public Page fixes and redesign


SergejSoft:
Only valid after design is provided by Design Department.

Master sheet: https://docs.google.com/presentation/d/1-BReyNlQlIGC5kH3DCvzt0XxcKf2goKse7lzYn4GhlQ/edit?usp=sharing

New elements:

- [ ] count down element https://zpl.io/V0yvwnK

- [ ] Timeline - move Today to Campaigning. 

https://zpl.io/V0yvwnK

- [ ] Youtube shall work as a popup and take the middle part of the screen. 
<iframe src=“your_page_url”
       allowfullscreen=“allowfullscreen”
       mozallowfullscreen=“mozallowfullscreen”
       msallowfullscreen=“msallowfullscreen”
       oallowfullscreen=“oallowfullscreen”
       webkitallowfullscreen=“webkitallowfullscreen”> </iframe>

- [ ] Pitch deck (as ifame) component next to youtube. 

- [ ]  Legal info with cap table need fix. See the form.

- [ ] ETO Investment terms. See the master link. page 14.

- [ ] Public View: Please look at the page 15 in the master link.
Key People section need to have 300 words.

PR #1242, opened 2018-08-20
#1220 public page about section

lizonp:
related: #1220

- [x] Change about section layout
- [x] Add video modal
- [x] Change social icons layout

screen shot 2018-08-20 at 15 35 18


app/components/shared/PeopleSwiperWidget.tsx
```diff
@@ -16,7 +15,7 @@ export interface IPerson {
   description: string;
   layout?: TSlidePersonLayout;
   role?: string;
-  socialChannels: IEtoSocialProfile[];
+  socialChannels: any;
```
trival: maybe giving up typing is not necessary?

app/components/shared/SlidePerson.tsx
```diff
@@ -3,15 +3,15 @@ import * as React from "react";
...
+import { SocialProfilesList } from "./SocialProfilesList";
 
 import * as styles from "./SlidePerson.module.scss";
 
 export type TSlidePersonLayout = "horizontal" | "vertical";
 interface IProps {
   srcSet: ISrcSet;
   name: string;
-  socialChannels: IEtoSocialProfile[];
+  socialChannels: any;
```
trival: : (

app/components/shared/SocialProfilesList.tsx
```diff
@@ -56,8 +75,9 @@ export const SocialProfilesList: React.SFC<IProps> = ({
           url && (
             <div className={styles.profile} key={type}>
               <a href={url} target="_blank" title={url}>
-                <InlineIcon svgIcon={SOCIAL_PROFILES[type as string]} />
+                <InlineIcon svgIcon={SOCIAL_PROFILES[type]} />
```
trival: I personaly prefer the old cast here, than the "any" typed properties in other components. To be even shorter you could write SOCIAL_PROFILES[type!] with the exclamation mark after type (to say typescript this is not undefined at this point)

trival: But the actual problem is, that the original type of "type" in socialChannels is optional. This is only to make the progress calculator not show 50% from the start (I did that back then...). But I could improve the progress calculator, so that we have propper and true typings everywhere. That would be the cleanest, but also a little overhead...

app/components/eto/registration/pages/EtoMedia.tsx
```diff
@@ -44,7 +44,16 @@ const EtoForm = (props: FormikProps<TPartialCompanyEtoData> & IProps) => (
       </p>
       <Row>
         <Col className="offset-1" xs={10}>
-          <FormField name="companyVideo.url" placeholder="url" />
+          <FormField name="companyPitchDeck.url" placeholder="url" />
```
trival: Is companyPitchDeck right! When I try to save and test a video, i get an error from the api. Also, the property is not used in EtoPublicComponent. There you still use companyVideo.

Issue #1238, opened 2018-08-20
Pls. fix. Public View


SergejSoft:
<img width="1403" alt="screen shot 2018-08-20 at 11 01 09" src="https://user-images.githubusercontent.com/2290474/44330890-6d2cb280-a468-11e8-8ec0-d3e84e8944ef.png">

PR #1247, opened 2018-08-21
Fix ETO Investment Terms widget layout

lizonp:


lizonp:
realted: #1238 

screen shot 2018-08-21 at 14 30 33
screen shot 2018-08-21 at 14 30 21

PR #1256, opened 2018-08-25
Form fields changes

lizonp:


app/components/shared/MediaLinksEditor.tsx
```diff
@@ -19,21 +19,44 @@ interface ISingleMediaLinkFieldInternalProps {
   name: string;
   url?: string;
   placeholder: string;
+  blankField: any;
```
trival: Somehow I would prefer that not to be "any". As far as i see, the requirement is an extra "publication" field, which i havn't found in the swagger schema yet, nor in the Yup interface definition. I would prefer to make such changes explicit, define them in the API interface, then use these types in the components as well...

Issue #1272, opened 2018-09-03
Public View Page to be seen for not signed users


SergejSoft:
Enable Public View for not signed users. 
To make it work as langing page.

PR #1283, opened 2018-09-05
#1272 Public View Page to be seen for not signed users

desfero:


app/components/App.tsx
```diff
@@ -37,13 +35,10 @@ class AppComponent extends React.Component<IStateProps> {
...
     return (
       <>
-        <Header />
-        <div className="wrapper">
-          <ScrollToTop>
-            <AppRouter />
-          </ScrollToTop>
-        </div>
-        <Footer />
+        <ScrollToTop>
```
desfero: @Moejoe90 @lizonp there are some changes to the layout to get rid of Header and Footer from ETO preview. Therefore if you can please review them as those are global app changes and maybe I missed something as it's my second day in the project. Thanks

app/components/eto/registration/pages/EtoTerms.tsx
```diff
@@ -33,7 +32,16 @@ interface IDispatchProps {
 
 type IProps = IStateProps & IDispatchProps;
 
-const currencies = ["eur_t", "eth"];
+interface ICurrencies {
+  [key: string]: string;
+}
+
+const CURRENCIES: ICurrencies = {
```
desfero: @trival @Moejoe90 I think it's worth to move 'CURRENCY' to some shared places as it's not specific to that component. Do you know some place where we can move that?

app/components/eto/EtoPreview.tsx
```diff
@@ -1,35 +1,35 @@
...
+}
+
 interface IDispatchProps {
   loadEtoPreview: (previewCode: string) => void;
 }
 
-type IProps = IStateProps & IDispatchProps;
+type IProps = IStateProps & IDispatchProps & IRouterParams;
 
 class EtoPreviewComponent extends React.Component<IProps> {
```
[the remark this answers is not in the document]
Moejoe90: From my side I like Recompose I see that it takes logic out of the component and makes code cleaner. 

What I would suggest is we keep this as is now to not break consistency and introduce this change later on with bigger refactoring for other components. In addition to adding it to our platform development guidelines. 

Or if you think this wouldn't take much time then we can introduce it now with this component as an example on how to use it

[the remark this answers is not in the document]
desfero: @Moejoe90 great, sound like a plan, thanks

package.json
```diff
@@ -19,7 +19,6 @@
     "@cypress/webpack-preprocessor": "2.0.1",
     "@storybook/addon-actions": "^3.4.10",
     "@storybook/react": "^4.0.0-alpha.16",
-    "@types/bignumber.js": "^5.0.0",
```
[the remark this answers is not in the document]
desfero: yes, types for bignumber are provided by the package https://github.com/MikeMcl/bignumber.js/blob/master/package.json#L26

desfero:
@SergejSoft noooo, again conflict :D

app/utils/withContainer.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from "react";
+
+export const withContainer: (
```
Moejoe90: :+1:

app/lib/api/eto/EtoApi.ts
```diff
@@ -15,32 +15,35 @@ const ETO_PREVIEW_PATH = "/eto-previews/";
 
 @injectable()
 export class EtoApi {
-  constructor(@inject(symbols.authorizedJsonHttpClient) private httpClient: IHttpClient) {}
+  constructor(
+    @inject(symbols.authorizedJsonHttpClient) private authorizedHttpClient: IHttpClient,
+    @inject(symbols.jsonHttpClient) private httpClient: IHttpClient,
```
Moejoe90: :+1:

app/components/eto/registration/pages/EtoTerms.tsx
Moejoe90: Let's discuss this tomorrow

PR #1297, opened 2018-09-05
Working withdraw 

Moejoe90:
Done 
- Sending Transaction With Light Wallet
- Sending Transactions with Ledger Nano
- More Robust Temporary Transactions Mechanism 
- Simple Withdraw E2E test

Missing 
- Transaction check before sending transaction 
- Additional E2E tests

TODO: 
- Abstract transaction checking for both Withdraw and Investment Flow check what can be reused
- Add field validation on transaction level this is solved when investment flow methods are abstracted or 
  used


app/components/wallet/wallet-balance/__test__/WalletWithdraw.spec.e2e.ts
```diff
@@ -0,0 +1,112 @@
...
+    cy.visit("/recover/seed");
+
+    typeLightwalletRecoveryPhrase(words);
+
+    cy.request({ url: mockApiUrl + "sendgrid/session/mails", method: "DELETE" });
+
+    cy.get(tid("wallet-selector-register-email")).type(email);
+    cy.get(tid("wallet-selector-register-password")).type("strongpassword");
+    cy.get(tid("wallet-selector-register-confirm-password")).type("strongpassword{enter}");
+    cy.wait(2000);
```
desfero: @Moejoe90 it's possible to get rid of `wait` and simply use some kind of pooling whether the action is completed? Waiting for me is an anti-pattern that really can break tests and force them to wait more than needed (https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting)

app/components/wallet/wallet-balance/__test__/WalletWithdraw.spec.e2e.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    cy.get(tid("recovery-success-btn-go-dashboard")).click();
+
+    cy.wait(5000);
+    assertLatestEmailSentWithSalt(email);
+
+    assertUserInDashboard();
+    cy.get(tid("authorized-layout-wallet-button")).click();
+    cy.get(tid("shared.wallet.wallet-balance.unlock-wallet.account-address")).then(
```
desfero: Does cypress support async/await or still it's a problem there?

app/components/wallet/wallet-balance/__test__/WalletWithdraw.spec.e2e.ts
```diff
@@ -0,0 +1,112 @@
...
+            cy.get(tid("modals.tx-sender.withdraw-flow.summery.withdrawSummery.accept")).click();
+            cy.get(tid("access-light-wallet-prompt-accept-button")).click();
+            cy.get(tid("modals.shared.signing-message.modal"));
+            cy.get(tid("modals.tx-sender.withdraw-flow.success"));
+
+            cy.get(tid("modals.tx-sender.withdraw-flow.tx-hash")).then(txHashObject => {
+              getTransactionReceiptRpc(NODE_ADDRESS, txHashObject.text()).then(data => {
+                const { from, gas, gasPrice, input, hash, to, value } = data.body.result;
+
+                const ethValue = new BigNumber(value).toString();
```
desfero: maybe the 'value' is already a string?

app/modules/tx/monitor/sagas.ts
```diff
@@ -8,17 +8,44 @@ import { neuTakeUntil } from "../../sagasUtils";
...
+function* txMonitor({ logger, notificationCenter }: TGlobalDependencies): any {
   while (true) {
     logger.info("Querying for pending txs...");
-
-    yield neuCall(updateTxs);
+    try {
+      yield neuCall(updateTxs);
+    } catch (e) {
+      notificationCenter.error("Error while trying to get pending transactions");
+      logger.error(e);
```
desfero: Maybe it's worth to add some short message to the error like `logger.error('Error getting pending txs', e)`. Then it's easier to find where the message comes from in tools like sentry as you can search where 'Error gettings pending txs' is in the code.

app/components/modals/txSender/shared/SigningMessage.tsx
```diff
@@ -3,9 +3,9 @@ import * as React from "react";
 import { LoadingIndicator } from "../../../shared/LoadingIndicator";
 
 export const SigningMessage = () => (
-  <>
-    <h3>Singing message</h3>
+  <div data-test-id="modals.shared.signing-message.modal">
+    <h3>Signing Transaction</h3>
     <p>Accept message on your signer</p>
```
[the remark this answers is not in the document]
Moejoe90: You are absolutely correct however there is a ticket for redesigning this 
https://github.com/Neufund/platform-frontend/issues/1299
so it will be time wasting to put them in translations now

app/components/modals/txSender/withdraw-flow/Success.tsx
```diff
@@ -3,9 +3,14 @@ import { ConfettiEthereum } from "../../../landing/parts/ConfettiEthereum";
...
-export const WithdrawSuccess = () => (
-  <div className="text-center">
+interface IProps {
+  txHash?: string;
+}
+
+export const WithdrawSuccess: React.SFC<IProps> = ({ txHash }) => (
+  <div className="text-center" data-test-id="modals.tx-sender.withdraw-flow.success">
     <ConfettiEthereum className="mb-3" />
     <h3 className={styles.title}>Transaction confirmed</h3>
```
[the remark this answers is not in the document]
Moejoe90: same for comment above

app/components/wallet/wallet-balance/__test__/WalletWithdraw.spec.e2e.ts
```diff
@@ -0,0 +1,112 @@
+import BigNumber from "../../../../../node_modules/bignumber.js/bignumber.js";
```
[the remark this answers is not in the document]
Moejoe90: Sadly this is related with how `cypress` deals with packages it doesn't see `node_modules` so I'm doing imports based on relative paths. this introduces another problem btw you will see in the deposit e2e test.
if you have ideas on how to solve this problem then please give me a hit

app/components/wallet/wallet-balance/__test__/WalletWithdraw.spec.e2e.ts
```diff
@@ -0,0 +1,112 @@
...
+      "open",
+      "leg",
+      "elegant",
+      "bracket",
+      "lamp",
+      "day",
+    ];
+
+    const email = "john-smith@example.com";
+    const testValue = (1).toString();
```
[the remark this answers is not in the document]
Moejoe90: I want to show that the value we are initially asserting is 1 as a number if you think that wasn't clear then it should probably be changed

app/e2e-test-utils/ethRpcUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+export const getTransactionReceiptRpc = (nodeAddress: string, txHash: string) =>
+  requestFromWeb3Node(nodeAddress, "eth_getTransactionByHash", [txHash]);
+
+export const getBalanceRpc = (nodeAddress: string, address: string) =>
+  requestFromWeb3Node(nodeAddress, "eth_getBalance", [address, "latest"]);
+
+export const sendTransactionRpc = (
+  nodeAddress: string,
```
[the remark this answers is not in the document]
trival: But, probably not, if this is cypress code

app/lib/web3/LightWallet.ts
```diff
@@ -32,6 +33,16 @@ export interface IVault {
   salt: string;
 }
 
+interface ITransaction {
```
trival: don't we have that already, at least in tx sender reducer

app/lib/web3/LightWallet.ts
```diff
@@ -212,7 +227,27 @@ export class LightWallet implements IPersonalWallet {
   }
 
   public async sendTransaction(data: Web3.TxData): Promise<string> {
-    return this.web3Adapter.sendTransaction(data);
+    if (!this.password) {
+      throw new LightWalletMissingPassword();
```
trival: could we add a ...Error suffix to the name? I feel that everything that you can throw should have Error in the name, otherwise it is not so clear that this is not actually some custom class instance

app/lib/web3/Web3Adapter.ts
```diff
@@ -67,6 +67,28 @@ export class Web3Adapter {
     return resultData.result;
   }
 
+  public async getTransactionReceipt(txHash: string): Promise<Web3.TxData> {
+    const getTransactionReceipt = promisify<Web3.TxData, string>(
+      this.web3.eth.getTransactionReceipt.bind(this.web3.eth),
+    );
+    return await getTransactionReceipt(txHash);
```
trival: the async function returns a promise anyway, so no need to await here.

I wonder if we couldn't write something like: 
```
this.getTransactionReceipt = promisify<Web3.TxData, string>(
      this.web3.eth.getTransactionReceipt.bind(this.web3.eth),
    );
```
directly in the constructor, it would do some nice caching... Even though our react style guide does not recommend it, this code here is also not the most beautiful as well

app/modules/tx/monitor/sagas.ts
```diff
@@ -8,17 +8,44 @@ import { neuTakeUntil } from "../../sagasUtils";
...
+): Promise<any> {
+  const nodePendingTx = await Promise.all(
+    apiPendingTx.map(async tx => {
+      const transactionReceipt = await web3Manager.internalWeb3Adapter.getTransactionReceipt(
+        tx.transaction.hash,
+      );
+      // transactionReceipt should be null if transaction is not mined
+      if (transactionReceipt) return tx;
+      else await apiUserService.deletePendingTx(tx.transaction.hash);
+    }),
```
trival: does that logic mean: if transaction is not mined, then delete it? Is that correct?

app/modules/tx/monitor/sagas.ts
```diff
@@ -8,17 +8,44 @@ import { neuTakeUntil } from "../../sagasUtils";
...
+  );
+  return nodePendingTx.filter(tx => (tx ? tx : undefined));
```
trival: this could be rewritten as .filter(tx => tx) which removes all falsy values...

app/modules/tx/monitor/sagas.ts
Moejoe90: Maybe i should change the comment 
If `transactionReceipt` is null then it means that there is still no receipt and the transaction is pending

app/modules/tx/monitor/sagas.ts
Moejoe90: I saw your trick earlier :D I'll start using it

app/components/wallet/wallet-balance/__test__/WalletWithdraw.spec.e2e.ts
Moejoe90: @desfero 
to clear your confusion :)
https://docs.cypress.io/api/commands/then.html

app/components/wallet/wallet-balance/__test__/WalletWithdraw.spec.e2e.ts
Moejoe90: Actually sorry my bad in this case it's completely correct to remove `wait` it was used before when dealing with backend API calls from tests. 
I'll remove this

app/e2e-test-utils/ethRpcUtils.ts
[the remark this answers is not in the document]
Moejoe90: You are totally right. I found a way to import web3 so i'm actually planning on changing this and using web3 completely so no worries about this for now

app/lib/web3/LightWallet.ts
Moejoe90: you're right i will use that. 
I also moved `ITxData` from the `TxSender` reducer  to `Web3Adapter` i think its a better location there

app/lib/web3/Web3Adapter.ts
Moejoe90: I was thinking of the same thing like you. I don't agree as well with the way it is now. i was planning to do full refactoring and putting all the `promisify` in the constructor

app/modules/tx/monitor/sagas.ts
desfero: @trival @Moejoe90 there is an even more concise way to write it `.filter(Boolean)` :D (https://twitter.com/kentcdodds/status/1009918457394225152)

app/components/shared/AccountAddress.tsx
```diff
@@ -27,21 +27,21 @@ export class AccountAddress extends React.Component<IAccountAddressProps & Commo
...
 
     return (
       <div className={cn(styles.accountAddress, className)}>
         <Avatar seed={address} />
 
         <div className={styles.addressWrapper}>
-          <div className={styles.address}>
+          <div className={styles.address} data-test-id="account-address.your.ether-address.div">
             {address}
             <input
```
Moejoe90: Is there a reason why `{address}` is a div and and `input` at the same time?

lizonp: Yes, it's super simple to copy sometghing from input instead of selecting correct textNode

app/modules/tx/monitor/sagas.ts
trival: @desfero oh i like that : )

app/components/shared/AccountAddress.tsx
Moejoe90: :+1:

Issue #1305, opened 2018-09-06
[Issuer] Disable form inputs when ETO state is not 'preview'


desfero:
1.
Given you are logged as issuer
When ETO status is not 'preview'
Then disable inputs in Legal info, Investment, Terms Eto Terms Equity Token Info, Token Holder Voting Rights forms.

2.
Given you are logged as issuer and you are on the Dashboard
When ETO status is not 'preview'
Then change CTA text from 'COMPLETE or 'Edit' to 'Show' for Legal info, Investment, Terms Eto Terms Equity Token Info, Token Holder Voting Rights forms.

PR #1311, opened 2018-09-10
#1305 Disable form inputs when ETO state is not 'preview'

desfero:
#1305


app/components/eto/dashboard/ETOFormsProgressSection.tsx
```diff
@@ -50,104 +55,90 @@ const ETOFormsProgressSectionComponent: React.SFC<IStateProps> = ({
...
+          <EtoFormProgressWidget
+            isLoading={loadingData}
+            to={section.redirectTo}
+            progress={shouldEtoDataLoad ? section.progress : 0}
+            disabled={!shouldEtoDataLoad}
+            readonly={section.readonly}
+            name={section.name}
+          />
+        </Col>
+      ))}
```
trival: nice

app/components/eto/dashboard/ETOFormsProgressSection.tsx
```diff
@@ -22,6 +25,7 @@ import { EtoFormProgressWidget } from "../../shared/EtoFormProgressWidget";
 import { etoRegisterRoutes } from "../registration/routes";
 
 export interface IStateProps {
+  etoStatus?: EtoState;
```
desfero: @Moejoe90 @trival there is a difference in namings of ETO status/state that we need to align and follow as there are `etoStatus` and `etoState` naming collisions as both of them represent basicaly the same value.

Moejoe90: :+1:

Issue #1200, opened 2018-08-10
Set `SUBMIT_PROPOSAL_THRESHOLD` to 1.0


Moejoe90:
`SUBMIT_PROPOSAL_THRESHOLD` is responsible for how many required fields must be filled before `Submit ETO` show's. 

Currently `SUBMIT_PROPOSAL_THRESHOLD` is set to 0.5 this means that only half of the required fields should be filled before `Submit ETO shows` 

This was done due to some anomalies in ETO forms once everything is set this needs to be set back to 1

PR #1322, opened 2018-09-10
Changes ETO form threshold

Moejoe90:
This changes the Threshold of the form for testing purposes now it will need 100% of the required fields 
https://github.com/Neufund/platform-frontend/issues/1200

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,9 +86,9 @@ jobs:
           - v11-deps-
       - run: yarn prepare
 
-      - run: yarn storybook:screenshots
+      # - run: yarn storybook:screenshots
       # Re-enable storybook:screenshots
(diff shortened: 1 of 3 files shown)


desfero:
@Moejoe90 I'm wondering if we increase the threshold probably it's also worth to remove that dirty hacks from forms (https://github.com/Neufund/platform-frontend/blob/master/app/components/eto/registration/pages/EtoMedia.tsx#L111) that allows sending invalid form data?

Moejoe90:
@desfero this was before a business request at least as I remember I would prefer to have validation on form level indeed and use `onSubmit` instead of this hack. 
The question is more to @SergejSoft is it fine now from a product point to use validation on field before save or not. 
I will merge this PR anyway if we agree to change to `onSubmit` then let's do it as another ticket

PR #1341, opened 2018-09-12
Improve slideshare UI and UX

lizonp:


app/components/shared/Slides.tsx
```diff
@@ -10,7 +10,15 @@ interface IProps {
 export const Slides: React.SFC<IProps> = ({ slideShareUrl, className }) => {
   return slideShareUrl ? (
     <Proportion className={className}>
-      <iframe src={slideShareUrl} allowFullScreen scrolling="no" />
+      <iframe
+        src={
+          slideShareUrl.includes("https://")
```
trival: Write a TODO to extract that into an own utils library (url.utils.ts maybe) with its own unit test. Then we can merge

app/components/eto/shared/EtoPublicComponent.tsx
```diff
@@ -184,7 +184,7 @@ export const EtoPublicComponent: React.SFC<IProps> = ({ companyData, etoData })
                 <div>{companyData.brandName}</div>
                 {companyData.companyWebsite && (
                   <a href={normalizedUrl(companyData.companyWebsite)} target="_blank">
-                    {companyData.companyWebsite || DEFAULT_PLACEHOLDER}
+                    {companyData.companyWebsite.split("//")[1] || DEFAULT_PLACEHOLDER}
```
trival: This can go to new url.utils.ts as well

PR #1350, opened 2018-09-15
enables parity signup and transaction

rudolfix:
how to test
1. install extension https://chrome.google.com/webstore/detail/parity-ethereum-integrati/himekenlppkgeaoeddcliojfddemadig?hl=en
2. run backend locally or if you test just sign-up run instant seal parity container
3. select wallet in parity (selecting from extension popup fails in my browser)
4. disable metamask
5. parity signer will appear when sign message will be sent to the signer

ad 2. extension works with just local parity node - maybe a proxy can be set up to work agains io node which will be best way to do it. we can try to attach light client to stage


Moejoe90:
https://github.com/paritytech/parity-extension/issues/123
The extension is deprecated maybe its better to not support it

Issue #1342, opened 2018-09-13
Do not wrap router link with additional button element


desfero:
**Background:**
There are a couple of places where we wrap the router link with additional button tag that makes keyboard navigation difficult. (for e.g. on the issuer dashboard each section link (Edit/Complete/Show) is wrapped in additional button tag). Doing it in that way is not semantically correct and should be always avoided.
<img width="267" alt="screen shot 2018-09-13 at 12 25 46" src="https://user-images.githubusercontent.com/7031578/45483114-40f70f80-b750-11e8-8b24-6a88f9058a95.png">

**Technical notes:**
There is a [good implementation](https://stackoverflow.com/a/49439893) that we can adjust for our needs to fix an issue.

**Testing notes::**
Following buttons should be checked:
Header -> Login and Register
Investor dashboard -> Manage Wallet 
KYC -> Go Back 
Landing -> Register NOW
Seed backup -> View Again and Backup phrase
Issuer ETO dashboard -> Edit/Complete/Show buttons on each section
Investor investment preview -> Details

PR #1374, opened 2018-09-18
#1342 Do not wrap router links in 'a' tag

desfero:
#1342 

@Moejoe90 @lizonp there are tons of import changes.

The main changes are:
- New `ButtonLink` component that replaces all `<Link><Button/></Link>`.
- `Buttons.tsx` are moved to the `buttons` folder with `ButtonLink` to have consistency with `forms` folder.


app/types.d.ts
```diff
@@ -77,3 +77,5 @@ export type TAcceptedFileType =
   | "image/svg+xml"
   | "image/*";
 // TODO: Correct TAcceptedFileType types it can contain more than one type
+
+export type Omit<T extends K, K> = Pick<T, Exclude<keyof T, keyof K>>;
```
Moejoe90: @desfero what does this do?

desfero: It's a helper type that typescript didn't provide. 
```
interface Base {
  onClick: () => {}
}

interface Derived extends Base {
  isLoading: boolean;
}

type WithoutOnClick = Omit<Derived, Base>;
// WithoutOnClick === { isLoading: boolean }

```
Used to get rid of `onClick` inside ButtonLink as `OnClick` there is provided by component

PR #1377, opened 2018-09-19
Add support for investing whole wallet balance

desfero:


app/components/modals/txSender/investment-flow/Investment.tsx
```diff
@@ -69,295 +67,243 @@ interface IStateProps {
...
+  withHandlers<IStateProps & IDispatchProps & IWithProps, IHandlersProps>({
+    investEntireBalance: ({ investmentType, wallets, changeEuroValue, gasCostEuro }) => () => {
+      const wallet = wallets.find(w => w.type === investmentType);
+
+      if (wallet === undefined || wallet.type === EInvestmentType.BankTransfer) {
+        throw new Error("Can't invest wallet entire balance. Wallet not found.");
+      }
+
+      const availableEurBalance = subtractBigNumbers([wallet.balanceEur, gasCostEuro]);
+      const balanceEurFormatted = formatMoney(availableEurBalance, MONEY_DECIMALS);
```
desfero: @Moejoe90 I'm not quite happy about passing UI formatted price as an action payload. 
I think we need to refactor it in future to separate ui formatting from business logic. 
As from state we receive wallet balance in a different format than we send back to as an action. 
That should be the same format as with the current approach we can lose some decimal precision due to rounding on ui side.
What do you think?

app/components/modals/txSender/investment-flow/InvestmentTypeSelector.tsx
```diff
@@ -8,35 +7,81 @@ import { Money } from "../../../shared/Money";
...
+  balanceEur: string;
   name: string;
   icon: string;
-  balanceEth?: string;
-  balanceNEuro?: string;
-  balanceEur?: string;
 }
 
+interface InEuroWallet {
+  type: EInvestmentType.ICBMnEuro;
```
desfero: @Moejoe90 I refactored it to Discriminated Union Types as it's a perfect use case for that Typescript feature.

Moejoe90: I see what you did there, 
Maybe it's better to go with Typescripts convention of using kind?
We can do that later as well as part of refactoring what do you think

app/components/modals/txSender/investment-flow/Investment.tsx
```diff
@@ -69,295 +67,243 @@ interface IStateProps {
...
 );
 
 export const InvestmentSelection: React.SFC = compose<any>(
   injectIntlHelpers,
   appConnect<IStateProps, IDispatchProps>({
     stateToProps: state => {
-      const i = state.investmentFlow;
-      const eto = selectEtoById(state.publicEtos, i.etoId)!;
-      const eur = selectEurValueUlps(i);
+      const investmentFlow = state.investmentFlow;
```
Moejoe90: I think its always way simpler to put everything in the return instead of declaring constants 
In the case of `eto` and `eur` they can be put directly in the return 
But its not a blocking request

app/components/modals/txSender/investment-flow/Investment.tsx
Moejoe90: I didn't fully understand can you elaborate more?

Moejoe90: Can you elaborate what
'As from state we receive wallet balance in a different format than we send back to as an action.'
Means

PR #1379, opened 2018-09-19
Full Withdraw 

Moejoe90:
This adds 

- Correct reverted transaction handeling
- Out of gas case when a transaction was accepted by the node but ran out of gas during transaction "Wrong gas estimation"
- Fixes Some issues related with even emitting
- Adds full token withdraw

TODO: 
- [ ] Take out Withdraw Transaction Generation from sender saga
- [ ] Add-on field validation for withdraw saga
- [ ] Add correct summary values during withdraw


app/components/eto/publicView/LegalInformationWidget.tsx
```diff
@@ -128,7 +128,6 @@ export const LegalInformationWidget: React.SFC<IProps> = ({ companyData, etoData
                 datasets: [
                   {
                     data: companyData.shareholders.map(d => d && d.shares),
-                    /* tslint:disable:no-unused-variable */
```
Moejoe90: I removed them as they seem not needed and anti pattern

app/modules/tx/sender/sagas.ts
```diff
@@ -90,26 +121,61 @@ function* ensureNoPendingTx(
   yield put(actions.txSender.txSenderWatchPendingTxsDone(type));
 }
 
+function* generateEthWithrdawTransaction(
+  { contractsService }: TGlobalDependencies,
+  action: TAction,
+): any {
+  if (action.type !== "TX_SENDER_GENERATE_TX") return;
```
desfero: @Moejoe90 @trival we need to think how to get rid of those if's inside each saga. As they make catching errors harder (consider that you change `yield fork(neuTakeEvery, "TX_SENDER_GENERATE", generateEthWithrdawTransaction);` than `getenrateEthWithdrawTransaction` will not be correctly processed due to `if (action.type !== "TX_SENDER_GENERATE_TX")`)

app/modules/tx/sender/sagas.ts
```diff
@@ -90,26 +121,61 @@ function* ensureNoPendingTx(
   yield put(actions.txSender.txSenderWatchPendingTxsDone(type));
 }
 
+function* generateEthWithrdawTransaction(
```
desfero: typo: With**dra**w

app/modules/tx/sender/sagas.ts
```diff
@@ -90,26 +121,61 @@ function* ensureNoPendingTx(
...
+  { contractsService }: TGlobalDependencies,
+  action: TAction,
+): any {
+  if (action.type !== "TX_SENDER_GENERATE_TX") return;
+
+  const txStateDetails = action.payload;
+
+  let txDetails: ITxData | undefined;
+  if (!txStateDetails) return;
```
desfero: Is it possible for `txStateDetails` to be empty?

app/modules/tx/sender/sagas.ts
```diff
@@ -90,26 +121,61 @@ function* ensureNoPendingTx(
...
+
+  // transaction can be fully covered by etherTokens
+
+  const txInput = contractsService.etherToken
+    .withdrawAndSendTx(txStateDetails.to, new BigNumber(txStateDetails.value))
+    .getData();
+  const ethVal = new BigNumber(txStateDetails.value);
+  const difference = ethVal.sub(etherTokenBalance);
+
+  txDetails = {
```
desfero: I think we can remove `let txDetails: ...` from line `132` and put `const txDetails = {`

app/modules/tx/sender/sagas.ts
```diff
@@ -90,26 +121,61 @@ function* ensureNoPendingTx(
...
+    to: contractsService.etherToken.address,
+    from,
+    data: txInput,
+    value: difference.comparedTo(0) > 0 ? difference.toString() : "0",
+    gas: txStateDetails.gas,
+    gasPrice: txStateDetails.gasPrice,
+  };
+  if (txDetails) {
```
desfero: `txDetails` is always set, so putting it inside `if` is redundant.

app/modules/tx/sender/sagas.ts
```diff
@@ -90,26 +121,61 @@ function* ensureNoPendingTx(
...
         gasPrice: "0x" + Number(txData.gasPrice!).toString(16),
         hash: txHash,
         input: txData.data || "0x0",
-        nonce: "0x" + Number(txData.nonce!).toString(16),
-        to: txData.to!,
-        value: "0x" + Number(txData.value!).toString(16),
+        // TODO:Connect nonce
+        nonce: "0x0",
+        to: txData.to,
+        value: "0x" + Number(txData.value).toString(16),
```
desfero: Can we reuse `addHexPrefix` for consistency?

app/modules/tx/sender/sagas.ts
```diff
@@ -122,42 +188,95 @@ function* sendTxSubSaga({ web3Manager, apiUserService }: TGlobalDependencies): a
...
 
     return txHash;
-  } catch (e) {
-    yield put(actions.txSender.txSenderError("Tx was rejected"));
-    throw e;
+  } catch (error) {
+    // @see https://github.com/paritytech/parity-ethereum/blob/61f4534e2a5f9c947661af0eaf4af5b76456efe0/rpc/src/v1/helpers/errors.rs#L304
+    if (
+      error.code === -32010 &&
+      error.message.startsWith("Insufficient funds. The account you tried to send transaction")
```
desfero: Ouuu, that for sure will broke our app in future. As error messages can be often changed :(

app/modules/tx/sender/sagas.ts
```diff
@@ -122,42 +188,95 @@ function* sendTxSubSaga({ web3Manager, apiUserService }: TGlobalDependencies): a
...
+    ) {
+      return yield put(actions.txSender.txSenderError("Not enough funds"));
+    }
+    if (
+      (error.code === -32000 && error.message === "intrinsic gas too low") ||
```
desfero: @Moejoe90 where we can find that message on `parity-ethereum` repo?

package.json
```diff
@@ -123,6 +123,7 @@
     "typestrict": "^0.0.9",
     "uglifyjs-webpack-plugin": "^1.3.0",
     "url-loader": "^1.1.0",
+    "web3-eth-accounts": "^1.0.0-beta.36",
```
desfero: beta.36 xD

app/modules/tx/sender/sagas.ts
Moejoe90: Yeah but these are errors coming from the ethereum node there is no way out of it

app/modules/tx/sender/sagas.ts
Moejoe90: Yeah took them from there check the link i pasted as a comment
:)

app/components/modals/txSender/TxSender.tsx
```diff
@@ -120,10 +121,7 @@ function renderBody({ state, blockId, txHash, type }: Props): React.ReactNode {
       return <SuccessComponent type={type} txHash={txHash!} />;
 
     case "ERROR_SIGN":
-      return <div>Error occurred!</div>;
-
-    case "REVERTED":
```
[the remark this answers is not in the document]
Moejoe90: In reality this simplifies the views there is no need for a separate Reverted state in the views its the same as an `error` from the logic point of view it is different in the sense that the transaction is on blockchain but from the users perspective its the same

app/lib/web3/Web3Manager.ts
```diff
@@ -17,17 +17,20 @@ import { IPersonalWallet } from "./PersonalWeb3";
...
   data?: string;
-  nonce?: string;
-  from?: string;
+  from: string;
   gasLimit?: string;
   input?: string;
 }
 
+export interface IRawTxData extends ITxData {
+  nonce?: string;
```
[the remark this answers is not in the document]
Moejoe90: Good catch it shouldn't be optional as a `raw Transaction` as a regular simple transaction it can as the provider sometimes fills these values like in the case of `metamask`

app/lib/web3/LightWallet.ts
```diff
@@ -218,25 +219,25 @@ export class LightWallet implements IPersonalWallet {
...
-
-    const txData: ITxData = {};
-
-    txData.to = addHexPrefix(data.to!);
-    txData.gasLimit = addHexPrefix(Number(data.gas || 0).toString());
-    txData.gasPrice = addHexPrefix(Number(data.gasPrice || 0).toString(16));
-    txData.nonce = addHexPrefix(Number(data.nonce || 0).toString(16));
-    txData.value = addHexPrefix(Number(data.value! || 0).toString(16));
-    txData.data = addHexPrefix(data.data || "0");
+    txData.nonce = await this.web3Adapter.getTransactionCount(txData.from);
```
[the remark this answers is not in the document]
Moejoe90: done

app/modules/tx/monitor/sagas.ts
```diff
@@ -24,8 +24,9 @@ async function cleanPendingTransactionsPromise(
         tx.transaction.hash,
       );
       // transactionReceipt should be null if transaction is not mined
-      if (!transactionReceipt) return tx;
-      else await apiUserService.deletePendingTx(tx.transaction.hash);
+      if (transactionReceipt && transactionReceipt.blockNumber)
+        await apiUserService.deletePendingTx(tx.transaction.hash);
+      else return tx;
```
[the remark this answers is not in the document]
Moejoe90: This is on purpose so that the values can be filtered later on and removed from the pool. 

if there is a transaction return undefined and later it will be filtered not sure if i made sense

app/modules/tx/sender/sagas.ts
Moejoe90: indeed

app/modules/tx/sender/sagas.ts
Moejoe90: done. 
We need to change how this data structure is populated as in. Take these values from the node directly instead of filling it with manual values which are somewhat arbitrary.

I will add this as todo and a ticket

package.json
Moejoe90: The state of web3 v 1.0 is really dire I must say
Anyway this is just for testing purposes

desfero: Agree, it's really insane that they cannot release finally stable version of 1.0

PR #1387, opened 2018-09-21
Fix imports of external modules inside e2e tests

desfero:


tsconfig.cypress.json
```diff
@@ -1,9 +1,8 @@
 {
```
desfero: @Moejoe90 @lizonp for consistency `tsconfig` was moved to the root

Moejoe90: :+1:

Issue #1375, opened 2018-09-19
Connect Strip View to Smart Contracts


SergejSoft:
@rudolfix pls. paste your info here.

rudolfix:
![image](https://user-images.githubusercontent.com/17202864/45747900-7d6eb380-bc07-11e8-86c9-84125e2610d9.png)

1. We show countdown when there is start_date available. Start data is UTC and we always show UTC date
2. When user is not logged in we show countown **to public phase**
3. When user is logged and can participate in whitelist we show countdown **to pre-eto**
4. if not - we show countdown **to public phase**
5. where are seconds? if we have 0 days please show seconds. we need to have seconds countdown
6. when countdown reaches 0 you need to stop it and display "waiting for next block", you need to constantly poll timedState() and refresh view accordingly

rudolfix:
When user can participate in whitelist
- s/he has icbm wallet
- you call
```
function calculateContribution(address investor, bool fromIcbmWallet, uint256 newInvestorContributionEurUlps)
external
        constant
        returns (
            bool isWhitelisted,
            uint256 minTicketEurUlps,
            uint256 maxTicketEurUlps,
            uint256 equityTokenInt,
            uint256 neuRewardUlps,
            bool maxCapExceeded
            )
```
and check `isWhitelisted`

rudolfix:
![image](https://user-images.githubusercontent.com/17202864/45750702-25887a80-bc10-11e8-9ee8-bae743464547.png)
 - visible in whitelist (pre-eto) only when investor can participate! see above. otherwise countdown
- visible in public phase

Data sources
- number of investors: from `etoCommitment::totalInvestment()`
- ETH: etherToken.balanceOf(etoCommitment.address)
- EUR: euroToken.balanceOf(etoCommitment.address)

green/gray progress bar is constructed from following data:
- starts at 0%
- ends at calculated percentage > 100% which we compute as follows
new_shares_to_issue / minimum_new_shares_to_issue * 100
- green vertical bar at 100% - place it proportionally to the value above.

now where green bar ends calculate from proportion totalTokensInt/(new_shares_to_issue*equity_tokens_per_share)
totalTokensInt you get from `etoCommitment::totalInvestment()`

rudolfix:
So progress bar is done in percentages
![image](https://user-images.githubusercontent.com/17202864/45767960-dc4e2000-bc3b-11e8-8aec-6f4217e579a5.png)

rudolfix:
[...]

PR #1396, opened 2018-09-24
#1375 Improve strip view widget

lizonp:
#1375


app/components/shared/Counter.tsx
```diff
@@ -22,9 +37,17 @@ export class Counter extends React.Component<IProps, IState> {
...
   };
 
+  timer: any = null;
+
   componentDidMount(): void {
-    this.interval = window.setInterval(() => {
-      this.setState(state => ({ date: state.date - 1000 }));
+    const { date } = this.state;
+
+    this.timer = setInterval(() => {
```
desfero: @lizonp there is already a variable called `interval` that is used in `componentWillUnmound`. Please use whether `interval` or `timer`, not both

app/modules/eto-flow/reducer.ts
```diff
@@ -1,42 +1,32 @@
-import { TPartialCompanyEtoData, TPartialEtoSpecData } from "../../lib/api/eto/EtoApi.interfaces";
 import { AppReducer } from "../../store";
 import { DeepReadonly } from "../../types";
 
 export interface IEtoFlowState {
+  etoPreviewCode?: string;
```
desfero: @trival @Moejoe90 eto related data were moved to `public-etos` (probalby that one should be renamed to `etos` or `etos-view` in future) because of the new `contract` field

PR #1409, opened 2018-09-25
Enables Portfolio Feature

Moejoe90:


app/config/.env.example
```diff
@@ -16,7 +16,7 @@ NF_ISSUERS_ENABLED=1
 NF_EQUITY_TOKEN_OFFERINGS_VISIBLE=1
 NF_ETO_LOAD_REAL_DATA=0
 NF_FOUNDERS_BANK_VISIBLE=1
-NF_PORTFOLIO_PAGE_VISIBLE=0
+NF_PORTFOLIO_PAGE_VISIBLE=1
 NF_CHECK_LOCKED_WALLET_WIDGET_ENABLED=1
 NF_WITHDRAW_ENABLED=1
 NF_PORTFOLIO_PAGE_VISIBLE=1
```
lizonp: That flag is duplicated

Moejoe90: Good catch

PR #1418, opened 2018-09-25
updates user on wallet type change

### LINKED ISSUE #1050: update wallet type and subtype when the differ from what is coming from backend
two cases:
1. we update data of user from before this feature was introduced
2. someone change wallet type ie. restoring lightwallet backup codes into nano

https://github.com/Neufund/platform-frontend/issues/1049

rudolfix:
in `auth` saga when wallet type is detected, it's updated in backend

tested manually both on NULL and changed wallet types

fixes #1050

### DIFF
--- a/app/lib/api/users/UsersApi.ts
+++ b/app/lib/api/users/UsersApi.ts
@@ -23,14 +23,12 @@ export class EmailAlreadyExists extends UserApiError {}
 
 const ensureWalletTypesInUser = (userApiResponse: IUser): IUser => ({
   ...userApiResponse,
-  walletType:
-    (userApiResponse.walletType
-      ? (userApiResponse.walletType.toUpperCase() as WalletType)
-      : WalletType.UNKNOWN) || undefined,
-  walletSubtype:
-    (userApiResponse.walletType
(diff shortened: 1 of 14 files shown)


app/lib/api/users/UsersApi.ts
```diff
@@ -21,15 +21,16 @@ export class UserApiError extends Error {}
...
 
-const upperCaseWalletTypesInUser = (userApiResponse: IUser): IUser => ({
+const ensureWalletTypesInUser = (userApiResponse: IUser): IUser => ({
   ...userApiResponse,
   walletType:
-    (userApiResponse.walletType && (userApiResponse.walletType.toUpperCase() as WalletType)) ||
-    undefined,
+    (userApiResponse.walletType
+      ? (userApiResponse.walletType.toUpperCase() as WalletType)
+      : WalletType.UNKNOWN) || undefined,
```
Moejoe90: In this case, there is no need for `|| undefined` anymore as the result of `userApiResponse.walletType?` will either return  `userApiResponse.walletType.toUpperCase()` or `WalletType.UNKNOWN`

rudolfix: true, I will remove it and see what happens

PR #1441, opened 2018-09-27
Restores transaction modal from pending transaction

rudolfix:
1. refactors pending transaction state to separate pending transaction issued by us and out of bound transactions
2. that allows to simplify code in several places
3. PR restores transaction modal in MINING state if pending transaction found
4. old Waiting for pending transaction is shown when there are out of bounds transactions


rudolfix:
this modal for ooo transactions for sure needs to be improved but it's the old modal that we already had...

Issue #1362, opened 2018-09-17
Spinning ETHER logo is broken


falinebambi:
Spinning ETHER logo is broken on here and other places where we used it https://platform.neufund.io/

https://codepen.io/PalmBeachagency/pen/VdzQRN

![screen shot 2018-09-17 at 17 33 39](https://user-images.githubusercontent.com/17235402/45633348-da4f5a00-ba9f-11e8-802f-82366472c569.png)

Moejoe90:
When testing with production build you can run it locally via
```yarn build:prod```
to build the production bundle then
```sh ./scripts/prod-like-srv.sh``` 
to run the server , ill add it to the ticket as well

PR #1455, opened 2018-09-29
#1362 Fix spinning ethereum logo

desfero:


package.json
```diff
@@ -8,6 +8,44 @@
     "> 1%",
     "not IE 11"
   ],
+  "scripts": {
```
desfero: Nothing changed here, only moved `scripts` before any dependencies as I often find annoying to scroll when I want to check what the script is doing.

Issue #1368, opened 2018-09-18
[ETO forms] Key individuals - media links bug


glupiawymowka:
Key individuals is taking media links from Media form, even though it should only be having links to Medium, Twitter and Linkedin
How to reproduce : 
1. Go to "Media" form
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/28092ae1-afe2-4cbf-9abc-b9d8a1e92907)

2. Add media link/s that are different than Medium, Linkedin or Twitter and save

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/ec04cffe-17ea-4c54-9fc5-0102ee804440)

3 Go to "Key Individuals" form and add at least one person without adding social channels
4 Go to "ETO page" and check "Team" 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/3fc2942a-353a-4823-b2ce-6f7cd10873f4)

NOTE: 
I went back to the "Media" form, added another link (instagram). The link was not attached to the Key Individual created BEFORE instagram link was added.
Next, second Key Individual was created and both Media links attached automatically to this one

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/1b0db7f5-2e4b-4187-836a-4d94be6ec16d)

desfero:
@SergejSoft as a note. This will only fix new individuals. All previous ones were added incorrectly (due to the bug in formik) to the database. If there are any advisors on production we can talk with the backend to clean up company `socialChannels` from individuals one.

Moejoe90:
@desfero From my point, i think updating Formik would be important as there might be a couple of other bugs and the changes it introduces should have lots of optimizations. 

The biggest issue i see from my point when doing this is the `Selector` written by thomas more specifically

Issue #1391, opened 2018-09-21
Fix performance problem on ETO Key Individuals form


desfero:
**Background:**
There is a noticeable performance issue on Key Individuals form when there are a lot of individuals.

**Technical notes:**
I can see that there are a lot of repeating each time input change is triggered. Need to play for `shouldComponentUpdate` or `FastField` from formik.

Moejoe90:
As mentioned before the first step to solve any performance issue is updating formik

For example this problem
https://github.com/jaredpalmer/formik/issues/430

SergejSoft:
@desfero  is this solved? If yes, pls. close this ticket

andywalton:
Hi @desfero - I assume this is still a problem?

Alexey, I've added this to the Neufund Platform Multi-Product / Jurisdiction epic, will aim to plan for this time permitting in a future sprint before 15 May 2019's release. 👍

desfero:
@andywalton it's still a problem. 
Just try to submit all fields inside key individuals form and you will see how slow UI respond to keyboard events while typing.

andywalton:
Okay thanks for the update @desfero and for noting it's blocked. We can work on this as a nice to have in future sprints (if possible before we go live with the Multi-product / jurisdiction ETO offering).

PR #1466, opened 2018-10-01
#1368 Update formik library 

desfero:


trival:
would this also solve https://github.com/Neufund/platform-frontend/issues/1391 ?

Issue #1439, opened 2018-09-27
Fix portfolio table layout


desfero:
Background: 
The scroll is visible on the portfolio table. Can you please check why it's there?
![image](https://user-images.githubusercontent.com/7031578/46171976-819f6e80-c2a2-11e8-94f7-644d5d3bb67d.png)

PR #1467, opened 2018-10-01
#1439 - Add blocks loading indicator && fix mocked table

lizonp:
related: #1439


app/components/shared/LoadingIndicator.tsx
```diff
@@ -3,13 +3,34 @@ import * as React from "react";
...
+          <span />
+          <span />
+        </div>
+      )}
+    </>
+  );
+};
+
+LoadingIndicator.defaultProps = {
+  type: "pulse",
```
desfero: Typescript suggest to use default parameters for `defaultProps` when we use stateless functional components to preserve correct typings (currently `type` is `undefined | TLoadingIndicator` inside `LoadingIndicator` but that's not the case because we specify default type`

app/components/shared/LoadingIndicator.tsx
```diff
@@ -3,13 +3,34 @@ import * as React from "react";
...
   light?: boolean;
+  type?: TLoadingIndicator;
 }
 
-export const LoadingIndicator: React.SFC<ILoadingIndicatorProps> = ({ className, light }) => (
-  <div className={cn(className, styles.spinner, { [styles.light]: light })} />
-);
+const LoadingIndicator: React.SFC<ILoadingIndicatorProps> = ({ className, light, type }) => {
+  return (
+    <>
```
desfero: Maybe for readability it's better to use `switch` over ternary operator? What do you think?
```
const LoadingIndicator: React.SFC<ILoadingIndicatorProps> = ({ className, light, type }) => {
  swith(type) {
    case 'pulse':
       return  <div className={cn(className, styles.spinner, { [styles.light]: light })} />;
    case 'blocks':
       return (
        <div className={cn(className, styles.blocks)}>
          <span />
          <span />
          <span />
          <span />
          <span />
        </div>
       )
  }

desfero:
@lizonp does that also fixed horizontal and vertical scroll issue?

app/components/shared/LoadingIndicator.tsx
lizonp: 👍

app/components/shared/LoadingIndicator.tsx
lizonp: 👍

lizonp:
> @lizonp does that also fixed horizontal and vertical scroll issue?

@desfero yes, it does

Issue #1407, opened 2018-09-25
Two new fields in ETO Termsheet


SergejSoft:
https://docs.google.com/presentation/d/1-BReyNlQlIGC5kH3DCvzt0XxcKf2goKse7lzYn4GhlQ/edit#slide=id.g4234efa875_0_0

The last comment on page 16th. 

Simply replicate the two fields and name them as in the comment. Write me or Aga if this is unclear.

@rudolfix already added this fields in the swagger. Pls. ask him if you can't find it.

PR #1468, opened 2018-10-01
#1407 Add new fields to ETO Terms form

lizonpatryk:
related: #1407


app/components/eto/registration/pages/EtoTerms.tsx
```diff
@@ -190,6 +212,12 @@ const EtoRegistrationTermsComponent: React.SFC<IProps> = ({ readonly, savingData
         </Row>
       </Col>
     )}
+
+    <Row className="justify-content-end">
```
desfero: Do we need to have `Save` button twice? (Here and on line 210)?

app/components/eto/registration/pages/EtoTerms.tsx
```diff
@@ -174,6 +174,28 @@ const EtoRegistrationTermsComponent: React.SFC<IProps> = ({ readonly, savingData
       </div>
     </div>
 
+    <Row>
+      <Col xs="12" md="6">
+        <div className="form-group">
```
desfero: `form-group` div is redundant, `FormField` already includes `form-group`

app/lib/api/eto/EtoApi.interfaces.ts
```diff
@@ -203,6 +203,8 @@ export const EtoTermsType = YupTS.object({
   whitelistDurationDays: YupTS.number(),
   additionalTerms: YupTS.string().optional(),
   signingDurationDays: YupTS.number(),
+  newSharesToIssueInFixedSlots: YupTS.number(),
+  whitelistDiscountFraction: YupTS.number(),
```
desfero: image
Should probably default to 0, as empty `%` looks weird

app/components/eto/registration/pages/EtoTerms.tsx
```diff
@@ -174,6 +174,28 @@ const EtoRegistrationTermsComponent: React.SFC<IProps> = ({ readonly, savingData
...
+          />
+        </div>
+      </Col>
+    </Row>
+
+    <div className="form-group">
+      <FormLabel>
+        <FormattedMessage id="eto.form.section.eto-terms.maximum-discount-for-the-fixed-slot-investors" />
+      </FormLabel>
+      <FormRange name="whitelistDiscountFraction" min={0} max={99} unit="%" step={1} />
```
desfero: Is it possible to save the form without error? I always get 
image

rudolfix:
Hey! please do not merge this. Fields are not added to the correct form. I will provide more details later!

rudolfix:
@AlexeyTuboltsev please push this PR forward
* check if @desfero review is implemented and do it if stuff is missing
* i think code that I added to `LegalInformationWidget` should be put somewhere else and `shareholders` collection passed via IProps. you can do it more reactish

Otherwise I moved new fields to correct form and I've added a few other fixes.

e2e tests that fill forms may need upgrade (however new fields are optional)

desfero:
@rudolfix @AlexeyTuboltsev e2e tests should be updated to include missing fields (`FullEtoRegistrationAndSubmit.spec.e2e.ts`), otherwise, they are going to fail after https://github.com/Neufund/platform-frontend/pull/1505 is merged

app/components/eto/publicView/LegalInformationWidget.tsx
```diff
@@ -13,10 +12,20 @@ import * as styles from "./LegalInformationWidget.module.scss";
 
 interface IProps {
   companyData: TCompanyEtoData;
-  etoData: TEtoWithCompanyAndContract;
 }
 
-export const LegalInformationWidget: React.SFC<IProps> = ({ companyData, etoData }) => {
+export const LegalInformationWidget: React.SFC<IProps> = ({ companyData }) => {
+  let shareholdersCopy = companyData.shareholders;
```
desfero: @AlexeyTuboltsev maybe it's worth to move that as a separate function
```
const generateShareholders = (
  shareholders: TCompanyEtoData["shareholders"],
  companyShares: number,
) => {
  const assignedShares = shareholders.reduce((s, v) => s + v.shares, 0);

  if (assignedShares < companyShares) {
    return [
      ...shareholders,
      {
        fullName: "Others",
        shares: companyShares - assignedShares,
      },
    ];
  }

  return shareholders;
};
```

what do you think?

app/test-e2e/utils/forms.ts
```diff
@@ -47,12 +52,29 @@ export const fillForm = (fixture: TFormFixture, { submit = true }: { submit?: bo
...
+        .trigger('change');
+    }
     // click on a field
     else if (field.type === "check") {
       cy.get(formField(key))
         .first()
         .check({ force: true });
     }
+    //check or uncheck a checkbox
+    else if (field.type === "checkBox") {
```
desfero: Good one, but can you also then remove type `check` as it's doing the same (check checkbox)

app/test-e2e/utils/forms.ts
```diff
@@ -4,8 +4,13 @@ import { formField, tid } from "./selectors";
 type TFormFieldFixture =
   | {
       value: string;
-      type: "submit" | "tags" | "file" | "date" | "select" | "check";
+      type: "submit" | "tags" | "file" | "date" | "select" | "check" | "range";
     }
+  | {
+      value: string,
+      setState: boolean,
```
desfero: maybe we can name that field `checked`? Similar to how standard `HTMLCheckbox` work?

app/components/eto/publicView/LegalInformationWidget.tsx
AlexeyTuboltsev: `const generateShareholders = (
  shareholders: TCompanyEtoData["shareholders"],
  companyShares: number,
) => {
  const assignedShares = shareholders.reduce((acc, shareholder) => acc + shareholder.shares, 0);

  if (assignedShares < companyShares) {
    return [
      ...shareholders,
      {
        fullName: "Others",
        shares: companyShares - assignedShares,
      },
    ];
  }

  return shareholders;
};`

app/test-e2e/utils/forms.ts
AlexeyTuboltsev: check does checks the first checkbox in the group, no matter how many there are. And it can't uncheck the box, if it was checked as default value. 
So you're right, we need to unify this both, but I'm kinda afraid to break all our test now, that's why I left it there.

app/test-e2e/utils/forms.ts
AlexeyTuboltsev: well, I'm not sure about this one. I wanted to express an intention - mustBeChecked, or setChecked would be more clear?

app/components/eto/publicView/LegalInformationWidget.tsx
desfero: awesome, i aslo suggest to replace
```  
acc += shareholder.shares;
return acc;
```
by
```
return acc += shareholder.shares;
```
as it's always better to not mutate then mutate ;)

app/test-e2e/utils/forms.ts
desfero: ok, so leave it as it is right now to progress with PR and unify that later ;)

app/test-e2e/utils/forms.ts
desfero: @Moejoe90 what do you think on your side?

app/components/eto/publicView/LegalInformationWidget.tsx
AlexeyTuboltsev: yes, sure

app/test-e2e/utils/forms.ts
AlexeyTuboltsev: yes. there are more problems among these helpers, we need to overhaul all of this at some point. I'm simply not that good at cypress to do it quickly

app/test-e2e/utils/forms.ts
Moejoe90: I would say `mustBeChecked` clearly indicates to me that this value must be true so if that's the case go with that

Issue #1365, opened 2018-09-18
add correct link to download acquisition agreement on investment


rudolfix:
please remove `fufu`

PR #1470, opened 2018-10-02
Add agreement download on investment summary

trival:
related: https://github.com/Neufund/platform-frontend/issues/1365


app/modules/public-etos/sagas.ts
```diff
@@ -183,9 +185,31 @@ function* loadCalculatedContribution(_: TGlobalDependencies, action: TAction): a
   }
 }
 
+function* downloadDocumentByType(_: TGlobalDependencies, action: TAction): any {
```
Moejoe90: This saga name would be a bit confusing especially when using search tools etc
I would prefer the saga name to represent public ETOs as 
`downloadPublicDocumentByType`

PR #1474, opened 2018-10-02
issuer dashboard process fixes

rudolfix:
- aligns when widgets are shown with the banckend workflow
- start/stop bookbuilding only visible when backend allows
- fixes a bug with hidden removal of cap table entry in Leagl dashboard
- provides token holder address to download templates on chain

the last one will need additional handling on the backend. coming soon


app/components/eto/EtoDashboard.tsx
```diff
@@ -258,6 +269,8 @@ export const EtoDashboard = compose<React.SFC>(
       shouldEtoDataLoad: selectShouldEtoDataLoad(s),
       requestStatus: selectKycRequestStatus(s.kyc),
       etoState: selectIssuerEtoState(s),
+      canEnableBookbuilding: (selectIssuerEto(s) || { canEnableBookbuilding: false })
```
Moejoe90: I would move this is a `selector` in selectors as its better to have no logic here. 
The selector can return a default value of false

app/components/shared/forms/formField/FormCategoryDistribution.tsx
```diff
@@ -85,7 +85,7 @@ const SingleCategoryDistributionComponent: React.SFC<IProps & IInternalProps> =
         </Col>
         {!isFirstElement && (
           <span className="pt-2">
-            {disabled && <ButtonIcon svgIcon={closeIcon} onClick={removeField} />}
+            {!disabled && <ButtonIcon svgIcon={closeIcon} onClick={removeField} />}
```
Moejoe90: :+1:

app/components/eto/EtoDashboard.tsx
```diff
@@ -113,13 +117,12 @@ const EtoStateViewRender: React.SFC<IEtoStateRender> = ({
     case EtoState.PREVIEW:
       return (
         <>
-          {shouldViewSubmissionSection && (
+          {(shouldViewSubmissionSection || !isTermSheetSubmitted) && (
```
Moejoe90: This shouldn't be an `||` but an `&&` as you want `shouldViewSubmissionSection` to be true before showing the submission section

PR #1481, opened 2018-10-02
Sets optional fields in issuer dashboard

rudolfix:
I sat down with aga and marked all fields as required or optional. Please merge it into your form fixes


app/test-e2e/etoRegistration/fixtures.ts
```diff
@@ -86,11 +86,6 @@ export const etoTermsForm: TFormFixture = {
 // key individuals
 // @TODO also upload individuals fotos and add social channels
 export const etoKeyIndividualsForm: TFormFixture = {
-  "team.members.0.name": "Zoe",
```
desfero: We should not remove fields from fixtures. The correct way is to update e2e test to click on `Add` button for `Team` section before filling fields (see how advisors are implemented)

app/test-e2e/etoRegistration/fixtures.ts
```diff
@@ -85,12 +85,15 @@ export const etoTermsForm: TFormFixture = {
 
 // key individuals
 // @TODO also upload individuals fotos and add social channels
-export const etoKeyIndividualsForm: TFormFixture = {
+
+export const etoKeyIndividualsFormOptional: TFormFixture = {
```
desfero: All of the fields are optional here (Individuals forms) so there is no need to extract some fields as a separate object.

desfero: All of the fields are optional inside Key Individuals form, so there is no need to have `etoKeyIndividualsFormOptional` object. That means we can also get rid of `  fillForm(etoKeyIndividualsFormOptional, false);`.

desfero:
@Moejoe90 @rudolfix working on a fix

Issue #1462, opened 2018-10-01
smart contract info for on chain eto is not show in preview link


rudolfix:
1. you need to be logged out
2. go to preview
3. all stuff from eto.contract state part is not displayed

my take is that you rely on previously loaded state - probably in eto listing view (dashboard)

https://platform.neufund.io/eto-preview/9ac0d353-3ddf-48a4-b13b-e64a32e89840. you should probably push single eto into public-etos state on load here

PR #1492, opened 2018-10-04
#1462 Get contract data for eto preview

desfero:
#1462


app/test-e2e/utils/index.ts
```diff
@@ -29,16 +29,23 @@ export const clearEmailServer = () => {
...
-      | undefined;
-
-    expect(response && response.email).to.be.eq(userEmail);
-    expect(loginLink).to.contain("salt");
-  });
+export const assertWaitForLatestEmailSentWithSalt = (userEmail: string, timeout: number = 20000) => {
+  let response: { email: string | undefined } | undefined;
+  let remaining = timeout;
+  let loginLink;
+  while (!response) {
```
Moejoe90: I'm a bit worried from this. What happens if there is no response. Or an error occurs

PR #1493, opened 2018-10-04
Run Cypress tests in parallel

desfero:


desfero:
Need to get paid plan before parallelization can be turned on

Moejoe90:
:(

PR #1500, opened 2018-10-05
Adds waitedClicked for E2E tests

Moejoe90:


rudolfix:
I really do not think that wait before click is really a problem

Moejoe90:
@rudolfix It is a problem in some cases for example 
`Access Wallet Modal` 
I think most of these issues come one way or another with modals

app/test-e2e/utils/index.ts
```diff
@@ -102,18 +102,18 @@ export const typeLightwalletRecoveryPhrase = (words: string[]) => {
...
-  cy.get(tid("btn-send")).click();
+  cy.get(tid("btn-send")).awaitedClick();
 };
 
 export const confirmAccessModal = (password: string) => {
   cy.get(tid("access-light-wallet-password-input")).type(password);
   cy.get(tid("access-light-wallet-confirm"))
-    .wait(1500)
-    .click();
+    .wait(1000)
```
desfero: hm, can we remove `wait` here and add param `timeToWait = 1000` so we can hide all `wait`s inside `.awaitedClick`? 
```
.awaitedClick(2000);
```

what do you think?

Moejoe90: good idea

PR #1505, opened 2018-10-08
Check percentage after submitting eto section

desfero:


Moejoe90:
@desfero 
What's the goal exactly of this PR?

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -205,7 +210,7 @@ class SocialProfilesEditorLayout extends React.Component<IProps & TFormikConnect
     const { selectedFields } = this.state;
 
     return (
-      <div className={className}>
+      <section data-test-id={name} className={className}>
```
Moejoe90: :+1:

app/test-e2e/etoRegistration/FullEtoRegistrationAndSubmit.spec.e2e.ts
```diff
@@ -46,25 +45,19 @@ describe("Eto Forms", () => {
     cy.get(tid("key-individuals-group-button-partners")).click();
     fillForm(etoKeyIndividualsForm);
```
Moejoe90: Is there a reason why this is not `fillAndAssert`?

desfero: @Moejoe90 because of the click functionality on `Add` button. For now, i left as it is but in the near future want to create some kind of custom type functionality for forms so we can hide clicking under some methods

PR #1506, opened 2018-10-08
Add bundle-analyzer

desfero:
In overall bundle look ok, but for some reason, lodash is duplicated.
We need to think about splitting bundle by at least 3 separate files (vendor, app, assets) as the current size of the bundle is enormous 1.5M gzipped :D

image


Moejoe90:
@desfero whats the best approach for bundle separation I don't have much experience in this

Moejoe90:
I merged this and will create a new ticket for discussion

Issue #1343, opened 2018-09-13
Do not use non-interactive tags for click handlers


desfero:
**Background:**
There are a couple of places where non-interactive elements attach click event handlers (for. e.g change/delete pledge when book-building). 
It causes usability issues that should be avoided as it's not possible to interact with them through the keyboard and it's considered an overall bad practice.

PR #1521, opened 2018-10-10
#1343  Do not use non-interactive tags for click handlers

AlexeyTuboltsev:
replace \<span\> and \<div\> with onclick handler to \<a\>s
sometimes involves rewriting respecting css, because \<a\>s take font, color, etc from parent element.
add some translations in EtoDocuments on the way
add a couple of stories


app/components/walletSelector/ledger/WalletLedgerChooserComponent.tsx
```diff
@@ -70,26 +70,22 @@ export const WalletLedgerChooserComponent: React.SFC<
...
         {!advanced && (
           <div className={styles.advanced}>
-            <span onClick={handleAdvanced} data-test-id="btn-advanced-simple">
+            <a onClick={handleAdvanced} data-test-id="btn-advanced-simple">
               <FormattedMessage id="wallet-selector.ledger.advanced-options" />
-              <i className="fa fa-chevron-down ml-2" aria-hidden="true" />
-            </span>
+            </a>
+            <i className="fa fa-chevron-down ml-2" aria-hidden="true" />
           </div>
```
trival: could this icon also be inside the link, or is there some issue with this, the other icon further down is inside.

app/components/Documents.tsx
```diff
@@ -65,7 +80,7 @@ export const GeneratedDocuments: React.SFC<{
 }> = ({ document, generateTemplate }) => {
   return (
     <Col xs={6} lg={3} key={"url"} className="mb-4">
-      <div
+      <a
```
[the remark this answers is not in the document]
AlexeyTuboltsev: well, they are focusable on chrome. I can't check in other browsers now. To make focus visible, we need to add an :onFocus css rule. The same is with button - focus is invisible, unless an :onFocus exists. So it's mostly the same amount of work to change to buttons or to As. But I can do it...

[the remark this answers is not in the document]
desfero: Strange... :) try to focus placeholder link on that jsbin https://jsbin.com/burutugage/edit?html,output  Does that work on your computer?

intl/locales/en-en.json
```diff
@@ -48,7 +48,12 @@
   "dashboard.neufund-kpi-widget.label5": "Number of ETOs",
   "dashboard.neufund-kpi-widget.label6": "TOTAL CAPITAL INVESTED",
   "dashboard.neufund-kpi-widget.list.label": "Total Proceeds<br />distributed",
+  "documents.agreement-and-prospectus-templates": "Agreement and prospectus templates",
+  "documents.approved-prospectus-and-agreements-to-upload": "Approved prospectus and agreements to upload",
+  "documents.download-document": "download",
+  "documents.generated-documents": "generated documents",
   "documents.legal-documents": "LEGAL DOCUMENTS",
+  "documents.please-fill-the-eto-forms-in-order-to-generate-templates": "",
```
desfero: is there a reason that it's not set to `Please fill the ETO forms in order to generate templates`?

app/components/walletSelector/ledger/WalletLedgerInitComponent.tsx
```diff
@@ -101,19 +101,23 @@ export class WalletLedgerInitComponent extends React.Component<
...
+          <button
+            onClick={this.handleShowSupport}
+            className={cn(
+              styles.drawerButton,
+              this.state.showLedgerBrowserSupport ? "mt-3 mb-4" : "mt-3",
+            )}
+          >
+            <FormattedMessage id="wallet-selector.ledger.start.browser-support" />
             <i
+              aria-hidden="true"
```
desfero: 👍

Issue #1510, opened 2018-10-09
Html nesting error in EtoMedia.tsx


AlexeyTuboltsev:
EtoMedia.tsx:
markup nesting error, \<p/\> cannot hold other elements
need to change all \<p/\> to \<div/\> in this component.

![screen shot 2018-10-09 at 15 46 30](https://user-images.githubusercontent.com/14041336/46673865-14150b80-cbdb-11e8-8d81-d0b466e8be96.png)

PR #1523, opened 2018-10-10
#1510 Html nesting error in EtoMedia.tsx

AlexeyTuboltsev:
#1510 

turn \<p\> into \<div\>s.
add keys to MediaLinkEditor FormFields


app/components/eto/registration/pages/EtoMedia.tsx
```diff
@@ -37,31 +37,33 @@ const EtoRegistrationMediaComponent = ({ savingData }: IProps) => (
     progressOptions={etoMediaProgressOptions}
   >
     <Section>
-      <p className="offset-1 mb-2 font-weight-bold">
+      <div className="offset-1 mb-2 font-weight-bold">
```
desfero: Do we need replace every `p` by `div`? P
Probably the issue is only with `Tooltip` component on line 52

trival: It throws a lot of console errors... if it holds more elements, it shouldn't be a 'p'

trival: but you are right, if it only contains a translation string, it should can stay a 'p', it is more semantic then

AlexeyTuboltsev: well, my thinking was, that these are not paragraphs in a text, but markup blocks, so the shouldn't be Ps. I don't insist though

desfero: Only 'div's on line 88 and 99 should be left as `p`. All other 'div's should be either 'div', 'label', 'legend' tags. But for now let's leave those `label` and `legends` as `div`s as fixing markup can take some time :( 
To summarise, let's revert back `div`s on line 88 and 99 to `p` and merge (and keep in mind that in future we need to fix markup :D)

is it ok? @trival @AlexeyTuboltsev everyone agrees?

app/components/shared/MediaLinksEditor.tsx
```diff
@@ -45,12 +45,14 @@ const SingleMediaLinkField: React.SFC<
 
             return isFieldWithCharactersLimit ? (
               <FormField
+                key={`${props.name}.${fieldName}`}
```
desfero: 👍

app/components/eto/registration/pages/EtoMedia.tsx
AlexeyTuboltsev: yes, I'm totally ok with it. I'll correct it

Issue #1445, opened 2018-09-28
Strip View. Claim Token


SergejSoft:
Please make claim token only visible for people invested in this ETO.

An investor can see campaigns. Take part in bookbuilding. If taken part in bookbuilding than he automatically can take part in pre-sale. And in public sale. The claim is only possible if you have invested.

If the user did not invest, simply show success element but not the button "claim your token."

![image 6](https://user-images.githubusercontent.com/2290474/46193139-6b72cc00-c2fd-11e8-998e-6b6eed291627.png)

PR #1527, opened 2018-10-10
#1445 Hide claim button if not invested in eto

desfero:
#1445 

80% of changes are related to moving from `TUserType` to `EUserType`.


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2015",
+    "target": "es2016",
```
desfero: To get support for Array.prototype.includes

trival: but is that ok with our platform requirement of supported browsers? I heard from Chris that es2015 is the targeted platform, but I don't know exactly

desfero: Array.prototype.includes is supporter everywhere so should be ok

Issue #1476, opened 2018-10-02
Public Page Stip View Fixes


SergejSoft:
- [x] Proper fields linking:

Please read again ticket https://github.com/Neufund/platform-frontend/issues/1375 and implement everything as described there. In view above for example: links to documents are totally wrong and counter works for sure until wrong date (probably until whitelist start not signing start)

Bug: Right now you just output state name like “SETUP”. There must be a mapping to strings, also refer to designs for correct names (SETUP is Campaigning in designs). In other words current implementation is a shortcut. We need to take time and improve this.
![image](https://user-images.githubusercontent.com/2290474/46350103-28de2600-c654-11e8-978b-744145927e12.png)

- [x] Convert values:
Please convert values from “wei” into floats before displaying. Also please unify how money (ether/eur) are displayed. It’s clear from the design that we use space as group separator. For example 100 000 000.

![image](https://user-images.githubusercontent.com/2290474/46350199-6642b380-c654-11e8-87a7-79162f4f1c3f.png)

- [x]  Timeline:
* •	Past dates are green, future dates are orange!
* •	also why there is no payout deadline on current public profile? See designs. You have the start date of PAYOUT state – so it’s super easy to implement
* •	and here you have correct state names. State on strip view must match!

![image](https://user-images.githubusercontent.com/2290474/46350247-91c59e00-c654-11e8-8bf7-8098f9078c17.png)

- [x] documents are missing. 
Here links do documents are missing. In here we link to templates. Must be loaded as pdfs. Moe know how to generate a link do please do it. Templates are always there.
* •	EQUITY link to template: investment_and_shareholder_agreement
* •	Token Sale: reservation_and_acquisition_agreement
* •	Token holder Right: template of type: company_token_holder_agreement

Also investment amount – on designs it’s in millions so please do as it says!
![image](https://user-images.githubusercontent.com/2290474/46350371-e832dc80-c654-11e8-9b60-3196d5ad1cb5.png)

desfero:
@SergejSoft 
```
counter works for sure until wrong date (probably until whitelist start not signing start)
```
Are you sure that we should show signing start? For my understanding, there should be `Public` start date shown when ETO is in 'Whitelisted' state

desfero:
@SergejSoft i going to also update how status is displayed on the issuer dashboard 
![image](https://user-images.githubusercontent.com/7031578/46461149-46c79a00-c7be-11e8-84a8-4e9afa3f7477.png)

[...]

PR #1538, opened 2018-10-11
#1476 Correct invested amount visuals

desfero:


app/components/eto/shared/InvestmentAmount.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type TExternalProps = {
+  preMoneyValuationEur: number | undefined;
+  existingCompanyShares: number | undefined;
+  newSharesToIssue: number | undefined;
+  minimumNewSharesToIssue: number | undefined;
+};
+
+const InvestmentAmount: React.SFC<TExternalProps> = ({
+  preMoneyValuationEur = 1,
```
desfero: @rudolfix as a note: I preserve the default behavior but probably it's not quite correct. Try to use `ISSUER_PREVIEW` fixture and see how `Invested amount` looks when values are not yet provided: 
image
Probably it should be set to `N/A`?

rudolfix:
that's other ticket @Moejoe90

Issue #1425, opened 2018-09-26
Console errors


desfero:
1. ETO view
```
warning.js:33 Warning: React does not recognize the `allowTransparency` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `allowtransparency` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
    in iframe (created by TwitterTimelineEmbed)
    in TwitterTimelineEmbed (created by Component)
    in div (created by Component)
    in div (created by Component)
    in Component (created by Component)
    in div (created by Col)
    in Col (created by Component)
    in div (created by Row)
    in Row (created by Component)
    in div (created by Component)
    in Component (created by nest(Component, Component))
    in div (created by Col)
    in Col (created by Component)
    in div (created by Row)
    in Row (created by Component)
    in div (created by Component)
    in div (created by Component)
    in Component (created by nest(Component, Component))
    in nest(Component, Component) (created by Connect(nest(Component, Component)))
    in Connect(nest(Component, Component)) (created by OnEnterAction)
    in OnEnterAction (created by Connect(OnEnterAction))
    in Connect(OnEnterAction) (created by Route)
    in Route (created by OnlyAuthorizedRouteComponent)
    in OnlyAuthorizedRouteComponent (created by Connect(OnlyAuthorizedRouteComponent))
    in Connect(OnlyAuthorizedRouteComponent) (created by Component)
    in Switch (created by Connect(Switch))
    in Connect(Switch) (created by Component)
    in Component (created by AppComponent)
    in ScrollToTopComponent (created by Connect(ScrollToTopComponent))
    in Connect(ScrollToTopComponent) (created by AppComponent)
    in AppComponent (created by Connect(AppComponent))
    in Connect(AppComponent) (created by OnEnterAction)
    in OnEnterAction (created by Connect(OnEnterAction))
    in Connect(OnEnterAction) (created by HotExportedConnect(OnEnterAction))
    in AppContainer (created by HotExportedConnect(OnEnterAction))
    in HotExportedConnect(OnEnterAction)
    in IntlContainerInjector (created by Component)
    in Component (created by InjectIntl(Component))
    in InjectIntl(Component) (created by Component)
    in IntlProvider (created by Component)
    in Component
    in Router (created by ConnectedRouter)
    in ConnectedRouter
    in InversifyProvider
    in Provider
```

rudolfix:
[...]

PR #1542, opened 2018-10-12
#1425 Fix console errors

desfero:
#1425


desfero:
ok seems something is broken, checking...

desfero:
false alarm, some non deterministic errors :D

PR #1543, opened 2018-10-12
Fix visual regression test of FormFieldColorful

desfero:


desfero:
@Moejoe90 confetti is still not deterministic but i don't know why :D

Issue #976, opened 2018-07-02
read wallet type from e-mail activation link and set specific wallet type on login


rudolfix:
I registered with nano, try to verify e-mail
Click on the link and go to login page **of light wallet**. this is super confusing.
we'll add "wallet_type" query parameter to login link from which you can take wallet type
then please set correct tab!

here is backend issue that needs to be solved Neufund/platform-backend/ issue 697

rudolfix:
so we are already sending this in the e-mails. when this is done we can finally do it
https://github.com/Neufund/platform-frontend/issues/1049

rudolfix:
BUMP! why it's not fixed yet? super easy to do now

Moejoe90:
@rudolfix 
I can't do this ticket because i'm not getting the wallet type in the login link i'm only getting it in the activation link
![image](https://user-images.githubusercontent.com/23242822/46817323-30ea4400-cd7f-11e8-86ec-d7c3f6b57cbc.png)

PR #1546, opened 2018-10-12
Directs to correct wallet when clicking activation link

Moejoe90:
Related https://github.com/Neufund/platform-frontend/issues/976

Things were cleaned a bit this relies basically on `wallet_type` from the login link `wallet_subtype` has no effect I think its fine this way as `walletType` is what matters.


app/components/shared/routing/OnlyAuthorizedRoute.tsx
```diff
@@ -43,9 +46,20 @@ class OnlyAuthorizedRouteComponent extends React.Component<TProps, IState> {
...
   }
 
-  private getRedirectionPath(): LocationDescriptor {
+  private selectRouteBasedOnWalletType = (walletType: EWalletType): string => {
+    switch (walletType) {
+      case EWalletType.LEDGER:
+        return loginWalletRoutes.ledger;
+      case EWalletType.BROWSER:
+        return loginWalletRoutes.browser;
+      default:
```
desfero: as a note, current approach is ok, but i think we can get better typings with `never` type

```
  private selectRouteBasedOnWalletType = (walletType: EWalletType): string => {
    switch (walletType) {
      case EWalletType.LEDGER:
        return loginWalletRoutes.ledger;
      case EWalletType.BROWSER:
        return loginWalletRoutes.browser;
      case EWalletType.LIGHT:
         return loginWalletRoutes.light;
       default:
         throw new Error("Route should be provided for wallet");
    }
}
```


Doing that way while adding new value to `EWalletType` enum typescript will throw compile time error.

app/components/shared/routing/OnlyAuthorizedRoute.tsx
```diff
@@ -57,14 +71,17 @@ class OnlyAuthorizedRouteComponent extends React.Component<TProps, IState> {
...
       ...rest
     } = this.props;
     const { redirect } = this.state;
     const InvestorComponentAsAny = InvestorComponent as any;
     const IssuerComponentAsAny = IssuerComponent as any;
 
     if (redirect) {
-      return <Route {...rest} render={() => <Redirect to={this.getRedirectionPath()} />} />;
+      return (
+        <Route {...rest} render={() => <Redirect to={this.getRedirectionPath(walletType)} />} />
```
desfero: I know that changes are not yours, but i'm not sure whether we need `Route` here if we then immediately render `Redirect`

app/components/walletSelector/WalletRouter.tsx
```diff
@@ -11,11 +11,13 @@ interface IProps {
...
-export const WalletRouter: React.SFC<IProps> = ({ rootPath }) => (
-  <SwitchConnected>
-    <Route path={`${rootPath}/light`} component={WalletLight} />
-    <Route path={`${rootPath}/browser`} component={WalletBrowser} exact />
-    <Route path={`${rootPath}/ledger`} component={WalletLedger} exact />
-    <Redirect to={getRedirectionUrl(rootPath)} />
-  </SwitchConnected>
-);
+export const WalletRouter: React.SFC<IProps> = ({ rootPath }) => {
+  return (
```
desfero: It's your preference to add `return`? Because i always remove `return` if can :D so we need to align that to go into endless loop of adding/removing :D

app/test-e2e/routing/routing.spec.e2e.ts
```diff
@@ -0,0 +1,53 @@
+import { registerWithLightWallet } from "../utils";
+import { tid } from "../utils/index";
```
desfero: it's the same import as the above one.

app/test-e2e/routing/routing.spec.e2e.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const assertUserInBrowserWalletLoginPage = () => {
+  cy.get(tid("modals.wallet-selector.wallet-browser.title"));
+};
+
+const assertUserInLedgerWalletLoginPage = () => {
+  cy.get(tid("modals.wallet-selector.ledger-wallet.title"));
+};
+
+describe("Platform Routing Tests", () => {
```
desfero: `Platform Routing Tests` is quite generic name and someone (even we) in future can forget the purpose.
Maybe `Authentication Routing Tests`?

app/components/shared/routing/OnlyAuthorizedRoute.tsx
Moejoe90: But the thing is there is a wallet type unknown. The reason for this is in previous iterations there were wallets with no type. 

Even though everywallet that registers now will have a type old activation links and so remain with no wallet. So the idea of never is good but not for this case

app/components/walletSelector/WalletRouter.tsx
Moejoe90: Actually it's not my preference good catch i was debugging and missed putting it back again

app/components/shared/routing/OnlyAuthorizedRoute.tsx
Moejoe90: Good point i will keep it as reference

app/test-e2e/routing/routing.spec.e2e.ts
Moejoe90: You have a point here

app/test-e2e/routing/routing.spec.e2e.ts
Moejoe90: Good catch.

app/components/shared/routing/OnlyAuthorizedRoute.tsx
desfero: yeap, you are right, UNKNOWN break everything :(

app/components/walletSelector/WalletRouter.tsx
desfero: I used always to debug props throught react devtool, give them a try

PR #1561, opened 2018-10-15
Fixes Initialization Steps

Moejoe90:
This fixes one issue related with the init reducer. When logging out all the state gets reset into its initial state. However `smartContractInit` is designed for optimization purposes to only run one time in the whole life cycle of the app. 

This PR solves this also solves wallet loading mechanism so that it only starts when the user is logged in. It also uses is smart contract initialized selector


app/modules/wallet/sagas.ts
```diff
@@ -77,6 +78,12 @@ export async function loadWalletDataAsync(
 }
 
 function* walletBalanceWatcher(): any {
+  const isSmartContractsInitialized = yield select(selectIsSmartContractInitDone);
```
desfero: @Moejoe90 @trival that's why i prefer for selectors to receive whole store as an argument

PR #1566, opened 2018-10-16
Bookbuilding e2e tests

desfero:


app/test-e2e/utils/userHelpers.ts
```diff
@@ -17,42 +19,49 @@ export const createAndLoginNewUser = async (
...
+    // mark backup codes verified
+    await markBackupCodesVerified(jwt);
+
+    if (params.clearPendingTransactions) {
+      await clearPendingTransactions(jwt, address);
+    }
+  });
 
-  cy.log(`Created wallet and ${userType} user with address ${address}`);
+  cy.visit(appRoutes.root);
```
Moejoe90: I wouldn't add the `visit` here it's better to keep it with each test on its own as this is just a preparation function

app/test-e2e/utils/userHelpers.ts
```diff
@@ -17,42 +19,49 @@ export const createAndLoginNewUser = async (
...
+  cy.get(tid("Header-logout")).should("exist");
```
Moejoe90: Same here its better not have this assertion

app/test-e2e/utils/userHelpers.ts
desfero: that makes sense, I thought it will not work without assertion but it should wait for all requests because of the wrapped `.then`

Issue #1578, opened 2018-10-16
do not sent authorization headers to ipfs download methods


rudolfix:
when logged out in public profile:
https://platform.neufund.net/eto-view/0x164e07Ae48cA7774663e90732d44b324f2e3C679
Click on reservation and acquisition agreement. It fails with 401
https://platform.neufund.net/api/immutable-storage/download/QmVPt2xpu4PjmBhDnX1MBijDHPo5rG4GBMqJWxNRfTDEhu?as_pdf=true&mime_type=application%2Fvnd.openxmlformats-officedocument.wordprocessingml.document&placeholders=
```
{,…}
debug: "Exception type#<class 'platform_common.jwt_validation.JwtInvalidTokenException'>#app.py#1612"
status: 400
title: "Not enough segments"
type: "platform_common.jwt_validation.JwtInvalidTokenException"
```

that's because of following header:
authorization: Bearer undefined`
where undefined stands for login cookie which of course is not available.
download is anonymous method and if JWT is not present do not set Authorization header at all

PR #1584, opened 2018-10-16
simplifies documents download

rudolfix:
fixes: #1578 
also fixes
- templates were downloaded instead of documents (in several places)
- eto/pr eto countdown was not shown in setup
- isEligibleForEto slightly improved (but still needs update)

that's needed for tomorrows tests

### DIFF
--- a/app/lib/api/ImmutableStorageApi.ts
+++ b/app/lib/api/ImmutableStorageApi.ts
@@ -14,7 +14,7 @@ const UPLOAD_DOCUMENT_PATH = "/upload";
 export class ImmutableStorageApi {
   constructor(
     @inject(symbols.authorizedJsonHttpClient) private httpClient: IHttpClient,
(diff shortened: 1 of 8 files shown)


rudolfix:
here is explanation of documents vs. templates
https://github.com/Neufund/platform-backend/wiki/5.8.-Use-case-Document-Templates-Lifecycle

app/lib/api/ImmutableStorageApi.ts
```diff
@@ -14,7 +14,7 @@ const UPLOAD_DOCUMENT_PATH = "/upload";
 export class ImmutableStorageApi {
   constructor(
     @inject(symbols.authorizedJsonHttpClient) private httpClient: IHttpClient,
-    @inject(symbols.authorizedBinaryHttpClient) private binaryHttpClient: IHttpClient,
+    @inject(symbols.binaryHttpClient) private binaryHttpClient: IHttpClient,
```
Moejoe90: :+1:

Issue #1388, opened 2018-09-21
Make Strip View Widget for corporate site.


SergejSoft:
We need to have a strip view widget for different ETOs. 
First, one to come will be our own strip view, second will be Founders Bank.

<img width="640" alt="screen shot 2018-09-20 at 17 14 10" src="https://user-images.githubusercontent.com/2290474/45828361-a327b580-bcf8-11e8-9eb3-7276932202b5.png">

This will be placed on our  corporate website neufund.org 
and on the website of our partner: Founders Bank and Maltese Stock Exchange.  
So we need to design iFrame

desfero:
Duplicate of https://github.com/Neufund/platform-frontend/issues/1271?

SergejSoft:
@falinebambi we need to have fixed with and hights for the iframe. Now all widgets have a different site.

falinebambi:
Link paths: 
Click CTA button and user not logged in goes to register page on platform 
Click CTA button and user logged in goes to listing page on platform
User clicks anywhere else on widget and goes to public listing page 

Slack me if you have any questions

falinebambi:
No personal data of login state is carried through in the widget
CTA always shows Invest Now
but you can not invest directly through the widget
it directs you to the platform 
(click "invest now" on the widget and then click "invest now" second time on the widget on public listing page)
the wording invest now on the external widget is there because it has higher marketing traction

banciur:
Done but PR was not connected to issue

PR #1591, opened 2018-10-17
#1388 Strip View Widget - Phase 1

drakkein:
Realated to #1388


drakkein:
I've created test bin with CSS proposed by @falinebambi 
https://jsbin.com/letewenuwo/edit?html,css,output

This is not what we wanted to achieve. I'll try to set some media queries.

drakkein:
There are some visual regression issues, due of background change. I think we should set bg for storybook to make it pass. @desfero

app/components/eto/EtoWidgetView.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    return (
+      <Col xs={12}>
+        <div className="mb-3">
+          <EtoOverviewStatus
+            preMoneyValuationEur={eto.preMoneyValuationEur}
+            existingCompanyShares={eto.existingCompanyShares}
+            equityTokensPerShare={eto.equityTokensPerShare}
+            minimumNewSharesToIssue={eto.minimumNewSharesToIssue}
+            contract={eto.contract!}
```
Moejoe90: Is contract address optional in the data structure?
If so i would investigate if the data structure is not precise.
Or add a fallback value if it's undefined, or change the interface of the component

app/components/eto/EtoWidgetView.tsx
```diff
@@ -0,0 +1,91 @@
...
+  appConnect<IStateProps, IDispatchProps, IRouterParams>({
+    stateToProps: (state, props) => ({
+      userType: selectUserType(state.auth),
+      eto: selectEtoWithCompanyAndContractById(state, props.etoId),
+    }),
+  }),
+  onEnterAction({
+    actionCreator: (dispatch, props) => {
+      dispatch(actions.publicEtos.loadEto(props.etoId));
+      dispatch(actions.wallet.loadWalletData());
```
Moejoe90: Why do you need this action?
Does the component use wallet data of any sort?

app/components/eto/EtoWidgetView.tsx
drakkein: In props interface contact could be undefinde, so this non-null assertion is unnecessary.

app/components/eto/EtoWidgetView.tsx
drakkein: This is leftover after commit #39ee1ffbf2e5dbd0bce154e161f909472f1607fb where wallet was required in ETOOverviewStatus. Good catch.

Issue #1556, opened 2018-10-15
[ETO OPPORTUNITIES] - Add founders quote + section separators 


glupiawymowka:
1. Last ETO opportunity section should never be blank. If there is no ETO status to display yet, it should always contain a FOUNDERS QUOTE. Here you can check the design : https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5b969ed70c99e06799f68416

2. Some ETO opportunities are missing one or more section separators (first one is correct, but missing the quote):

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/c169e955-b371-4e96-b276-be1542e9c5e2)
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/296cab12-0c08-4332-bc16-40dfa4f67522)

PR #1607, opened 2018-10-18
#1556 Add founders quote to eto listing state

desfero:
#1556


app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
```diff
@@ -152,82 +262,21 @@ const EtoOverviewStatusLayout: React.SFC<
...
-              tokenName={props.tokenName}
-              totalInvestors={props.contract!.totalInvestment.totalInvestors.toNumber()}
-              totalEquivEurUlps={props.contract!.totalInvestment.totalEquivEurUlps}
-              timedState={timedState}
-            />
-          )}
-          {timedState === ETOStateOnChain.Refund && (
-            <RefundWidget etoId={props.etoId} timedState={timedState} />
-          )}
+          <EtoStatusManager {...props} />
```
desfero: Moved to separate component for readability (maybe i even move that outside of that file in future)

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
```diff
@@ -65,14 +73,116 @@ const PoweredByNeufund = () => {
...
+const EtoStatusManager = (
+  props: IProps & CommonHtmlProps & IWithIsEligibleToPreEto & IStateProps,
+) => {
+  // It's possible for contract to be undefined if eto is not on chain yet
+  const timedState = props.contract ? props.contract.timedState : ETOStateOnChain.Setup;
+
+  switch (timedState) {
+    case ETOStateOnChain.Setup: {
+      if (props.isAuthorized) {
+        if (props.campaigningWidget.isActivated) {
```
drakkein: There is still a lot of nested ifs, maybe we could create something like AuthorizedStatusComponent and move this block into it?

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
```diff
@@ -65,14 +73,116 @@ const PoweredByNeufund = () => {
...
+            stage="ETO"
+          />
+        );
+      }
+    }
+
+    case ETOStateOnChain.Public: {
+      return (
+        <InvestWidget
+          raisedTokens={parseInt(
```
drakkein: Why we use parsing here instead of Number constructor? Does this value require parsing? Number can handle exponential notation. 🤔

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
desfero: I don't know right now the best way to accomplish that as those components are not interconnected so i need to pass a lof of props to get rid of conditionals. Maybe for now let's leave it as it is (or maybe you have some idea how to refactor that correctly)?

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
desfero: fixed, can be simplified to `toNumber`

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
drakkein: Other than passing props... Hmmm... I don't have better idea.

PR #1634, opened 2018-10-22
Update en-en.json

meneramandarin:
wording changes regarding legal matters


intl/locales/en-en.json
```diff
@@ -576,14 +576,14 @@
...
   "settings.backup-seed-widget.view-again": "View again",
   "settings.backup-seed-widget.write-down-recovery-phrase": "Write down your recovery phrase and keep it in a secure place. In case you should ever forget your password you can recover your Light Wallet with the recovery phrase.",
-  "settings.book-building-widget.manually-stop": "You can choose manually to stop the bookbuilding process. The Pre-ETO will start 2 weeks after this date.",
-  "settings.book-building-widget.proposal-accepted": "Your proposal has been accepted and your company profile is now public. The next step is now bookbuilding.",
-  "settings.book-building-widget.start-book-building": "Start bookbuilding",
-  "settings.book-building-widget.stop-book-building": "Stop bookbuilding",
+  "settings.book-building-widget.manually-stop": "You can choose manually to stop the whitelisting process. The Pre-ETO will start 2 weeks after this date.",
+  "settings.book-building-widget.proposal-accepted": "Your proposal has been accepted and your company profile is now public. The next step is the whitelisting.",
+  "settings.book-building-widget.start-book-building": "Start whitelisting",
```
desfero: @Moejoe90 i do think we need a story to align naming of files around bookbuilding as it's renamed by business to whitelisting?

Moejoe90: I think just need to make sure that both are the same thing,

For i had initially the idea that bookbuilding is different. If that's not the case then please push it forward

meneramandarin:
yes, we decided to rename bookbuilding into whitelisting. it's the same thing with a different name. aga asked me to change it.

intl/locales/en-en.json
meneramandarin: yes, we decided to rename bookbuilding into whitelisting. it's the same thing with a different name. aga asked me to change it.

Issue #1609, opened 2018-10-18
Book Building limit


SergejSoft:
We need to implement limit for people taking part in Bookbuilding. 
The max limit of an investor in book building 500.

This number must a value which we could change easily. 
pls. see if there is this type of validation on the front. 

The error to display in case number is reached. "This Campaign has reached the maximum number of Pre-sale investors. Please consider investing in the Public Sale instead."

desfero:
Agreed with @SergejSoft that we are simply going to hide investor widget if maxim count is reached

desfero:
@falinebambi we need some designs how it should be displayed to the user :)

PR #1644, opened 2018-10-22
#1609 Hide whitelisting after reaching limit

desfero:


rudolfix:
@falinebambi we need to show something when ETO is fully booked! this is I think pretty important indication that ETO is good

falinebambi:
image

Issue #1622, opened 2018-10-19
Add loading indicator for ETO OPPORTUNITIES


Moejoe90:
**Background** 
Currently, when the eto's are loading there is no indicator  that makes me wait simply from nothing there is a list 
![image](https://user-images.githubusercontent.com/23242822/47225719-120f3180-d3bf-11e8-9819-7b871cd8d06f.png)

**Task** 
Investigate if all states are there as in `loading` state and add a loading indicator

falinebambi:
https://codepen.io/PalmBeachagency/pen/GXJLom
![screen shot 2018-10-23 at 17 47 57](https://user-images.githubusercontent.com/17235402/47373291-cffe2c80-d6eb-11e8-80b2-1b09e1e50978.png)

PR #1651, opened 2018-10-23
#1622 Eto list loading indicator

desfero:
There is a lot of changes in importing of loading indicator as it's now nested inside own folder (loading-indicator).

Important files to check are:
EtoList.tsx
LoadingIndicatorHaxagon.tsx and styles


.sass-lint.yml
```diff
@@ -9,13 +9,7 @@ files:
     - 'coverage/**'
     - 'dist/**'
     - 'test/**'
-    - 'app/styles/react-slick-overrides.scss'
```
desfero: Moved to scsc files directly as we need to think in components. If i want to remove component then i should not do any changes in any kind of config files

trival: 👍

Issue #1484, opened 2018-10-02
Refactor Forms


desfero:
Background:
There are places in forms that need to be refactored so forms error messages/validation can be reused.

- [ ]  Create `FormRadioGroupField` that can be used to generate radio groups so we can stop repeating each time the same code (for e.g. `Eto Terms` form `enableTransferOnSuccess` can be used as a base implementation)
- [x]  Replace all error messages by `FormError` component (like I did in `FormField` here https://github.com/Neufund/platform-frontend/pull/1487/files#diff-898555fd97cfc464d7f31f650ee2c78cR113)
- [x]  `FormLabel` should use `label` tag for correct usability / semantic / accessibility
- [x]  Create own version of `MaskedInput` that will provide thorough `render` prop our `Input` component so any change to `Input` styles change `MaskedInput` style 
 - [ ] `FormRangeBase` add label and error message https://github.com/Neufund/platform-frontend/blob/99cc39929cdb0e6b8fc1090d931aaba2c068370c/app/components/shared/forms/form-field/FormRange.tsx#L96
 - [ ] Refactor `FormFieldDate` to only call `onChange` when all values where provided correctly (date, month, year) (see https://github.com/Neufund/platform-frontend/pull/3147#discussion_r296647689)

pstroinski:
To verify with the formik PR

Moejoe90:
This is done in theory with the Formik PR

PR #1654, opened 2018-10-24
#1519 Add proper input mask on investment form

### LINKED ISSUE #1519: undeletable zeros when providing ETH amount and other weirdness
1. type 250 in ETH then delete it with backspace. you see number below which cannot be deleted but can be navigated with arrows
image

2.  when you press any character which is not a digit the cursor goes right but the character under it stays

3. type 121.7652 and then start deleting with backspace. when you reach the decimal dot, the number morphs into 1210000.000 why?

do you convert ETH back to EUR then set EUR and then convert it back to ETH. it's hard to understand what is wrong here

trival:
fixes #1519


app/components/modals/tx-sender/investment-flow/Investment.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+  EInvestmentErrorState,
+  EInvestmentType,
+} from "../../../../modules/investment-flow/reducer";
+import { injectIntlHelpers } from "../../../../utils/injectIntlHelpers";
+import { InvestmentSelectionComponent } from "./Investment";
+import { wallets } from "./InvestmentTypeSelector.stories";
+
+const Investment = injectIntlHelpers(InvestmentSelectionComponent);
+
+storiesOf("Investment/Modal", module).add("default", () => (
```
desfero: @trival if it's modal maybe it's worth to wrap it in modal so regression test is === production view

app/components/modals/tx-sender/investment-flow/Investment.tsx
```diff
@@ -161,9 +164,20 @@ export const InvestmentSelectionComponent: React.SFC<IProps> = ({
             placeholder={`${intl.formatIntlMessage(
               "investment-flow.min-ticket-size",
             )} ${minTicketEur} €`}
-            controlCursor
             value={formatEur(euroValue)}
-            onChange={e => changeEuroValue(e.target.value)}
+            className="form-control"
+            renderInput={props => (
+              <MaskedInput
```
desfero: @trival can you please add new sub task for https://github.com/Neufund/platform-frontend/issues/1484.
```
Create own version of `MaskedInput` that will provide thorough `render` prop our `Input` component so any change to `Input` styles change `MaskedInput` style 
```

app/components/shared/forms/form-field/FormFieldRaw.tsx
```diff
@@ -66,12 +36,28 @@ export class FormFieldRaw extends React.Component<FieldGroupRawProps> {
       charactersLimit,
       errorMsg,
       onChange,
-      controlCursor,
+      renderInput,
```
desfero: can be changed to `renderInput: Input = Input` so now on line 71 we can call it as a component `<Input {...props} />`

app/modules/investment-flow/sagas.ts
```diff
@@ -36,11 +36,20 @@ import {
 
 function* processCurrencyValue(action: TAction): any {
   if (action.type !== "INVESTMENT_FLOW_SUBMIT_INVESTMENT_VALUE") return;
+  const state: IAppState = yield select();
```
desfero: @Moejoe90  @trival do we want to use global state or always pass some kind of selector?

app/components/modals/tx-sender/investment-flow/Investment.tsx
```diff
@@ -1,8 +1,10 @@
 import * as React from "react";
 import { FormattedMessage } from "react-intl-phraseapp";
+import MaskedInput from "react-text-mask";
 import { Col, Container, FormGroup, Label, Row } from "reactstrap";
 import { withHandlers, withProps } from "recompose";
 import { compose } from "redux";
+import createNumberMask from "text-mask-addons/dist/createNumberMask";
```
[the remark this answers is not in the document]
trival: ok

app/components/modals/tx-sender/investment-flow/Investment.stories.tsx
trival: ok, i prefer to rename it then

app/modules/investment-flow/sagas.ts
Moejoe90: Good that you mention this point. 
From my side, I want to push the convention of always using selectors if we are taking any info from the state. It is always easier to read and manage than accessing state directly even if it was simple extraction. 
The only worrying part of doing 
`yield select(someSelector);`
Is the loss of type safety , the pro's of this is you never run into bugs related to you not updating the state

desfero: On my side i'm a fan of 
```
const txData: ITxData = yield select<IAppState>(selectTxDetails);

```
we can omit IAppState and lose type-safety but it's ok if you agree to always pass the whole state to selectors.

About cons that you mention:
```
You don't know which part of the state this selector deals with but that is fine from my point
 ```
For me it's an advantage because I'm not interested where that data comes from at that point, I only need result (and by that, i don't need to go each time to the selector and check what slice of state it needs)

File: app/components/modals/tx-sender/investment-flow/Investment.tsx (hunk 2 above)
Comment (trival) [reply app/components/modals/tx-sender/investment-flow/Investment.tsx]: done

File: app/modules/investment-flow/sagas.ts (hunk 4 above)
Comment (trival) [reply app/modules/investment-flow/sagas.ts]:
```
const txData: ITxData = yield select<IAppState>(selectTxDetails);
```

I really don't see the benefit of this when you need more selections at some place. Just more code and more casting in my opinion

trival: also I don't really think this is a problem or style issue at all. I understand that using selectors is generally preferable to accessing raw state, but using selectors before or after yielding is just a matter of practicality i think, whatever supports concise and correct code better...

Moejoe90: @trival If we take that approach we will change our selectors and i already started doing that so no need for 
`yield select((state: IAppState) => selector(state.moduleState))` 
A simple 
`yield select(selector)` 
Works and that really looks concise and clear

PR #1656, opened 2018-10-24
Eto state conditions in investment

trival:
sets more conditions for wallet visibility and selectability during investment. 

Main logic update is in app/modules/investment-flow/sagas.ts, function getActiveInvestmentTypes.
The rest is mostly updating an Enum name.


trival:
https://github.com/Neufund/platform-frontend/pull/1654 needs to be merged first

trival:
also handles https://github.com/neufund/platform-frontend/issues/1401

app/modules/investment-flow/sagas.ts
```diff
@@ -178,6 +187,45 @@ function* setGasPrice(): any {
...
+  const etoId = selectInvestmentEtoId(state.investmentFlow);
+  const eto = selectEtoWithCompanyAndContractById(state, etoId);
+
+  let activeTypes: EInvestmentType[] = [
+    EInvestmentType.InvestmentWallet,
+    EInvestmentType.BankTransfer,
+  ];
+
+  // no bank transfer 3 days befor eto end
+  const etoEndDate = eto && eto.contract && eto.contract.startOfStates[EETOStateOnChain.Signing];
```
rudolfix: I know that it sounds like splitting hairs but:
1. of current `timedState` == `Whitelist` hide bank transfer 3 days before public phase
2. condition above should be used if `timedState` is public.

That will prevent also people on whitelist from sending money that will be received when we are in public already

app/modules/public-etos/selectors.ts
```diff
@@ -65,3 +65,21 @@ export const selectPublicEtos = (state: IAppState): TEtoWithCompanyAndContract[]
...
+  return contracts && contracts.timedState;
+};
+
+export const selectEtoOnChainStateById = (
+  state: IPublicEtoState,
+  etoId: string,
+): EETOStateOnChain | undefined => {
+  const code = selectEtoPreviewCode(state, etoId);
+  if (code) {
+    return state.contracts[code] && state.contracts[code].timedState;
```
rudolfix: are you sure that internal state is by `code`, not `etoId`? I think using `code` is better btw. so it's ok, just making sure

trival: yes, contracts are stored by previewCode, so all should be fine

app/modules/investment-flow/sagas.ts
trival: 👍

PR #1657, opened 2018-10-24
Improve messages

drakkein:


app/components/dashboard/notification-widget/Notification.tsx
```diff
@@ -28,8 +29,13 @@ export const Notification: React.SFC<INotificationProps> = ({
   text,
   onClick,
   actionLinkText,
+  clickable,
 }) => (
-  <div data-test-id="notification" className={cn(styles.notification, type)}>
+  <div
+    data-test-id="notification"
+    className={cn(styles.notification, type, clickable ? styles.clickable : "")}
+    onClick={clickable ? onClick : () => null}
```
desfero: It's not correct to attach onClick handler to 'div' as then it cannot be reached by 'tab'. Please use our Button component when onClick is provided

app/components/dashboard/notification-widget/Notification.module.scss
```diff
@@ -9,6 +9,10 @@
   padding-bottom: $authorized-layout-vertical-padding-sm;
 }
 
+.clickable {
```
desfero: this is not needed when we migrate to Buton

app/components/dashboard/notification-widget/Notification.tsx
desfero: If' it's possible please update also onClick on line 46 to use our Button witn INLINE styles

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
```diff
@@ -115,14 +119,24 @@ const EtoStatusManager = ({
...
+          <InvestWidget
+            raisedTokens={eto.contract!.totalInvestment.totalTokensInt.toNumber()}
+            investorsBacked={eto.contract!.totalInvestment.totalInvestors.toNumber()}
+            tokensGoal={(eto.newSharesToIssue || 1) * (eto.equityTokensPerShare || 1)}
+            etoId={eto.etoId}
+          />
+        );
+      } else {
+        return (
+          <Link to={appRoutes.settings}>
```
desfero: ouuu :( it's an antipattern and invalid semantically code to have <a><button></button></a>, please use 'ButtonLink' component

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
```diff
@@ -115,14 +119,24 @@ const EtoStatusManager = ({
...
     case ETOStateOnChain.Public: {
-      return (
-        <InvestWidget
-          raisedTokens={eto.contract!.totalInvestment.totalTokensInt.toNumber()}
-          investorsBacked={eto.contract!.totalInvestment.totalInvestors.toNumber()}
-          tokensGoal={(eto.newSharesToIssue || 1) * (eto.equityTokensPerShare || 1)}
-          etoId={eto.etoId}
-        />
-      );
+      if (!settingsUpdateRequired) {
```
desfero: It's better to have true condition first as it's more readable

app/modules/notifications/reducer.ts
```diff
@@ -64,9 +65,10 @@ export const notificationsReducer: AppReducer<INotificationsState> = (
 export const settingsNotification = () => ({
   id: Date.now(),
   type: NotificationType.WARNING,
-  text: "Please update your account before proceeding. Head to Settings now. ",
+  text: "You will not be able to INVEST until you complete required steps in Settings.",
```
desfero: We need to split notification right now for two different user types as this one is a generic one and we don't want to show that new text for issuer. For issue it still should stay same as previous one

app/modules/notifications/reducer.ts
```diff
@@ -64,9 +65,10 @@ export const notificationsReducer: AppReducer<INotificationsState> = (
...
   actionLinkText: "Go to settings",
```
desfero: We can hide actionLinkText when whole notification is clickable as it's a not good to have anchor inside button

app/components/dashboard/notification-widget/Notification.spec.tsx
```diff
@@ -25,17 +25,20 @@ describe("<Notification />", () => {
...
 
     const componentWarning = shallow(<Notification {...props} type={NotificationType.WARNING} />);
 
-    expect(componentWarning.hasClass(NotificationType.WARNING)).to.be.true;
+    expect(componentWarning.render().hasClass(NotificationType.WARNING)).to.be.true;
   });
 
   it("should call correct click handlers for close button", () => {
-    const component = shallow(<Notification {...props} />);
+    const component = mount(<Notification {...props} />);
```
desfero: We should not use shallow here as it's going to break encapsulation, maybe it's possible to somehow trigger onclick on the 'Button' element of mock that 'Button' in some way.

Issue #1459, opened 2018-10-01
INVEST button should be only visible for users with valid on-chain KYC claim


rudolfix:
This ticket has two parts (and you can split it):

- [x]  PART 1
1. add new singleton contract to `ContractsService` instantiated at startup. The artifacts is `IdentityRegistry` and you can find it here
2. add a method to obtain identity claim for currently logged on user and deserialize it
https://github.com/Neufund/platform-contracts/blob/master/contracts/Identity/IdentityRegistry.sol#L36

and deserialize it into following record
https://github.com/Neufund/platform-contracts/blob/master/contracts/Identity/IIdentityRegistry.sol#L15
we have deserialization procedure in our test suite:
https://github.com/Neufund/platform-contracts/blob/master/test/helpers/identityClaims.js

- [x]  PART 2
Show INVEST button only if the claim `isVerified` IS set and claim `isAccountFrozen` IS NOT set

when to load claims? we should probably load ethereum claims together with checking backend claims (even in the same saga....)
it's also connected to this ticket #1422 #1423 #1424

glupiawymowka:
looks good
![image](https://user-images.githubusercontent.com/18315589/48912303-87b26580-ee75-11e8-9323-353544c26bf8.png)

PR #1662, opened 2018-10-25
#1459 Hide Invest button for not verified user - Phase 1

desfero:


app/components/dashboard/eto-list/EtoList.tsx
```diff
@@ -27,7 +27,7 @@ const EtoListComponent: React.SFC<IStateProps> = ({ etos }) => (
     <Col xs={12}>
       {etos ? (
         etos.map(eto => (
-          <div className="mb-3">
+          <div className="mb-3" key={eto.previewCode}>
```
Moejoe90: :thinking: 
i saw this before

app/modules/kyc/sagas.ts
```diff
@@ -92,6 +96,23 @@ function* kycRefreshWidgetSagaWatcherStop(): any {
   }
 }
 
+/**
+ * Individual Request
+ */
+function* loadIdentityClaim({ contractsService, logger }: TGlobalDependencies): Iterator<any> {
+  const identityRegistry: IdentityRegistry | undefined = contractsService.identityRegistry;
```
rudolfix: you can assume that if `ContractsService` is instantiated, `identityRegistry` is always there

Moejoe90: IMO we need to merge https://github.com/Neufund/platform-frontend/issues/1685 before that

Issue #1610, opened 2018-10-18
Public ETO view - wallet address selector error


drakkein:
While user is logged out on public ETO view page, wallet address selector is throwing an error.

```
Error while loading wallet data:  TypeError: Cannot read property 'address' of undefined
    at selectEthereumAddress (selectors.ts:13)
    at eval (index.js:86)
    at Object.eval [as selectEthereumAddressWithChecksum] (index.js:36)
    at effects_1.select (sagas.ts:15)
    at runSelectEffect (proc.js:708)
    at runEffect (proc.js:445)
    at next (proc.js:325)
    at proc (proc.js:280)
    at runForkEffect (proc.js:564)
    at runEffect (proc.js:445)
    at next (proc.js:325)
    at currCb (proc.js:398)
    at takeCb (proc.js:476)
    at Object.put (channel.js:85)
    at eval (channel.js:173)
    at eval (channel.js:198)```

PR #1666, opened 2018-10-26
#1610 Fix public view console error when unauthorized

desfero:


package.json
```diff
@@ -10,25 +10,25 @@
   ],
   "scripts": {
     "start": "(cp -n ./app/config/.env.example .env || true) && NODE_ENV=development webpack-dev-server --config ./webpack/webpack.config.dev.js",
-    "start:remote": "webpack-dev-server --config ./webpack/webpack.config.dev-remote.js",
```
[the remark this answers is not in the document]
Moejoe90: Ok how about you add `start:widnows` as settings as well this way everyone benefits

app/components/eto/EtoPublicView.tsx
```diff
@@ -49,7 +49,10 @@ export const EtoPublicView = compose<IProps, IRouterParams>(
   onEnterAction({
     actionCreator: (dispatch, props) => {
       dispatch(actions.publicEtos.loadEto(props.etoId));
-      dispatch(actions.wallet.loadWalletData());
+
+      if (props.userType) {
+        dispatch(actions.wallet.loadWalletData());
```
Moejoe90: Now that i think about it does EtoPublicView need the wallet information for any reason

desfero: seem you are right, wallet is not needed there

PR #1671, opened 2018-10-26
E2E tests for ETO required fields

desfero:


typings/cypress.d.ts
```diff
@@ -0,0 +1,7 @@
+declare namespace Cypress {
```
desfero: @Moejoe90 is there any reason that we prefer to patch library in really insane way than simply monkey patch typings?

cypress/support/commands.js
```diff
@@ -29,3 +29,75 @@ Cypress.Commands.add("awaitedClick", { prevSubject: "element" }, (subject, waitD
     .wait(waitDuration)
     .click();
 });
+
+const resolveMediaType = (headerContents) => {
```
desfero: I don't really know what this code is doing as it was copied from some cypress issue :D

app/test-e2e/utils/forms.ts
```diff
@@ -44,7 +51,10 @@ const isSubmitField = (field: TFormFieldFixture) =>
  * @param fixture - Which form fixture to load
  * @param submit - wether to submit the form or not, default true
  */
-export const fillForm = (fixture: TFormFixture, { submit = true }: { submit?: boolean } = {}) => {
+export const fillForm = (
```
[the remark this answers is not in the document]
desfero: yeap, can do it in a future, same for logging, it can be done as cy.login(...)

Issue #1570, opened 2018-10-16
Remove No. and add placeholder "amount"


SergejSoft:
![Screen Shot 2018-10-16 at 12.40.16.png](https://images.zenhubusercontent.com/57dd639be40e5714b16f0753/1726b860-07d4-4da0-aa7c-84f29b6731a7)

/Users/soft/platform-frontend/app/components/shared/forms/formField/FormCategoryDistribution.tsx
line:139

glupiawymowka:
"No" is gone correctly, but "Amount" cuts strangely when the window gets smaller

![image](https://user-images.githubusercontent.com/18315589/48002374-ae4d6e00-e10b-11e8-9ad2-d51b72f05037.png)

SergejSoft:
Please break the input fields in two lines in the smaller screen sizes. XS, S.

glupiawymowka:
done

Issue #1520, opened 2018-10-09
[ISSUER] FormTransformingField works incorrectly


rudolfix:
when you enter a value in the form it is transformed into value and then transformed back and displayed again.

try to type 7 when transform ratio is 0.01 (for example whitelist discount on Investment Terms)

you'll get some crazy value due to rounding errors

![image](https://user-images.githubusercontent.com/17202864/46702870-1b610700-cc25-11e8-8cd1-455e8118ff28.png)

what is stored must be decoupled from what is displayed

rudolfix:
sorry for giving it to you after the YupTS but here we have serious problems with forms - seems like implementation went in wrong direction and now we have to replace this component

glupiawymowka:
Entered 7 there, still getting the same result 

![image](https://user-images.githubusercontent.com/18315589/48002926-439d3200-e10d-11e8-905f-a09fae5f5650.png)

PR #1672, opened 2018-10-27
At/#1520/#1570 cross validation and placeholders in value fields

AlexeyTuboltsev:
#1520: add placeholder text to value fields in forms

#1570: add cross validation to compound (key-value) fields in forms. 
uses formik's validate method on Field to add custom cross validation on change. Custom validation is derived from schema provided to form, I only add cross check of the neighbouring field and one error. 
After that I need to clean up the data on save to exclude {undefined, undefined} fields from array or set the whole array to undefined if it's empty.
Initial values must always be {undefined,undefined} (not {"",0} or the like), otherwise we get these (invalid) values saved to backend.


app/components/shared/forms/form-field/FormInput.tsx
```diff
@@ -90,6 +110,22 @@ export class FormInput extends React.Component<FormInputProps> {
...
+                            ),
+                          );
+                        }}
+                        onBlur={() => {
+                          setFieldTouched(name);
+                          validateForm();
+                          if (setAllFieldsTouched !== undefined && field.value !== undefined) {
+                            setAllFieldsTouched(true);
+                          }
+                        }}
```
trival: So now you we only validate on blur, in all input fields of the application?

app/components/shared/forms/form-field/FormInput.tsx
```diff
@@ -31,6 +31,20 @@ export interface IFormInputExternalProps {
 
 export type FormInputProps = IFormInputExternalProps & FieldAttributes<any> & CommonHtmlProps;
 
+const transform = (value: string, charactersLimit?: number) => {
+  return value === undefined ? "" : computedValue(value, charactersLimit);
+};
+
+const transformBack = (value: number | string) => {
+  if (value && typeof value === "number") {
```
trival: this removes 0 as number input! the check `value &&` can be ommited

app/components/shared/forms/form-field/FormInput.tsx
```diff
@@ -53,11 +67,14 @@ export class FormInput extends React.Component<FormInputProps> {
       min,
       max,
       size,
+      compoundFieldValidation,
```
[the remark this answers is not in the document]
trival: please add propper typing for customValidation function

app/components/shared/forms/form-field/FormTransformingField.tsx
```diff
@@ -22,10 +23,20 @@ interface IFieldGroup {
 type FieldGroupProps = IFieldGroup & FieldAttributes<any> & CommonHtmlProps;
```
trival: change `FieldAttributes<any>` to `FieldAttributes<{}>` to get proper typing

app/components/shared/forms/form-field/FormTransformingField.tsx
```diff
@@ -68,8 +87,16 @@ export class FormTransformingField extends React.Component<FieldGroupProps> {
...
+                        setFieldTouched(name);
                         setFieldValue(name, transformBack(e.target.valueAsNumber, ratio));
                       }}
+                      onBlur={() => {
+                        setFieldTouched(name);
+                        validateForm();
+                        if (setAllFieldsTouched !== undefined && field.value !== undefined) {
+                          setAllFieldsTouched(true);
+                        }
+                      }}
```
trival: same comments as in FormField component, general customOnBlur etc...

Issue #1199, opened 2018-08-10
Breaking change in Metamask (will stop working) on November 2nd


banciur:
`On November 2nd, MetaMask and other dapp browsers will stop automatically injecting web3. Instead, dapps should request access using a new postMessage API.`

more info - https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8

rudolfix:
here's how to test it now. "Preparing your dapp" section in the link above

# this is important to be solved now

glupiawymowka:
confirmed access for our platform, logged it, did a transaction, YAY

![image](https://user-images.githubusercontent.com/18315589/48135420-30b96780-e29d-11e8-9d01-c37259b82cd8.png)

PR #1677, opened 2018-10-28
#1199 Change connect flow for browser wallet to new EIP

banciur:
It will introduce new step in process of unlocking wallet. Where you have to approve that website will get access to your account number.

first_step

If you reject it wallet selector will show following screen.

rejection

In case of signing and sending transaction saga `accessWalletAndRunEffect` I didn't implement screen above we just nag user to confirm data approval or press cancel

To do it correctly we would need to change more components. Generally problem is in fact that until now connect flows are done as loops where we issue promise that will be resolved or rejected almost instantly. Here promise is pending till user perform action which might take a while that in my opinion requires some more changes. But as I heard that whole wallet chooser and signer components modules will be rewritten so in my opinion what I proposed here is good enough for now.

@glupiawymowka To test it you would have to install special build of Metamask. Instruction and links are there: https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8#1f3f

Then when you run this special metamask you have to turn `Privacy Mode` in settings


app/lib/web3/BrowserWallet.ts
```diff
@@ -84,21 +86,47 @@ export class BrowserWallet implements IPersonalWallet {
 
 @injectable()
 export class BrowserWalletConnector {
+  data_approval_pending: boolean;
```
desfero: can you please use camelCase for naming? Same in WalletBrowser.tsx

app/lib/web3/BrowserWallet.ts
```diff
@@ -84,21 +86,47 @@ export class BrowserWallet implements IPersonalWallet {
...
+          this.data_approval_pending = false;
+        }
+      }
+    }
+
+    /*
+    Problem we have here is that there are two behaviours of promise returned from metamask injectedWeb3Provider.enable()
+    New: when promise is resolved it means that you now can obtain accounts so we can continue
+    In old one enable always resolves but metamask can actually still be locked so getAccountAddress will return nothing.
+    That's why there are those two checks. After 2 Nov 2018 we can remove second check.
```
desfero: Can you add a TODO comment what we should remove after 2 Nov 2018 + github issue for that

Issue #1659, opened 2018-10-25
[KYC] country of residence wording


glupiawymowka:
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/456f3024-6600-4155-bd98-6d4eb4aba4d8)

Since we don’t ban investors because of where they’re coming from, rather than residence country, this should be clearer.
Please change to  “Unfortunately we do not accept investors or companies with residency in your country due to regulatory restrictions”.

meneramandarin:
i cannot find an entrance for this in the locals file. anywhere else i can change it @desfero?

desfero:
@meneramandarin they were not placed in locales file. Btw i moved then to locales file and also fixed that working change

meneramandarin:
okay, thanks.

PR #1681, opened 2018-10-28
#1659 Country of residence wording

desfero:


app/lib/api/util/schemaHelpers.tsx
```diff
@@ -53,10 +60,8 @@ export const citizen = Yup.bool();
 //TODO: add to translations
```
trival: lets remove the TODO : )

desfero: ouuuu, fixed

PR #1684, opened 2018-10-29
At/#1464 issuer dashboard default null values

AlexeyTuboltsev:


app/components/eto/utils.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { expect } from "chai";
+
+import { applyDefaults } from "./utils";
+
+describe("applyDefaults", () => {
+  it("sets defaults if value is missing/undefined", () => {
+    const data = { key1: "value", key2: 222 };
+    const defaults = { key2: "defaultValue 2", key3: "defaultValue 3" };
+    const output = {
```
trival: since the only difference between your function and `{...defaults, ...values}` is the handling of `null`, maybe the test should show that with an extra test case

.circleci/config.yml
```diff
@@ -11,8 +11,8 @@ jobs:
       # Download and cache dependencies
       - restore_cache:
           keys:
-          - v12-deps-{{ checksum "package.json" }}
-          - v12-deps-
+          - v13-deps-{{ checksum "package.json" }}
+          - v13-deps-
       - run: mkdir -p /home/circleci/.cache/Cypress && cp -R /home/circleci/.cache/Cypress ~/.cache/Cypress
```
trival: revert, if this is unnecessary

PR #1686, opened 2018-10-29
At/#1423 reload settings and poll for KYC status

AlexeyTuboltsev:
Polling for KYC status change, for both individual and business.

I couldn't reproduce the situation that the settings don't reload sol I'll leave it like this for now


app/components/settings/kyc-states/KycStatusWidget.tsx
```diff
@@ -54,6 +54,7 @@ export type IKycStatusWidgetProps = IStateProps & IDispatchProps & IOwnProps;
 const statusTextMap: UnionDictionary<TRequestStatus, React.ReactNode> = {
   Accepted: <FormattedMessage id="settings.kyc-status-widget.status.accepted" />,
   Rejected: <FormattedMessage id="settings.kyc-status-widget.status.rejected" />,
+  Ignored: <FormattedMessage id="settings.kyc-status-widget.status.rejected" />,
```
trival: is the translation id also rejected for "Ignored" state?

app/components/settings/Settings.tsx
```diff
@@ -111,6 +111,7 @@ export const Settings = compose<React.SFC>(
   }),
   onEnterAction({
     actionCreator: dispatch => {
+      // dispatch(actions.kyc.kycLoadBusinessData())
```
Moejoe90: There is a comment here

app/lib/api/KycApi.interfaces.ts
```diff
@@ -108,7 +108,13 @@ export const KycFileInfoShape = Yup.object().shape({
 });
 
 // request state
-export type TRequestStatus = "Draft" | "Pending" | "Outsourced" | "Rejected" | "Accepted";
```
Moejoe90: I would change this to an enum as per our guidelines but this is not blocking

app/modules/kyc/sagas.ts
```diff
@@ -54,12 +57,18 @@ function* kycRefreshWidgetSaga(): any {
...
+        outsourcedStatus &&
+        (outsourcedStatus === "started" ||
+          outsourcedStatus === "canceled" ||
+          outsourcedStatus === "aborted" ||
+          outsourcedStatus === "review_pending"))
     ) {
-      yield put(actions.kyc.kycLoadIndividualRequest(true));
+      requestType === "individual"
+        ? yield put(actions.kyc.kycLoadIndividualRequest(true))
+        : yield put(actions.kyc.kycLoadBusinessRequest());
```
Moejoe90: I'm not sure why does load individual request have true as input i would investigate what this means and if load business request requires true or not
Maybe you did this already

Issue #1580, opened 2018-10-16
use preview code instead of eto id in public profile and all routes where eto-id is used


rudolfix:
basically we can merge eto-preview and eto-view into one route (we can keep both as aliases as you wish) but always use eto-preview as id

reason: eto id changes and people that bookmarked eto or saved it etc will loose it when it goes on chain. like it happened today
https://platform.neufund.net/eto-view/0xC557A62fB0D9cAE419C5Bc10001dcF9031770B79
stopped working and switched to
https://platform.neufund.net/eto-view/0x164e07Ae48cA7774663e90732d44b324f2e3C679

PR #1687, opened 2018-10-29
#1580 Use preview code instead of eto id in public profile

desfero:


app/components/eto/EtoPublicView.tsx
```diff
@@ -1,10 +1,11 @@
 import * as React from "react";
-import { branch, compose } from "recompose";
+import { compose } from "redux";
 
+import { branch, renderComponent } from "recompose";
```
drakkein: please move this to first group of imports

Issue #1558, opened 2018-10-15
[ETO OPPORTUNITIES + PUBLIC VIEW] "Neufund" widget should be a link, low prio


glupiawymowka:
The "Neufund" widget should be linked to https://neufund.org

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/e7a8c252-43c4-4e5c-8b86-15c309461e69)

PR #1696, opened 2018-10-30
#1558 powered by neufund in eto dashboard must be a link

AlexeyTuboltsev:


app/components/shared/forms/form-field/utils.ts
```diff
@@ -41,3 +42,11 @@ export const countedCharacters = (val: InputProps["value"] = "", limit: number)
 
   return `${val.length}/${limit}`;
 };
+
+export const isFieldRequired = (validationSchema: any, name: string) => {
+  if (validationSchema) {
```
drakkein: can we return false by default if schema is not provided?

app/components/shared/forms/form-field/utils.ts
```diff
@@ -1,6 +1,7 @@
 import { FormikErrors, FormikTouched } from "formik";
-import { get } from "lodash";
+import { get, isFunction } from "lodash";
 import { InputProps } from "reactstrap";
+import { getFieldSchema, isRequired } from "../../../../utils/yupUtils";
```
drakkein: let's stick to convention where 3rd party imports are in first group of imports, and all the others grouped below.

app/components/shared/forms/form-field/utils.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { expect } from "chai";
 
-import { isNonValid, isValid } from "./utils";
+import * as Yup from "yup";
```
drakkein: If we are grouping imports yup should be moved to first group

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.module.scss
```diff
@@ -155,6 +155,12 @@
   .neufund {
     font-size: 16px;
     margin-top: -2px;
+    text-decoration: none;
+
+    &:hover,
+    &:active {
```
drakkein: Maybe we should change color on focus also? @desfero

app/components/shared/forms/form-field/utils.ts
AlexeyTuboltsev: yes, you're right.
But I'll do it in another branch I'm working on, it landed in this one by mistake

PR #1704, opened 2018-10-31
At/#1598 register with lite wallet button inactive

AlexeyTuboltsev:
1. reset isSubmitting on props change.
2. add translations


app/test-e2e/settings/LightWalletRegister.spec.e2e.ts
```diff
@@ -66,7 +67,7 @@ describe("Light wallet login / register", () => {
     });
   });
 
-  it("should return an error when logging with same email", () => {
+  it.only("should return an error when logging with same email", () => {
```
Moejoe90: We need to take off only :D

Issue #1660, opened 2018-10-25
[bookbuilding] tool tip needed


glupiawymowka:
1. "Show my email" is unclear, should be changed to "REVEAL MY EMAIL TO COMPANY" 
2. A tooltip with the following explanation should be added: "When ticking this box your email address will be shared with the company doing the ETO." 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/660f8c36-5302-454f-87ff-92d3b45f5520)

glupiawymowka:
here it is 

![image](https://user-images.githubusercontent.com/18315589/48012588-a3ea9e80-e122-11e8-8100-94e0e517ba33.png)

PR #1705, opened 2018-10-31
#1660 Bookbuilding tooltip

desfero:
Final change:
image


desfero:
@trival I also wonder about why we have own tooltip. As our implementation is quite naive (for e.g. tooltip is not going to show on the bottom when there is not enough space on top). So let's keep that in mind and change Tooltip.tsx to use reacstrap one

PR #1706, opened 2018-10-31
Improve shared folder by fixing typo mistakes and new stories for regression testing

desfero:


app/utils/connectedRouting.tsx
```diff
@@ -1,7 +1,7 @@
 import { SwitchProps } from "react-router";
```
Moejoe90: :+1:

Issue #1528, opened 2018-10-10
Portfolio overhaul and fixes


rudolfix:
- [x] hide NEU panel
![image](https://user-images.githubusercontent.com/17202864/46762105-8a4a6880-ccd6-11e8-8cbe-5f34cd3ef2a1.png)
Please hide this for now. We’ll not make it for first ETO. Instead we’ll add NEU to YOUR ASSETS!
- [x] simplify DIVIDENDS FROM NEU
![image](https://user-images.githubusercontent.com/17202864/46762154-ac43eb00-ccd6-11e8-8cbc-c84921346e4f.png)
Please rename it to OUTSDANDING PAYOUTS (this will be changed by lawyers anyway ;>). Here you will see not only payouts coming from NEU, but also from other tokens
Also table is wrong and we need to work on it after fist ETO.
You can just use “MY PROCEEDS” from Manage Wallet here. And change the title to the above. Will be good for now
- [ ] List NEU in YOUR ASSETS
![image](https://user-images.githubusercontent.com/17202864/46762224-d5647b80-ccd6-11e8-8ce0-8eeb7b18c140.png)
Fill columns as follows
•	BALANCE: on Neumark.balanceOf(user id/wallet address). If balance == 0 then row is not displayed
•	PRICE – it’s available in price feed, we already have it in state
•	VALUE EUR – multiply balance * price
•	NEU REWARD – empty
•	DOCUMENTS – (temporarily): link to page (open in new tab): https://commit.neufund.org/commit/status?address=<user_id> (example of working address: 0x430513b91748d977Aa38e4db691764a97Fe28236)

NEU is a token as any other so later we'll fully unify it with listing. Now you need probably need to hack this component. Up to you how you add this "outstanding data"

glupiawymowka:
checked

1. case 1 - no assets (.net)
![image](https://user-images.githubusercontent.com/18315589/48556210-9cc24e00-e8e3-11e8-9a5a-71001b429981.png)

2. case 2 - existing assets (.io)
![image](https://user-images.githubusercontent.com/18315589/48556397-1eb27700-e8e4-11e8-91aa-e5f6fafe2c96.png)

glupiawymowka:
@SergejSoft or @rudolfix  please let me know what do you think. Currently, there is an information "There is nothing to show" for reserved assets, but no such information for "Your assets"
Should it be changed? 

[...]

PR #1715, opened 2018-10-31
#1528 Portfolio overhaul and fixes

drakkein:


drakkein:
@rudolfix @desfero I see that row is disappearing only when there is an error during wallet request. Can we have short discussion on Monday how we handle this? I guess we should cache wallet data to have access to last correct state in case of error.

app/components/shared/NewTable.tsx
```diff
@@ -20,8 +20,13 @@ interface IPlaceholderTableRow {
...
+  | React.ReactElement<INewTableRow>
+  | React.ReactElement<INewTableRow>[]
+  | boolean;
+
 interface INewTable {
-  children: React.ReactElement<INewTableRow> | React.ReactElement<INewTableRow>[];
+  children: INewTableChildren | INewTableChildren[];
   placeholder?: TTranslatedString;
   className?: string;
   keepRhythm?: boolean;
```
AlexeyTuboltsev: can you please change the uniqueId() on the line 45? (it's not your code but we can't leave it like this ))

app/components/shared/NewTable.tsx
```diff
@@ -20,8 +20,13 @@ interface IPlaceholderTableRow {
   numberOfCells: number;
 }
 
+type INewTableChildren =
+  | React.ReactElement<INewTableRow>
+  | React.ReactElement<INewTableRow>[]
+  | boolean;
```
AlexeyTuboltsev: I suggest to change this to null. Element or null is idiomatic react way to deal with conditional rendering

app/components/portfolio/PortfolioLayout.tsx
```diff
@@ -201,6 +186,34 @@ const PortfolioLayout: React.SFC<TPortfolioLayoutProps> = ({ myAssets, pendingAs
...
+              />
+              <>{"-"}</>
+              <Link
+                to={withParams(externalRoutes.commitmentStatus, { walletAddress })}
+                target={"_blank"}
+              >
+                <FormattedMessage id="portfolio.section.your-assets.tokenholder-agreement" />
+              </Link>
+            </NewTableRow>
+          )}
```
AlexeyTuboltsev: if you change the conditional to {condition ? <NewTableRow> : null} 
you won't set newTableRow type to boolean, which is a bit strange and counterintuitive

app/components/portfolio/PortfolioLayout.tsx
```diff
@@ -39,72 +46,50 @@ const getNeuReward = (equityTokenInt: BigNumber, equivEurUlps: BigNumber): strin
...
+              size="large"
+              moneyValue="100000000"
+              moneyChange={-20}
+              tokenValue="1000000"
+              tokenChange={20}
+            />
+          </Col>
+        </Row>
+      </>
+    )}
```
AlexeyTuboltsev: please add some top margin to the next element, if assets are not visible.

rudolfix:
> @rudolfix @desfero I see that row is disappearing only when there is an error during wallet request. Can we have short discussion on Monday how we handle this? I guess we should cache wallet data to have access to last correct state in case of error.

@drakkein sure, I had a busy day today but tomorrow let's squeeze some time. around 2pm. probably. thanks!

Issue #1727, opened 2018-11-04
Add list parameter to newsletter subscribe field in platform.neufund


banciur:

banciur:
@glupiawymowka It should work in same way as before.

PR #1729, opened 2018-11-05
#1727 Update call to newsletter endpoint

banciur:


banciur:
@glupiawymowka It should work in same way as before.

Issue #1517, opened 2018-10-09
[BUG - IMPORTANT] invest entire balance - gas cost incorrectly added ^


rudolfix:
Please follow the same rules as with the investment calculation

1. when investing from any of ICBM Locked Accounts - you SHOULD spend entire wallet balance up to a wei precision: all the gas costs will come from ether balance

2. when investing from ether/ether token - only in that case you should SUBSTRACT gas cost from the total eth balance and the remainder should be spent.

mind wei precision as per #1515 

also when there is 0 balance on ether/ether token I see something like below which looks like you actually ADD the gas cost not SUBSTRACT?

![image](https://user-images.githubusercontent.com/17202864/46695250-5b1cf400-cc0f-11e8-9471-89b281534479.png)

rudolfix:
@rudolfix bug is still there - entire investment balance of Investmwnt Wallet is set and then gas cost is added on top - and your balance is exceeded

also button INVEST stays grey and no error message is displayed until you actually change the amount

glupiawymowka:
done

Issue #1515, opened 2018-10-09
investment flow - when converting from one currency to another - mind the roundings


rudolfix:
per ticket #1514 we should always convert currencies with wei precision. please make sure you call all smart contracts methods with wei precision. examples
1. someone enters ETH amount and you call `calculateContribution`
2. someone enters EUR amount and you call `transfer` on an Ether token

however we should not display wei precision: no one will understand values with 18 digits after dot.

1. display EUR with two digits precision
2. display ETH with four digits precision

**always round up to the required precision to be on a safe side**

this is particularly important for direct bank transfer where amount will be copied to clipboard and we need to make sure that required number of tokens will be bough

trival:
done, all the rounding is done as suggested.

Issue #1514, opened 2018-10-09
cannot invest minimum ticket from Ether wallet - are we losing precision?


rudolfix:
steps to reproduce
1. Select ETH Investment wallet
2. Type 100 EUR (which is minimum ticket) in Euro Amount
3. Confirm and sign transaction
4. Transaction will revert

IMO transaction reverts because what happens is that 100 EUR is converted into ETH via price oracle and this amount is sent to smart contract which calculates it back to euro. I suppose when you do your calculation you round value somehow or you round it differently than smart contract. Please check the following
1. we have price from price oracle in full precision (up until wei) in state. it's not rounded at any point
2. you do all conversions on BigNumbers. js numbers are never used
3. you do not round to 4 decimal places etc. you always use wei precision and round only to display
4. when converting EUR to ETH you probably use inverse price somewhere. correct? do not calculate inverse yourself, take it from smart contract (at the same time we take other rates - you can ask for multiple rates in the same call what I hope we do). just ask EUR to ETH price (we already ask ETH to EUR)
5. in all price conversions we'll use div() please round results of div as follow number.div(number).round(0, 4) so you round as smart contract rounds

trival:
@rudolfix I have implemented all your suggestions. Still the problem is not with rounding.
In Ether investment, the fixed value is always the ether value, and Euro is always only estimation, calculated from ether price. When token price oracle changes, the eth value stays fixed, the euro value is recalculated. This is because it is Ether investment! But of course, if price goes up, the euro goes below the minimum automatically. 

The problem is, I have to keep one value that is the source of truth, and one that is calculated. Which one should I choose?? Eur or Eth? The same problem is with the invest full Wallet edge case. I think we cannot handle these border conditions 100% correctly with all these conversions going on...

glupiawymowka:
min ticket is working now

PR #1739, opened 2018-11-06
[WIP] Investment currency fixes

trival:
Related: 
https://github.com/Neufund/platform-frontend/issues/1514
https://github.com/Neufund/platform-frontend/issues/1515
https://github.com/Neufund/platform-frontend/issues/1517


app/components/modals/tx-sender/investment-flow/Investment.tsx
```diff
@@ -258,8 +259,7 @@ export const InvestmentSelectionComponent: React.SFC<IProps> = ({
           <div>
             <FormattedMessage id="investment-flow.total" />:{" "}
             <span className={styles.orange}>
-              {formatThousands(formatEur(totalCostEur))} € ≈ ETH{" "}
-              {formatThousands(formatEth(totalCostEth))}
+              {formatEurTsd(totalCostEur)} € ≈ ETH {formatEthTsd(totalCostEth)}
```
desfero: I'm just wondering why we duplicate logic around money displaying. We can use here `Money` component 
```
<Money value={totalCostEur} currency="eur" />{" = "}<Money value={totalCostEth} currency="eth" />
```
i know that the ui will look differently 
```
<amount> EUR = <amount> ETH
```
but then it's consisten with whole project (where currency code is after amount and currency symbol is before amount)
Also i will decrease code duplication as current `formatEthTsd` and `formatEurTsd` are same what `Money` component is doing

trival: I will stay with the current thing for the sake of design. Money element uses the same formatting function under the hood, so we are still at least half consistent. But the main reason was that investment originally had no decimal fraction on the display, so no cents. It now changed to the standard representation, so maybe the whole formating could be revisited...

Issue #1743, opened 2018-11-06
Icon Identity on Blockchain


SergejSoft:
1. Add an active icon in the INVESTOR Dashboard that is visible after KYC is finished.  (see mockup)

2. Once clicked or Hoovered over we show the modal with Information that "User Identity is verified on Blockchain"

@mpneu to provide zeplin file and source files for the icon and Desing for modal.

I suggest using this modal with different text. https://zpl.io/awmeY41

mpneu:
Zeplin: https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/dashboard?seid=5be1ab8df21721617a885cbd

rudolfix:
@mpneu could you please put correct text in the zeplin? this is IPFS text there

PR #1755, opened 2018-11-07
#1743 Identity on Blockchain

desfero:


app/components/layouts/LayoutAuthorizedMenu.tsx
```diff
@@ -169,10 +263,21 @@ export const LayoutAuthorizedMenuComponent: React.SFC<IStateProps> = ({ userType
...
+      isIdentityModalOpened: selectGenericModalIsOpen(state.genericModal),
+      isVerifiedInvestor: selectIsVerifiedInvestor(state),
+    }),
+    dispatchToProps: dispatch => ({
+      openIdentityModal: () => dispatch(actions.genericModal.showModal(IdentityModal)),
+    }),
+  }),
+  withHandlers<IStateProps, IWithProps>({
+    isLinkActive: props => match => match && !props.isIdentityModalOpened,
+    isIdentityButtonActive: props => () => props.isIdentityModalOpened,
```
trival: why is this a function and not a simple boolean prop? I guess for parity with isLinkActive, which is required from react-router-dom, so it is somehow fine, but maybe also unneccesary...

desfero: yeap, agree with you, it's simpler to understand with simple boolean

PR #1761, opened 2018-11-07
Bank transfer e2e test

trival:


app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
```diff
@@ -0,0 +1,46 @@
...
+      cy.wait(1000);
+
+      cy.get(tid("invest-modal-gas-cost")).should($e => {
+        expect($e.text()).to.contain("ETH 0.0000"); // no gas cost on bank transfer
+      });
+      cy.get(tid("invest-modal-invest-now-button")).click();
+
+      cy.get(tid("invest-modal-bank-transfer-summary-amount")).should($m => {
+        expect($m.text())
+          .to.contain("123")
```
desfero: Any reason to not check that together? Like `should('contain', '123 €')`?

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
```diff
@@ -0,0 +1,46 @@
...
+      cy.get(tid("eto-invest-now-button-" + PUBLIC_ETO_ID)).click();
+      cy.get(tid("investment-type.selector.BANK_TRANSFER")).check({ force: true });
+      cy.wait(1000);
+
+      cy.get(tid("invest-modal-eur-field"))
+        .clear()
+        .type("123");
+      cy.wait(1000);
+
+      cy.get(tid("invest-modal-gas-cost")).should($e => {
```
desfero: Any reason you use the function as a parameter over `should('contain' 'ETH 0.0000')`? Same in 31 and 40 :)

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
```diff
@@ -0,0 +1,46 @@
...
+          .to.contain("123")
+          .and.to.contain("€");
+      });
+
+      cy.get(tid("invest-modal-summary-confirm-button")).click();
+      cy.get(tid("invest-modal-bank-transfer-details-title"));
+      cy.get(tid("invest-modal-bank-transfer-details-amount")).should($m => {
+        expect($m.text().trim())
+          .to.contain("123")
+          .and.length(6); // 123.00 or 123,00
```
desfero: Why not to check exact match? `should('contain', 123.00)`. Does decimal separator depend on some external state?

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
trival: I thought it would be hard to spot errors if we made a change to EUR for example. Only the actual value is important here i think.

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
trival: yes, editor autocompletion : ) I don't like magic strings

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
trival: there are plans to change decimal separator depending on locale soon afaik.

trival: but a more pecise regex could be better here.

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
trival: But, OK, after refactoring, I see that it is more concise 👍

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
trival: here, the match with the short string form did not work...

app/test-e2e/x-eto/InvestWithBankTransfer.spec.e2e.ts
```diff
@@ -0,0 +1,46 @@
...
+    createAndLoginNewUser({
+      type: "investor",
+      kyc: "business",
+      seed: INV_EUR_ICBM_HAS_KYC_SEED,
+      clearPendingTransactions: true,
+    }).then(() => {
+      cy.visit("/dashboard");
+      // click invest now button
+      cy.get(tid("eto-invest-now-button-" + PUBLIC_ETO_ID)).click();
+      cy.get(tid("investment-type.selector.BANK_TRANSFER")).check({ force: true });
```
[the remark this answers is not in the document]
trival: the order is important, since changing the investment type resets the money inputs. I would keep it explicit here.

PR #1764, opened 2018-11-07
Update identity modal description

desfero:


SergejSoft:
can we add a headline text: Your Identity is now stored on the blockchain.

Issue #1756, opened 2018-11-07
[bookbuilding] fix behaviour when pledge was rejected in metamask


glupiawymowka:
.net, metamask

1. Go to bookbuilding and type an amount, then "back now"
2. First problem - the window below is unclear. It says "confirm your pledge" but no information WHERE to do it. I would click "CANCEL" automatically as this is what I see first (same problem as the other similar windows). My suggestion, *at least* put "Go to your external wallet to confirm your pledge" 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/8bbb5a99-3a77-43c1-ba7f-41c516839391)

3. Go to signature request in metamask, then cancel. You'll get an error bar saying that saving the pledge failed. Close the "confirm your pledge" window

4. Check the image below to see what stays in this case. 
     Basically, there is a strange empty field you can't edit, and a "change" or "delete" button. 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/fc8aa859-8986-44d6-b014-a522e8c81296)

Additional : it gets even crazier when you choose "delete" at this point to get rid of the empty field. This forces another signature request which you could also cancel and go like this forever

Maybe it should automatically go back to the very first state of bookbuilding if the signing was canceled?

desfero:
@glupiawymowka ok, fixed that second issue as it was quite simple to do. Still, we need to improve error message per your request but I'm waiting for some design or text from @SergejSoft.
As a note, if we want to have error message like `Go to your external wallet to confirm your pledge` then we need to do some core changes as right now we have same title and message for lightwallet, metamask and ledger.

glupiawymowka:
second part is done, another ticket to be created for the first one

glupiawymowka:
First point is no longer valid regarding upcoming wording changes

PR #1781, opened 2018-11-10
#1756 Fix behaviour when pledge was rejected

desfero:


rudolfix:
@desfero will it also work in case of http error status on /pledge endpoint? I just had this situation where backend returned error 500 and got into the same state as per ticket #1756 
image

desfero:
@rudolfix yeap, that should also fix the issue you mentioned as the transition to view is done after the pledge was saved to redux store

PR #1803, opened 2018-11-12
At/#1788 losing data on backend error

AlexeyTuboltsev:
stop redirecting user to dashboard if form saving was not successful on the backend. 
remove some erroneous error messages ))
move all validation logic to EtoApi.interfaces.ts and let yup handle it


app/lib/api/eto/EtoApi.interfaces.tsx
```diff
@@ -123,7 +123,7 @@ export const EtoLegalInformationType = YupTS.object({
   companyStage: YupTS.string().optional(),
   numberOfFounders: YupTS.number().optional(),
   lastFundingSizeEur: YupTS.number().optional(),
-  companyShares: YupTS.number(),
+  companyShares: YupTS.number().enhance(v => v.min(100)),
```
Moejoe90: i would move all these into `constants`

Issue #1386, opened 2018-09-21
KYC Beneficial owner should have validation for `Percentage Held` field


desfero:
Background:
It's possible to save ownership field as an incorrect percentage value.
![image](https://user-images.githubusercontent.com/7031578/45867355-93a17e80-bd83-11e8-8ef9-277de2e4a85f.png)

glupiawymowka:
fixed

![image](https://user-images.githubusercontent.com/18315589/48557175-4e627e80-e8e6-11e8-9052-3ac48f33505e.png)
*tested on .io*

PR #1805, opened 2018-11-12
#1386 KYC - Add maximum value validator to beneficial ownership percentage

drakkein:


app/lib/api/KycApi.interfaces.ts
```diff
@@ -89,7 +89,7 @@ export interface IKycBeneficialOwner extends IKycPerson {
 }
 export const KycBeneficialOwnerSchema = KycPersonSchema.concat(
   Yup.object().shape({
-    ownership: Yup.number(),
+    ownership: Yup.number().max(100),
```
Moejoe90: lets make it a constant on top of the file with a clear name

app/lib/api/KycApi.interfaces.ts
```diff
@@ -7,6 +7,8 @@ import {
   restrictedCountry,
 } from "./util/schemaHelpers";
 
+const percentageLimit = 100;
```
Moejoe90: full upper case please :D

drakkein: It supposed to be ez 😆

app/lib/api/KycApi.interfaces.ts
```diff
@@ -87,9 +89,10 @@ export interface IKycBeneficialOwner extends IKycPerson {
   ownership?: number;
   id?: string;
 }
+
 export const KycBeneficialOwnerSchema = KycPersonSchema.concat(
   Yup.object().shape({
-    ownership: Yup.number(),
+    ownership: Yup.number().max(PERCENTAGE_LIMIT),
```
desfero: we should export percentage as a custom schema time and use everywhere we need to validate percentage, as here we should also set `min` to `0` as it's not possible to have a negative percentage value

desfero: Refer to implementation of https://github.com/Neufund/platform-frontend/blob/cae3d0ab274295a646e5e973e1c1de1fb62fc658/app/lib/api/eto/EtoApi.interfaces.tsx#L28

Issue #1789, opened 2018-11-11
implement UI for external pending (ooo) transactions


rudolfix:
When there are out of bounds transactions (`oooTransactions`) in (`PendingTxsSchema`) we still see this
![image.png](https://images.zenhubusercontent.com/5a72da804b5806bc2bc52ecc/d11f81a5-81d6-407d-a5ff-38efbde6ef16)
We should show a nice UI here. We can reuse existing UI below with different texting:
![image.png](https://images.zenhubusercontent.com/5a72da804b5806bc2bc52ecc/b9fb5cb4-39f8-4014-b752-9a0e4aacdedb)
Texting changes
PENDING TRANSACTION WITHDRAW -> EXTERNAL PENDING TRANSACTION DETECTED
text below -> You have submitted transaction outside of Neufund Platform and it still remains in the Ethereum pending pool waiting to be executed and included in new block. Neufund Platform prevents submitting more transactions if you have still pending one. In most cases this leads to unexpected problems as only one transaction with the same nonce will be mined. We'll wait until transaction specified below is mined.

Transaction Hash: take the top transaction of `oooTransactions`, display it + link to EtherScan

How to test it: I think you should just somehow mock `oooTransactions`

Issue #1790, opened 2018-11-11
[ADDITIONAL FIX] implement validation logic for external (ooo) pending transactions


rudolfix:
When watching pending transactions to be mined here
```
if (txs.oooTransactions.length) {
      const txHash = txs.oooTransactions[1].hash;
      // go to miner
      yield put(actions.txSender.txSenderWatchPendingTxs(txHash));
      logger.info(`Waiting for out of bound transactions: ${txs.oooTransactions.length}`);
      yield neuCall(watchPendingOOOTxSubSaga, txHash);
    }
```
- please fix index `txs.oooTransactions[1].hash;`
- please run this in a loop. `oooTransactions` is an array and all elements must be mined to proceed

It would be good if you can mock the `oooState` at least as far as to if statement executes once and starts watching. 

-------------

first #1789  must be completed

then we need to properly validate if transactions returned by backend are really pending. we do the same for our transactions. what backend is returning is just indication. source of truth is Ethereum and client needs to check it.

if there are any transactions in `oooTransactions` for each of those please retrieve the receipt. If transaction is mined, please skip it.

If you skipped all transactions in `oooTransactions` then go to next step in transaction execution flow.

Please do such validation on each new block and of course when you open the dialog

PR #1820, opened 2018-11-13
Pending ooo transaction modal

trival:
fixes
https://github.com/Neufund/platform-frontend/issues/1789
https://github.com/Neufund/platform-frontend/issues/1790

additionally, tweaks the husky precommit task to only lint and format staged files


.huskyrc
```diff
@@ -1,5 +1,5 @@
 {
   "hooks": {
-    "pre-commit": "yarn lint:fix"
+    "pre-commit": ["concurrently -p \"[{name}]\" -n \"TSC,LINT-STAGED\" -c \"bgYellow.bold,bgMagenta.bold\" \"yarn tsc\" \"lint-staged\""]
```
Moejoe90: :+1:

app/components/modals/tx-sender/TxSender.tsx
```diff
@@ -102,7 +102,7 @@ const SuccessComponent: React.SFC<{ type?: ETxSenderType; txHash?: string }> = (
 function renderBody({ state, blockId, txHash, type, error }: Props): React.ReactNode {
   switch (state) {
     case ETxSenderState.WATCHING_PENDING_TXS:
-      return <WatchPendingTxs />;
+      return <WatchPendingTxs txHash={txHash!} blockId={blockId!} />;
```
Moejoe90: I'm a bit weary from the fact that we assume that txHash and blockId is there 
This is probably part of the problem of <partial> i guess

app/components/modals/tx-sender/shared/WatchPeningTxs.tsx
```diff
@@ -1,10 +1,35 @@
...
-    <LoadingIndicator />
-  </>
+import * as styles from "./TxPending.module.scss";
+
+export interface IProps {
+  txHash: string;
+  blockId: number;
+}
+
+export const WatchPendingTxs: React.SFC<IProps> = ({ txHash, blockId }) => (
```
Moejoe90: For sure this won't generate an Error is txHash, blockId we undefined

app/modules/tx/monitor/sagas.ts
```diff
@@ -53,6 +56,61 @@ function* txMonitor({ logger, notificationCenter }: TGlobalDependencies): any {
...
+      type: EEventEmitterChannelEvents.NEW_BLOCK;
+      blockId: number;
+    }
+  | {
+      type: EEventEmitterChannelEvents.TX_MINED;
+      tx: Web3.Transaction;
+    }
+  | {
+      type: EEventEmitterChannelEvents.ERROR;
+      error: any;
```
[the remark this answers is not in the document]
trival: it is unclear, since in javascript you can actually throw anything you like, and since this comes from the external web3 library, we cannot be certain I guess...

Issue #1793, opened 2018-11-11
on LockedAccount migration or upgrade show AMENDED ICBM  RESERVATION AGREEMENT


rudolfix:
1. in Upgrade flow:
On transaction confirmation dialog  - exactly as for Investment Transaction
2. in Migration tool:
Above "Generate Transaction" button but only visible when the button is not disabled

Name of the agreement AMENDED ICBM  RESERVATION AGREEMENT

Please allow investor to download agreement via ipfs hash
Hash is available from `LockedAccount::currentAgreement()` - see ABI
If there's `ipfs:` prefix on the hash then remove it.

download it via immutable storage as pdf. that's it

PR #1835, opened 2018-11-14
#1793 Add Amended ICBM reservation agreement download

drakkein:


app/modules/wallet/sagas.ts
```diff
@@ -88,7 +89,37 @@ function* walletBalanceWatcher(): any {
...
+    apiImmutableStorage,
+    intlWrapper: {
+      intl: { formatIntlMessage },
+    },
+  }: TGlobalDependencies,
+  action: TAction,
+): any {
+  if (action.type !== "WALLET_ICBM_DOWNLOAD_AGREEMENT") return;
+
+  const [, , agreementUri] = yield contractsService.euroLock.currentAgreement();
```
Moejoe90: It should be problems agreementUrl correct?

app/modules/wallet/sagas.ts
```diff
@@ -88,7 +89,37 @@ function* walletBalanceWatcher(): any {
...
+  yield neuCall(
+    downloadLink,
+    generatedDocument,
+    formatIntlMessage("wallet.icbm.reservation-agreement"),
+    ".pdf",
+  );
+}
+
 export function* walletSagas(): any {
+  yield fork(neuTakeEvery, "WALLET_ICBM_DOWNLOAD_AGREEMENT", downloadICBMWalletAgreement);
```
Moejoe90: This is good,
The only thing is moving this new aga and action to the icbmWallet folder instead of folder

PR #1846, opened 2018-11-15
#1528 Portfolio change placeholders 

drakkein:


app/components/portfolio/PortfolioLayout.tsx
```diff
@@ -100,6 +100,9 @@ const PortfolioLayout: React.SFC<TPortfolioLayoutProps> = ({
       <Col className="mb-4">
         <NewTable
           keepRhythm={true}
+          placeholder={
+            <FormattedMessage id="portfolio.section.reserved-assets.table.header.placeholder" />
```
AlexeyTuboltsev: That would be to easy! Unfortunately placeholders only accepts a string, not a ReactNode. We have to use injectIntlHelpers HOC and formatIntlMessage for it. Example: app/components/eto/dashboard/choosePreEtoDateWidget/ChoosePreEtoDateWidget.tsx

drakkein: As I can see in types.d.ts TTranslatedString is accepting elements
`export type TTranslatedString = string | React.ReactElement<FormattedMessage>;`

Issue #1809, opened 2018-11-13
[mobile, landing page] register & login buttons are half-hidden


glupiawymowka:
On mobile (vertical position), register & login buttons are only partly visible (but still clickable)

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/f49d3fc8-6b35-44ce-8722-f94ba0132950)

glupiawymowka:
fix'd

PR #1864, opened 2018-11-19
#1809 Fix login and register buttons display on mobile views

drakkein:


app/components/layouts/Header.module.scss
```diff
@@ -13,6 +17,10 @@
     left: .75rem;
   }
 
+  @include media-breakpoint-down(sm) {
```
drakkein: hiding logo text form mobile view enables us to have working version even on iphone se

Issue #1870, opened 2018-11-19
[USABILITY FIX] LOGIN | Ledger - add paragraph


glupiawymowka:
Under "Select your wallet for registration" (to be changed by Aga to "Select your wallet to login") headline, add the following paragraph and format with current styling:

“Your Ledger Nano has many addresses. Your ETH and NEU balance is displayed next to each address. Please choose the one you have registered with.”

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/83be8dcc-60f6-44ee-85a9-7008912d7ae6)

glupiawymowka:
all good

![image](https://user-images.githubusercontent.com/18315589/48910592-a6fac400-ee70-11e8-8150-1571289be361.png)

Issue #1867, opened 2018-11-19
[USABILITY FIX] REGISTER | Ledger - add paragraph 


glupiawymowka:
Under "Select your wallet for registration" headline, add the following paragraph and format with current styling:

“Your Ledger Nano has many addresses. Your ETH and NEU balance is displayed next to each address. Please choose the one you want to register with.”

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/5306e3a6-d09f-484d-a208-96dc7707214a)

AlexeyTuboltsev:
also solves #1870

glupiawymowka:
done <3

PR #1880, opened 2018-11-19
[WIP] At/#1867 add text to ledger multiple addresses

AlexeyTuboltsev:
#1867 #1870 
Registration/Login: Add explanation text if ledger has multiple addresses


app/components/wallet-selector/ledger/WalletLedgerChooserComponent.stories.tsx
```diff
@@ -0,0 +1,40 @@
+import { storiesOf } from "@storybook/react";
+import * as React from "react";
```
trival: 👍

Issue #1903, opened 2018-11-20
[USABILITY FIX] REGISTER/LOGIN WITH METAMASK - visual fixes


glupiawymowka:
Both for login and register: 

1. Please remove step 1 completely
2. Please make steps 2 & 3 into 1 & 2, then center them
3. Please delete metamask icon below

![image](https://user-images.githubusercontent.com/18315589/48765060-ac5de000-ecb1-11e8-8913-53aa3a786e7f.png)

glupiawymowka:
done

![image](https://user-images.githubusercontent.com/18315589/48915527-43c45e00-ee7f-11e8-9d59-20d1f2c21b23.png)

PR #1922, opened 2018-11-20
#1903 visual fixes to metamask register screen

AlexeyTuboltsev:
remove metamask logo and one of three step imgs.
btw rebuild StepCard element to remove markup clutter

### DIFF
--- a/app/components/shared/StepCard.module.scss
+++ b/app/components/shared/StepCard.module.scss
@@ -4,4 +4,8 @@
   justify-content: center;
   align-items: center;
   align-self: start;
+
+  span {
+    text-align: center;
+  }
 }
--- a/app/components/shared/StepCard.module.scss
+++ b/app/components/shared/StepCard.module.scss
@@ -1,0 +1,7 @@
+.step-card {
+  display: flex;
+  flex-flow: column;
+  justify-content: center;
+  align-items: center;
+  align-self: start;
+}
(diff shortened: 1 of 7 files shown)


app/components/shared/StepCard.tsx
```diff
@@ -1,19 +1,14 @@
...
-        <img src={img} className="mb-3" />
-      </Col>
-    </Row>
-    <Row>
-      <Col>{text}</Col>
-    </Row>
-  </Col>
+  <div className={styles.stepCard}>
+    <img src={img} className="mb-3" />
+    <div>{text}</div>
```
drakkein: Maybe we should change this div to paragraph or span

app/components/shared/StepCard.module.scss
```diff
@@ -0,0 +1,7 @@
+.step-card {
```
drakkein: I think we should add center align for text like it is in Ledger steps

app/components/wallet-selector/browser/WalletBrowser.tsx
```diff
@@ -77,11 +75,10 @@ export const WalletBrowserComponent: React.SFC<
             </Row>
           </>
         )}
-        <Row className="mb-4 text-center">
-          <StepCard img={walletIcon} text={formatIntlMessage("wallet-selector.browser.steps.1")} />
+        <div className={styles.stepCardWrapper}>
           <StepCard img={browserIcon} text={formatIntlMessage("wallet-selector.browser.steps.2")} />
```
drakkein: StepCard text is accepting React.node so we were able to use but this is also OK, no need to change. 👍

PR #1925, opened 2018-11-20
Investment fixes

trival:
* new env var to change time when bank transfer is disabled
* take min eto ticket from calculated contributions


app/modules/investor-tickets/selectors.ts
```diff
@@ -76,6 +76,11 @@ export const selectEquityTokenCountByEtoId = (etoId: string, state: IAppState) =
   return contrib && contrib.equityTokenInt.toString();
 };
 
+export const selectEtoMinTicketEurUlpsById = (etoId: string, state: IAppState) => {
```
rudolfix: please take both min and max ticket here!

trival: 👍

PR #1938, opened 2018-11-20
kyc updates

huesforalice:
* Adds text to first kyc page
* Fixes the upload form
* Make the submit success screen more clear


app/components/shared/UploadedFiles.tsx
```diff
@@ -11,6 +11,7 @@ interface IProps {
 
 export const UploadedFiles: React.SFC<IProps> = ({ files }) => (
   <div className={cn("c-uploaded-files", styles.uploadedFiles)}>
+    <h6>Uploaded files</h6>
```
trival: can we also translate this text?

huesforalice: done

PR #1942, opened 2018-11-20
Logsout/Login from all open browser session 

Moejoe90:
This adds automatic logout/login using local storage


app/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+export function* startRedirectChannel(): any {
+  window.addEventListener("storage", (evt: StorageEvent) => {
+    if (evt.key === STORAGE_JWT_KEY && evt.oldValue && !evt.newValue) {
+      redirectChannel.put({
+        type: EUserAuthType.LOGOUT,
+      });
+    }
+    if (evt.key === USER_JWT_KEY && !evt.oldValue && evt.newValue) {
```
Moejoe90: I had to add a new user-storage because relying on STORAGE_JWT_KEY creates a whole lot of mess

app/modules/init/reducer.ts
```diff
@@ -1,7 +1,11 @@
 import { AppReducer } from "../../store";
 import { DeepReadonly } from "../../types";
 
-export type TInitType = "appInit" | "smartcontractsInit" | "walletInit";
+export enum EInitType {
+  appInit = "appInit",
```
desfero: Just for consistency can we use `APP_INIT`, `SMART_CONTRACTRS_INIT` and `WALLET_INIT`?

app/test-e2e/general-app/auto-actions.spec.e2e.ts
```diff
@@ -0,0 +1,38 @@
+import { createAndLoginNewUser, JWT_KEY, INVESTOR_WALLET_KEY } from "../utils/userHelpers";
+import { goToDashboard, assertUserInDashboard, assertUserInLanding } from "../utils/index";
+
+describe("auto-logout/auto-login", () => {
+  it("should logout automatically when a user logs out from another tab", () => {
```
desfero: 👍

app/modules/terms-of-service-modal/sagas.ts
```diff
@@ -0,0 +1,81 @@
+import { Effect, fork, put, select } from "redux-saga/effects";
+import { AuthMessage } from "./../../components/translatedMessages/messages";
```
desfero: should be below

app/test-e2e/general-app/auto-actions.spec.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+      assertUserInLanding();
+      createAndLoginNewUser({
+        type: "investor",
+        kyc: "business",
+      }).then(() => {
+        const jwt = window.localStorage.getItem(JWT_KEY);
+        const walletData = window.localStorage.getItem(INVESTOR_WALLET_KEY);
+        cy.clearLocalStorage();
+        cy.wait(5000);
+        cy.window().then(Windows => {
```
desfero: maybe `win`? As `Windows` mean that it's a contructor

app/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+      throw new Error("Message signing failed");
+    }
+  }
+}
+
+/**
+ * Multi browser logout/login feature
+ */
+
+const redirectChannel = channel();
```
desfero: better type safety?

```suggestion
const redirectChannel = channel<{ type: EUserAuthType }>();
```

app/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Saga that starts an Event Channel Emitter that listens to storage
+ * events from the browser
+ */
+
+export function* startRedirectChannel(): any {
+  window.addEventListener("storage", (evt: StorageEvent) => {
+    if (evt.key === STORAGE_JWT_KEY && evt.oldValue && !evt.newValue) {
```
desfero: if there is a problem with `STORAGE_JWT_KEY` can we for consistency use `USER_JWT_KEY` here?

Moejoe90: I don't want to do that. At least initially I didn't because logout is completely dependant on `STORAGE_JWT_KEY`. This is newly introduced `USER_JWT_KEY` which should be called `USER_KEY`

app/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+    const userAction = yield take(redirectChannel);
+    switch (userAction.type) {
+      case EUserAuthType.LOGOUT:
+        yield put(actions.auth.logout());
+        break;
+      case EUserAuthType.LOGIN:
+        yield put(actions.init.start(EInitType.appInit));
+        break;
+    }
+    yield delay(5000);
```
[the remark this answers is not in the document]
Moejoe90: indeed it passed through me actually

app/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,153 @@
+import { channel, delay } from "redux-saga";
+import { call, Effect, put, select, take } from "redux-saga/effects";
+import { USER_JWT_KEY } from "./../../../lib/persistence/UserStorage";
```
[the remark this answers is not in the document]
Moejoe90: done

app/modules/init/reducer.ts
Moejoe90: yes we can

app/test-e2e/general-app/auto-actions.spec.e2e.ts
Moejoe90: Windows was just an arbitrary name its removed completely now

app/modules/init/reducer.ts
Moejoe90: I found out the init reducer relies on `[]` arrays so,I will keep that names like this for now

desfero: hm, but why it's a problem to change enum key name?

app/modules/auth/jwt/sagas.ts
desfero: what about this one?

app/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+    const userAction = yield take(redirectChannel);
+    switch (userAction.type) {
+      case EUserAuthType.LOGOUT:
+        yield put(actions.auth.logout());
+        break;
+      case EUserAuthType.LOGIN:
+        yield put(actions.init.start(EInitType.appInit));
+        break;
+    }
+    yield delay(REDIRECT_CHANNEL_WATCH_DELAY);
```
desfero: This should probably be inside `while` loop?

app/test-e2e/general-app/auto-actions.spec.e2e.chrome.ts
```diff
@@ -0,0 +1,37 @@
+import { assertUserInDashboard, assertUserInLanding, goToDashboard } from "../utils/index";
```
desfero: We don't use `-` in tests file names, so maybe `AutomaticLoginAndLogout`

app/modules/auth/jwt/sagas.ts
Moejoe90: this didn't workout

app/modules/init/reducer.ts
Moejoe90: i don't want the state to have kebap case as `APP_INIT`

app/modules/auth/jwt/sagas.ts
Moejoe90:
```
Argument of type '{ type: typeof EUserAuthType; }' is not assignable to parameter of type 'Buffer<{}>'.
  Object literal may only specify known properties, and 'type' does not exist in type 'Buffer<{}>'. [2345]
```

Issue #1948, opened 2018-11-20
Public View. Pitchdeck, documents


SergejSoft:
1.I suggest to completely hide it if we don't find a solution to make it work. 
<img width="636" alt="screen shot 2018-11-20 at 21 18 18" src="https://user-images.githubusercontent.com/2290474/48800485-2de15c80-ed0a-11e8-832a-84475d01fe0a.png">
2. Some documents don't have a link to view.  (lightpaper and pitch deck)
<img width="447" alt="screen shot 2018-11-20 at 21 21 24" src="https://user-images.githubusercontent.com/2290474/48800559-5ff2be80-ed0a-11e8-995e-ce77b843c8ed.png">

PR #1967, opened 2018-11-21
Hides company links without urls

Moejoe90:
#related https://github.com/Neufund/platform-frontend/issues/1948

I found that this happens because there are no url's so now this hides the url's. The solution to change the data structure is a bit more complicated


app/components/eto/public-view/DocumentsWidget.tsx
```diff
@@ -45,15 +45,17 @@ const DocumentsWidgetLayout: React.SFC<TDispatchProps & TExternalProps & CommonH
...
-                  name={link.title || ""}
-                  altIcon={<InlineIcon svgIcon={icon_link} />}
-                />
-              </Col>
-            ))}
+            companyMarketingLinks.map((link, i) => {
+              return link.url && link.url !== "" ? (
+                <Col sm="6" md="12" lg="6" key={i} className={styles.document}>
+                  <DocumentLink
+                    url={link.url || ""}
```
drakkein: We don't need optional empty string, cuz we are not rendering anyway when url is empty

app/components/eto/public-view/DocumentsWidget.tsx
```diff
@@ -45,15 +45,17 @@ const DocumentsWidgetLayout: React.SFC<TDispatchProps & TExternalProps & CommonH
...
-                <DocumentLink
-                  url={link.url || ""}
-                  name={link.title || ""}
-                  altIcon={<InlineIcon svgIcon={icon_link} />}
-                />
-              </Col>
-            ))}
+            companyMarketingLinks.map((link, i) => {
+              return link.url && link.url !== "" ? (
+                <Col sm="6" md="12" lg="6" key={i} className={styles.document}>
```
drakkein: I think we should use something more unique than index for a key, maybe url

Moejoe90: hmmm for now I would keep index the url can be repeated so its not unique

app/components/eto/public-view/DocumentsWidget.tsx
Moejoe90: While you have a point i don't want to do any refactoring and `DocumentLink` expects both types

Issue #1962, opened 2018-11-21
[IMPORTANT] [BUG] Please be more resilient to node bugs when mining the transaction


rudolfix:
We should not go to REVERT state
![image](https://user-images.githubusercontent.com/17202864/48848349-fe7f2e00-eda3-11e8-85ea-d33b48d69e9d.png)
when
1. we ask for a block and node returned HTTP error or there is connection error
2. we ask for transaction hash or receipt and node returned error (as above)
3. any other case of temporary error

in case of temporary error we should just keep the window open and mining.

steps to reproduce
1. start to withdraw ether on stage
2. switch off WIFI when still waiting for the block

error message will appear as above

PR #1970, opened 2018-11-21
Fixes TxMiner exit on error behavior

Moejoe90:
Related https://github.com/Neufund/platform-frontend/issues/1962
Tested with the same approach mention in ticket 
Take note that: Turning off the wifi now will keep the modal running but everything else will break

### DIFF
--- a/app/modules/tx/sender/sagas.ts
+++ b/app/modules/tx/sender/sagas.ts
@@ -84,6 +84,7 @@ export function* txSendProcess(
     yield call(connectWallet);
     yield put(actions.txSender.txSenderWalletPlugged());
     const txHash = yield neuCall(sendTxSubSaga);
+
     yield neuCall(watchTxSubSaga, txHash);
   } catch (error) {
     logger.error(error);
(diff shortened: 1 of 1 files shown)


app/modules/tx/sender/sagas.ts
```diff
@@ -257,16 +258,16 @@ function* watchTxSubSaga({ logger }: TGlobalDependencies, txHash: string): any {
...
         case EEventEmitterChannelEvents.ERROR:
           logger.error("Error while tx watching: ", result.error);
-          return yield put(
-            actions.txSender.txSenderError(ETransactionErrorType.ERROR_WHILE_WATCHING_TX),
-          );
+          break;
+        // Terminal errors - Tx Mining should exit
         case EEventEmitterChannelEvents.OUT_OF_GAS:
-          logger.error("Error Transaction Reverted: ", result.error);
+          logger.info("Error Transaction out of gas: ", result.error);
```
rudolfix: is it going to sentry? we should definitely log out of gas and transaction hash as well!

app/modules/tx/sender/sagas.ts
```diff
@@ -257,16 +258,16 @@ function* watchTxSubSaga({ logger }: TGlobalDependencies, txHash: string): any {
...
           return yield put(actions.txSender.txSenderError(ETransactionErrorType.OUT_OF_GAS));
         case EEventEmitterChannelEvents.REVERTED_TRANSACTION:
-          logger.error("Error Transaction Reverted: ", result.error);
+          logger.warn("Error Transaction Reverted: ", result.error);
```
rudolfix: is it going to sentry? we should definitely log REVERT and transaction hash as well!

app/modules/tx/sender/sagas.ts
desfero: @rudolfix info is not immediately logged to sentry, only logged as a part of breadcrumb when error/warning occurs in the same session

app/modules/tx/sender/sagas.ts
desfero: Yes, `warn`'s are logged to sentry

app/modules/tx/sender/sagas.ts
desfero: Will change it to `warn`

Issue #1910, opened 2018-11-20
[USABILITY FIX] DASHBOARD - please add paragraph in MyPortfolio Widget


glupiawymowka:
1. Please hide "My Portfolio" Widget label

2. Please add the following paragraph below "Welcome to Neufund" header 

“From the dashboard, you can view your portfolio, check your wallet balances, and view available investment opportunities.", styled as the other paragraphs, black 

![image](https://user-images.githubusercontent.com/18315589/48768191-d36be000-ecb8-11e8-8b43-14235a3306dc.png)

glupiawymowka:
done
![image](https://user-images.githubusercontent.com/18315589/48939810-c8fa5200-ef15-11e8-9ca5-e0a8a6737f22.png)

Issue #1899, opened 2018-11-20
[USABILITY FIX] KYC WIDGET - new wording + link


glupiawymowka:
Aga wanted to change the text, but I still can't see it. 
If there is another text than given below, please change it into this one: 

1. PENDING VERIFICATION - 

"Thank you for submitting your identity verification request! We will review your documents within 1 business day and contact you if we require further information. Otherwise, we will notify you by email once your account has been successfully verified.

To contact support, please click here.” ---> link to freshdesk ticket

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/c83ab6da-ac71-4455-b459-5a6da7f80da3)

2. REJECTED VERIFICATION
“We regret to inform you that after careful review, your identity verification request has been rejected. To contact support, please click here” ---> link to freshdesk ticket

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/f89324de-5c8f-4cf1-bf4c-7138ad25ca98)

In both cases, please make "HERE" clickable, and link it to : https://support.neufund.org/support/tickets/new

SergejSoft:
TEST rejected Verifications.

PR #2003, opened 2018-11-22
#1910 #1899 add/change translations to kyc panel and to myPortfolio

AlexeyTuboltsev:
#1910 #1899 
add/change translations to kyc panel and to myPortfolio


app/components/dashboard/my-portfolio/MyPortfolioWidget.tsx
```diff
@@ -58,13 +61,7 @@ export const MyPortfolioWidgetComponent: React.SFC<IProps> = ({
...
-    <Panel
-      headerText={
-        <FormattedMessage id="components.dashboard.my-portfolio.my-portfolio-widget.header-text" />
-      }
-      className={className}
-      style={style}
-    >
+    <Panel className={className} style={style}>
       <>
         {isLoading ? (
```
drakkein: Again It's not yours, but since you are doing changes here we could do the loading indication with a branch from recompose

Issue #1795, opened 2018-11-12
[Settings] Align Your Ethereum Address


glupiawymowka:
Address should be centered regarding area under headline not whole container height 

1. CURRENT LOOK 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/f6104086-1fe5-4dc1-a36e-8f8815141eaa)

2. DESIGN 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/c69fa45c-89c5-47d8-a977-4f74de9a3d0b)

glupiawymowka:
not centered (tested on stagig and .io)
![image](https://user-images.githubusercontent.com/18315589/48779246-7895b100-ecd7-11e8-9cfe-7978ba3a8a34.png)

glupiawymowka:
done <3

PR #2004, opened 2018-11-22
#1795 Fix wallet address align

drakkein:


drakkein:
@huesforalice There are only stories related to AccountAddress component itself, without panel wrapper. I guess we should add version with panel to be sure about it's behavior.

PR #2052, opened 2018-11-25
adds support for terms of service

huesforalice:


app/components/layouts/LayoutAuthorized.tsx
```diff
@@ -10,6 +10,7 @@ import { Footer } from "./Footer";
 import { Header } from "./Header";
 import { LayoutAuthorizedMenu } from "./LayoutAuthorizedMenu";
 
+import { AcceptTosModal } from "../modals/accept-tos-modal/AcceptTosModal";
```
desfero: please move to imports above

app/components/modals/ModalComponentBody.tsx
```diff
@@ -14,7 +14,7 @@ export const ModalComponentBody: React.SFC<IModalComponentProps> = ({
 }) => (
   <Panel className={styles.modal} {...props}>
     <div className={styles.header}>
-      <ButtonClose data-test-id="modal-close-button" onClick={onClose} />
+      {onClose && <ButtonClose data-test-id="modal-close-button" onClick={onClose} />}
```
desfero: If we don't have onClose then we should probably hide all `header` otherwise we have empty div

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from "react";
+import { FormattedMessage } from "react-intl-phraseapp";
+import { Modal } from "reactstrap";
+
+import { actions } from "../../../modules/actions";
```
desfero: we don't need spaces here between imports, all of them are a part of single category (`own`)

app/lib/web3/ContractsService.ts
```diff
@@ -103,11 +102,7 @@ export class ContractsService {
  * In dev mode it will validate contract code to ease web3 development pains. In prod it will assume that address is correct, saving some network calls.
  */
 async function create<T>(ContractCls: IContractCls<T>, web3: any, address: string): Promise<T> {
-  if (process.env.NODE_ENV === "production") {
-    return new ContractCls(web3, address);
-  } else {
-    return await ContractCls.createAndValidate(web3, address);
```
desfero: Any reason we don't want to validation contract on dev during creation?

app/test-e2e/settings/LightWalletRegister.spec.e2e.ts
```diff
@@ -67,14 +68,16 @@ describe("Light wallet login / register", () => {
     });
   });
 
-  it("should return an error when logging with same email", () => {
+  it.only("should return an error when logging with same email", () => {
```
desfero: .only, lint's should fail

app/test-e2e/settings/LightWalletRegister.spec.e2e.ts
```diff
@@ -67,14 +68,16 @@ describe("Light wallet login / register", () => {
...
     const email = convertToUniqueEmail("dave@neufund.org");
     const password = "strongpassword";
 
     // register once and then verify email account
     cy.visit("/register");
     typeEmailPassword(email, password);
     assertUserInDashboard();
+    acceptTOS();
```
desfero: 🥇 for updating tests

app/test-e2e/settings/LightWalletRegister.spec.e2e.ts
```diff
@@ -67,14 +68,16 @@ describe("Light wallet login / register", () => {
...
+    cy.wait(5000); // wait for the tos to settle
```
desfero: we can hide this to `acceptTos`. Less `.wait` in main code is always better and more clear. BTW do we need to wait 5 seconds? Maybe 1s is enough? Or it's possible to assert that tos is settled on UI?

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+  onDownloadTos = () => {
+    this.setState({
+      tosDownloaded: true,
+    });
+    this.props.onDownloadTos();
+  };
+
+  render(): React.ReactNode {
+    return (
+      <div className="text-center">
```
desfero: `section` tag should be used here

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+  <Modal isOpen={props.isOpen} centered>
+    <ModalComponentBody>
+      <AcceptTosModalInner {...props} />
+    </ModalComponentBody>
+  </Modal>
+);
+
+export const AcceptTosModal = appConnect<IStateProps, IDispatchProps>({
+  stateToProps: s => ({
+    isOpen: !!(s.auth.user && s.auth.user.latestAcceptedTosIpfs !== s.auth.currentAgreementHash),
```
desfero: can we move it to selectors, like 

```
isOpen: selectIsLatestAgreementAccepted(s),
hasEverAccepted: selectIsAgreementAccepted(s),
```

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+    hasEverAccepted: !!(s.auth.user && s.auth.user.latestAcceptedTosIpfs),
```
desfero: strange, this one is probably not used and is even not in `IStateProps`

app/modules/auth/sagas.ts
```diff
@@ -254,6 +255,42 @@ function* handleSignInUser({
   }
 }
 
+function* handleAcceptCurrentAgreement({
+  apiUserService,
+  intlWrapper: {
+    intl: { formatIntlMessage },
+  },
+}: TGlobalDependencies): Iterator<any> {
+  const currentAgreementHash = yield select((s: IAppState) => s.auth.currentAgreementHash);
```
desfero: unfortunately, due to the way iterators works, we need to type correctly `currentAgreementHash`. I need to check in future whether it's possible to disable `any` in variables as we did with function `arguments` :)


Also, please use selectors here yield select(selectCurrentAggrementHash);

Same on line `280`

app/modules/auth/sagas.ts
```diff
@@ -254,6 +255,42 @@ function* handleSignInUser({
...
+  yield neuCall(
+    ensurePermissionsArePresent,
+    [SIGN_TOS],
+    formatIntlMessage("settings.modal.accept-tos.permission.title"),
+    formatIntlMessage("settings.modal.accept-tos.permission.text"),
+  );
+  const user: IUser = yield apiUserService.setLatestAcceptedTos(currentAgreementHash);
```
desfero: Do you think it's worth to add some logging here in case apiUserService throws an error? And also we can show some notification like ('There was a problem with accepting Terms and Conditions')

app/components/layouts/LayoutAuthorized.tsx
huesforalice: done

app/components/modals/ModalComponentBody.tsx
huesforalice: It looks better this way, as the margin to the top will be wrong if we remove the div. I had it that way before.

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
huesforalice: done (I think)

app/lib/web3/ContractsService.ts
huesforalice: this is an error, I will reset it.

app/test-e2e/settings/LightWalletRegister.spec.e2e.ts
huesforalice: done

app/test-e2e/settings/LightWalletRegister.spec.e2e.ts
huesforalice: We only need it in the cases where we log out or navigate away right after accepting the modal, which is only here. Otherwise it will slow down all the other tests.

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
huesforalice: Done

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
huesforalice: yeah, I planned to use it, removed it now.

app/modules/auth/sagas.ts
huesforalice: done

app/modules/auth/sagas.ts
huesforalice: done

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
huesforalice: done

Issue #2046, opened 2018-11-25
Direct bank transfer popup - wording + tooltip


Vanilizer:
Alexey, 

as we suspected copy cannot handle too many "bolds" in the text. As you can see in the screenshot (this is on production by today) In the first text paragraph you can see <STRONG> in text. 

Could you fix it the way that anything marked STRONG should be bold and orange? 

This is also connected to adding a tool tip to the check box: https://github.com/Neufund/platform-frontend/issues/2035

![image](https://user-images.githubusercontent.com/26923972/48980395-8af15f80-f0c8-11e8-8734-cdc5ff8b868c.png)

glupiawymowka:
*tested on .io*

![image](https://user-images.githubusercontent.com/18315589/49159042-dce1f180-f323-11e8-9cd3-822daaa75f9d.png)

Issue #2035, opened 2018-11-23
Add a tooltip in Direct Bank Transfer Popup


Vanilizer:
Please add a tooltip with following text to the check box text:

"Transaction fees are necessary to process transactions on the Ethereum Network. These fees are paid in ETH. For your convenience, we convert your 10 EUR to ETH to cover your initial transaction costs. Please keep in mind that you may need to deposit further ETH at a later point depending on your transaction activity."

![image](https://user-images.githubusercontent.com/26923972/48951285-45a12680-ef3e-11e8-99e5-6b61f7291e4b.png)

glupiawymowka:
*tested on .io*
![image](https://user-images.githubusercontent.com/18315589/49159188-33e7c680-f324-11e8-9354-f236348188ff.png)

PR #2057, opened 2018-11-26
#2035 #2046 direct bank transfer tooltip

AlexeyTuboltsev:
direct bank transfer:
add tooltip to checkbox label
fix display of text (change FormattedMessage to FormatedHTMLMessage)


AlexeyTuboltsev:
visual regression is ok

Issue #2015, opened 2018-11-23
BUG in widget behaviour


Vanilizer:
As a verified and not ICBM user I should be able participate in any public sale listing on Neufund. 

Eldo Token is in PUBLIC SALE STATE. However I see the progress bar and the button asking me to verify

This must be a bug. 

![image](https://user-images.githubusercontent.com/26923972/48934302-b7a84a00-ef03-11e8-9180-0b6ff8af83a4.png)

Moejoe90:
@rudolfix 
I just checked this ticket it appears to me that there might be some difference between the back-office KYC approval and on Chain Claim approval

rudolfix:
it's related #1424 - we react to onchain status here not to the backend. we should possibly return verified when both backedna and on-chain say so

Moejoe90:
@rudolfix can you check 
https://github.com/Neufund/platform-frontend/pull/2061
This does exactly that

rudolfix:
#2061 would solve temporarily our problem

PR #2061, opened 2018-11-26
adds on chain claim

Moejoe90:
This changes the kyc component to show approved only when on chain claim is approved.
IMO this should be done by backend as it is hacky this way but anyway should solve

https://github.com/Neufund/platform-frontend/issues/2015


app/modules/auth/selectors.ts
```diff
@@ -20,8 +21,8 @@ export const selectUnverifiedUserEmail = (state: IAuthState): string | undefined
 
 export const selectUser = (state: IAuthState): IUser | undefined => state.user;
 
-export const selectBackupCodesVerified = (state: IAuthState): boolean =>
-  !!state.user && !!state.user.backupCodesVerified;
+export const selectBackupCodesVerified = (state: IAppState): boolean =>
+  (!!state.auth.user && !!state.auth.user.backupCodesVerified) || !selectIsLightWallet(state.web3);
```
rudolfix: we should be super careful here because this gets written to backend database. Do we need to change the value for non-light wallets? previously they were `false`

app/modules/notifications/selectors.ts
```diff
@@ -21,9 +21,8 @@ export const selectIsActionRequiredSettings = (state: IAppState): boolean => {
   }
   return (
     !selectIsUserEmailVerified(state.auth) ||
-    (!selectBackupCodesVerified(state.auth) &&
-      selectWalletType(state.web3) === EWalletType.LIGHT) ||
-    !includes(["Outsourced", "Pending", "Accepted"], selectKycRequestStatus(state.kyc))
+    (!selectBackupCodesVerified(state) && selectWalletType(state.web3) === EWalletType.LIGHT) ||
```
rudolfix: if we still check LIGHT here we do not need to modify selector for selectBackupCodesVerified

app/modules/auth/selectors.ts
rudolfix: I we do not write value of this selector to backend, then ok to change it. just make sure it's OK in all places it is used

app/modules/kyc/selectors.ts
```diff
@@ -7,13 +7,16 @@ import {
...
 import { IKycState } from "./reducer";
 
-export const selectKycRequestStatus = (state: IKycState): TRequestStatus | undefined => {
-  const userKycType = selectKycRequestType(state);
+export const selectKycRequestStatus = (state: IAppState): TRequestStatus | undefined => {
+  const userKycType = selectKycRequestType(state.kyc);
   switch (userKycType) {
     case "business":
-      return state.businessRequestState!.status;
+      return state.kyc.businessRequestState!.status;
```
rudolfix: business and individual claims are handled in the same way on-chain.

app/modules/kyc/selectors.ts
```diff
@@ -7,13 +7,16 @@ import {
...
     case "individual":
-      return state.individualRequestState!.status;
+      return state.kyc.individualRequestState!.status === "Accepted" &&
```
rudolfix: Logic here looks a little bit wrong... It does not take into account rejected and ignored states. Let's sit down tomorrow and draw some state diagram

app/modules/auth/selectors.ts
Moejoe90: This selector for sure is not writing to the dataBase. 
any change to the backend should be explicit

app/modules/notifications/selectors.ts
Moejoe90: already removed

app/modules/kyc/selectors.ts
Moejoe90: It returns all states as they come from the backend so no need to extra states to be handled. 
The only special case I would say is `accepted`

app/modules/auth/selectors.ts
rudolfix: my point is that whenever we update user record we use this selector to get if codes were verified. now it will return `true` in cases it was returning `false`. so this will be written to db. am I correct? @Moejoe90

Moejoe90: This selector is only used for visual components so that fits it purpose perfectly. 
In order to write to our database for example `user db` the first this we do is get the data model as it then we explicitly change what needs to be changed. 
for example here 
https://github.com/Neufund/platform-frontend/blob/04bbb226842e8e186f4d1b281ee51622f7132f9c/app/modules/profile/sagas.ts#L47

PR #2074, opened 2018-11-26
Adds a second catch

Moejoe90:
should solve 
https://circleci.com/gh/Neufund/platform-frontend/21517#tests/containers/1

### DIFF
--- a/app/components/layouts/Header.tsx
+++ b/app/components/layouts/Header.tsx
@@ -3,9 +3,9 @@ import * as React from "react";
 import { Link } from "react-router-dom";
 import { Navbar } from "reactstrap";
 
+import { EUserType } from "../../lib/api/users/interfaces";
 import { actions } from "../../modules/actions";
 import { selectIsAuthorized, selectUserType } from "../../modules/auth/selectors";
-import { EUserType } from "../../lib/api/users/interfaces";
 import { appConnect } from "../../store";
 import { appRoutes } from "../appRoutes";
 import { Button, ButtonLink, EButtonLayout } from "../shared/buttons";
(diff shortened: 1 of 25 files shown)


app/components/layouts/Header.tsx
```diff
@@ -33,7 +35,7 @@ export const HeaderComponent: React.SFC<IStateProps & IDispatchProps> = props =>
       <Button
         layout={EButtonLayout.SECONDARY}
         theme="white"
-        onClick={props.logout}
+        onClick={() => props.logout(props.userType)}
```
desfero: Do we need to send userType to `logout` action?

app/test-e2e/eto/Campaining.spec.e2e.ts
```diff
@@ -41,54 +42,100 @@ describe("Eto campaining state", () => {
...
 
-        cy.get(tid("eto-bookbuilding-amount-backed")).should("contain", "€10 000");
-        cy.get(tid("eto-bookbuilding-investors-backed")).should("contain", "1");
+        cy.get(tid("eto-flow-start-bookbuilding")).awaitedClick();
 
         // give it a chance to settle before logging out
         cy.wait(5000);
+        logout();
 
+        cy.reload();
```
desfero: Are you sure we need to relaod? `reload` means that our app is not reactive and probably there is some bug in app state :(

app/components/layouts/Header.tsx
Moejoe90: yes it was a mistake to take it off 
because `AUTH_LOGOUT` clears the reducer
so yes we have to do it

app/test-e2e/eto/Campaining.spec.e2e.ts
Moejoe90: yes indeed this is not good but mind you this is related with `createUser` in e2e tests. Its really not a conventional way of doing things so there is something wrong there but no time to fix it now

Issue #2027, opened 2018-11-23
Widget |Embaded should open in the tab


SergejSoft:
If user clicks on the widget, the ETO Preview page should be open in the new tab. 
(href blan)

If clicked in Embaded vidgget we redirect users to NOT LOGED IN View  with Login and registre button.

If clicked on the Platform we redirect to Loggged in view. 

<img width="1440" alt="screen shot 2018-11-23 at 13 02 36" src="https://user-images.githubusercontent.com/2290474/48942562-39a66c00-ef20-11e8-8519-324940e626e5.png">

PR #2105, opened 2018-11-27
#2027 Fix widget routing

drakkein:


app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.tsx
```diff
@@ -164,9 +164,14 @@ function applyDiscountToPrice(price: number, discountFraction: number): number {
 
 function onEtoNavigationClick(
   navigate: () => void,
+  previewCode: string | undefined,
 ): (event: React.MouseEvent<HTMLDivElement>) => void {
   return function({ target, currentTarget }: React.MouseEvent<HTMLDivElement>): void {
     if (target === currentTarget) {
+      if (previewCode) {
```
huesforalice: to be honest, I would define window.open as a sideeffect which should be called from a saga. @Moejoe90 what is your take on this?

drakkein: ye, we should do this, also we should replace window.open in <Link> component. 👍

PR #2106, opened 2018-11-27
Removes false error logs from sentry

Moejoe90:


app/modules/access-wallet/sagas.ts
```diff
@@ -180,21 +182,3 @@ export function* accessWalletAndRunEffect(
 export function* connectWallet(): any {
   yield connectWalletAndRunEffect(call(() => {}));
 }
-
-/**
```
Moejoe90: This wasn't used anywhere in the app
@huesforalice should be fine to remove

app/modules/access-wallet/sagas.ts
```diff
@@ -23,6 +23,8 @@ import { EWalletType } from "../web3/types";
 import { mapSignMessageErrorToErrorMessage, MismatchedWalletAddressError } from "./errors";
 import { selectIsSigning } from "./reducer";
 
+export class MessageSignCancelledError extends Error {}
```
desfero: hm, do you think it's worth to move it to `errors.ts` file?

Moejoe90: Yes it is

PR #2112, opened 2018-11-28
At/#2087 translations

AlexeyTuboltsev:
change all string to translated messages, mostly in errors propagated from sagas

most important points:
1) we enumerate all errors (and other messages),
2) add a TMessage type that has the form {messageType:Enum, messageData: <any additional data you want to insert in the translated message>}
3) add function createMessage() - it's just a shorter way to create TMessage
4) TMessage is serializable and it goes to redux State. 
5) at rendering place - in the component -  we retrieve actual \<FormattedMessage/\>  with getMessageTranslation(message:TMessage)

there are still a couple of todos in sagas: we need to refactor errorModal to accept TMessage.


AlexeyTuboltsev:
visual regression is ok

AlexeyTuboltsev:
and there are still many sagas with strings, but we'll be able to fix them quickly when this pr is merged

PR #2123, opened 2018-11-29
Add widget e2e tests

drakkein:


app/test-e2e/eto/EtoWidget.spec.e2e.ts
```diff
@@ -0,0 +1,127 @@
+import { tid } from "../../../test/testUtils";
+import { appRoutes } from "../../components/appRoutes";
+import { e2eRoutes } from "../../components/e2eRoutes";
+import { insecureWithParams, withParams } from "../../utils/withParams";
+import { etoFixtureAddressByName } from "../utils";
+
+Cypress.Commands.add("iframe", { prevSubject: "element" }, $iframe => {
```
Moejoe90: Good job with this add please add it to `command.js` 
All custom function are added there 

If this is command is based on another snippet or a copy from somewhere else please include the link of the original implementation as a comment on top

app/test-e2e/eto/EtoWidget.spec.e2e.ts
```diff
@@ -0,0 +1,127 @@
...
+      );
+    });
+  });
+});
+
+describe("Eto widget page", () => {
+  it("Any", () => {
+    const ETO_ID = etoFixtureAddressByName("ETOInSetupState")!;
+
+    (window as any).open = cy.stub().as("windowOpen");
```
Moejoe90: This can be a utilty function

app/test-e2e/eto/EtoWidget.spec.e2e.ts
```diff
@@ -0,0 +1,127 @@
...
+          characterData: true,
+          subtree: true,
+        },
+      );
+    });
+  });
+});
+
+describe("Eto widget page", () => {
+  it("Any", () => {
```
Moejoe90: Any?

app/test-e2e/eto/EtoWidget.spec.e2e.ts
```diff
@@ -0,0 +1,127 @@
...
+    const ETO_ID = etoFixtureAddressByName("ETOInSetupState")!;
+
+    (window as any).open = cy.stub().as("windowOpen");
+
+    cy.visit(withParams(e2eRoutes.embededWidget, { etoId: ETO_ID }));
+
+    (cy.get("iframe") as any).iframe().within(($iframe: any) => {
```
Moejoe90: IMO we don't need ` as any` if the command is added to `command.js` file it will be chainable

app/test-e2e/utils/index.ts
```diff
@@ -240,3 +240,10 @@ export const etoFixtureAddressByName = (name: string) => {
   )! as EthereumAddress;
   return makeEthereumAddressChecksummed(address);
 };
+
+export const etoPreviewCodeByName = (name: string) => {
+  const address = Object.keys(ETO_FIXTURES).find(
+    a => ETO_FIXTURES[a].name === name,
+  )! as EthereumAddress;
```
Moejoe90: I would make an enum of all keys to get out of this problem as now i can get undefined but this is not blocking

app/components/AppRouter.tsx
```diff
@@ -109,6 +111,15 @@ export const AppRouter: React.SFC = () => (
       exact
     />
     <OnlyAuthorizedRoute path={appRoutes.kyc} investorComponent={Kyc} issuerComponent={Kyc} />
+
+    {/*Routes used only in E2E tests*/}
+    {(window as any).Cypress && (
```
Moejoe90: @desfero @trival  @AlexeyTuboltsev 
I'm thinking about this i can understand @drakkein approach and i'm sure it works but somehow i can't help but think maybe there is something better

app/test-e2e/eto/EtoWidget.spec.e2e.ts
desfero: 👍

app/test-e2e/utils/index.ts
drakkein: This piece of code is unused anyway, I'll remove it

app/test-e2e/eto/EtoWidget.spec.e2e.ts
```diff
@@ -2,97 +2,75 @@ import { tid } from "../../../test/testUtils";
...
-        .should("have.attr", "target", "_blank");
-
-      cy.wrap($iframe)
-        .find(tid("eto-overview-smart-contract-on-chain-button"))
-        .should("have.attr", "href")
-        .and("match", new RegExp(insecureWithParams(appRoutes.etoPublicView, { previewCode: "" })));
-
-      cy.wrap($iframe).find(tid("eto-overview-powered-by"));
-    });
+    cy.get("iframe")
```
desfero: @drakkein is it possible to simplify command to receive selector?
```
cy.iframe("iframe")...
```

app/test-e2e/utils/index.ts
```diff
@@ -241,9 +241,4 @@ export const etoFixtureAddressByName = (name: string) => {
   return makeEthereumAddressChecksummed(address);
 };
 
-export const etoPreviewCodeByName = (name: string) => {
-  const address = Object.keys(ETO_FIXTURES).find(
-    a => ETO_FIXTURES[a].name === name,
-  )! as EthereumAddress;
-  return makeEthereumAddressChecksummed(address);
-};
+export const stubWindow = (hookName: string) => ((window as any).open = cy.stub().as(hookName));
```
desfero: do we need to case as any `window` here? I see there is a `open` method in typings so it would be better to case right hand side

PR #2139, opened 2018-11-30
#2073_Remove double header from light wallet recovery

AlexeyTuboltsev:
remove header and explanation from RegisterLightWallet component if it's recovery.
fix css for footnote.

### DIFF
--- a/app/components/wallet-selector/light/RegisterLightWallet.tsx
+++ b/app/components/wallet-selector/light/RegisterLightWallet.tsx
@@ -120,20 +120,19 @@ export const RegisterWalletComponent: React.SFC<
   IDispatchProps & IStateProps & { restore: boolean }
 > = props => (
   <>
-    <h2
-      className={cn(styles.title, "text-center mb-4")}
-      data-test-id="modals.wallet-selector.register-restore-light-wallet.title"
-    >
-      {props.restore ? (
-        <FormattedMessage id="wallet-selector.neuwallet.restore-prompt" />
-      ) : (
-        <FormattedMessage id="wallet-selector.neuwallet.register-prompt" />
-      )}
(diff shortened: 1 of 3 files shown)


app/components/wallet-selector/light/RegisterLightWallet.tsx
```diff
@@ -120,20 +120,19 @@ export const RegisterWalletComponent: React.SFC<
...
-        <FormattedMessage id="wallet-selector.neuwallet.restore-prompt" />
-      ) : (
-        <FormattedMessage id="wallet-selector.neuwallet.register-prompt" />
-      )}
-    </h2>
-    <p className={styles.explanation}>
-      <FormattedMessage tagName="span" id="wallet-selector.neuwallet.explanation-1" />
-    </p>
-
+    {props.restore ? null : (
```
banciur:
```suggestion
    {props.restore && (
      <>
        <h2
          className={cn(styles.title, "text-center mb-4")}
          data-test-id="modals.wallet-selector.register-restore-light-wallet.title"
        >
          <FormattedMessage id="wallet-selector.neuwallet.register-prompt" />
        </h2>
        <p className={styles.explanation}>
          <FormattedMessage tagName="span" id="wallet-selector.neuwallet.explanation-1" />
        </p>
      </>
    )}
```

banciur: Isn't `&&` preferred way to render conditionally?

desfero: hm, but still we need to negate here :)

banciur: right :D

PR #2148, opened 2018-12-02
[WIP] Update bignumber

banciur:


desfero:
It's not possible to update `bignumber` at this point as there is the mismatch between `web3` `bignumber` and our own `bignumber` package

Issue #1513, opened 2018-10-09
Split production bundle


Moejoe90:
Based on
https://github.com/Neufund/platform-frontend/pull/1506 
we have a relatively large bundle size. 

The bundle should be split and optimized into smaller sizes.

PR #2150, opened 2018-12-03
#1513 Split production bundle to vendor, main and webpack runtime

desfero:
Enable long-term caching of vendor bundle.
Next step is to move assets to a separate bundle: #2151
Bundle:
image


app/lib/web3/CommitmentDeprecated.ts
```diff
@@ -1,419 +0,0 @@
-/* GENERATED BY TYPECHAIN VER. 0.1.4 */
```
desfero: @Moejoe90 is this one not needed anymore?

webpack/webpack.config.prod.js
```diff
@@ -20,10 +25,11 @@ module.exports = merge(configCommon, {
         },
       }),
       new OptimizeCSSAssetsPlugin(),
+      new webpack.HashedModuleIdsPlugin(),
     ],
   },
   output: {
-    filename: "[hash].[name].min.js",
```
desfero: this will enable long-term caching as if there are no changes in vendor `chunkhash` will output same hash

app/lib/web3/CommitmentDeprecated.ts
Moejoe90: :+1

desfero:
visual regression is ok, a couple of false positives only

Issue #2101, opened 2018-11-27
[ETO PUBLIC VIEW] graph labels are cut


glupiawymowka:
Hey, please show the full graph labels. Now they're cut 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/39586553-2cb5-4dec-b65f-b31f8b67cee6)

PR #2157, opened 2018-12-03
#2101 Fix tooltips in doughnut chart

AlexeyTuboltsev:
add custom tooltip function, tooltip styles, typings

### DIFF
--- a/app/components/shared/charts/ChartDoughnut.stories.tsx
+++ b/app/components/shared/charts/ChartDoughnut.stories.tsx
@@ -15,8 +15,4 @@ const data = {
 
 storiesOf("Charts/Doughnut", module)
   .addDecorator(story => <div style={{ width: "25em" }}>{story()}</div>)
-  .add("default", () => (
-    <div>
-      <ChartDoughnut data={data} />
-    </div>
-  ));
+  .add("default", () => <ChartDoughnut data={data} />);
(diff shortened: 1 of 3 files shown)


app/components/shared/charts/ChartDoughnut.tsx
```diff
@@ -41,7 +146,7 @@ export const ChartDoughnut: React.SFC<IProps> = ({ data, layout, className, defa
           legend={{ display: false }}
           height={100}
           width={100}
-          options={{ tooltips: { enabled: hasData(data) } }}
+          options={{ ...chartOptions }}
```
desfero: do we need to spread here? why not simply `chartOptions`?

app/components/shared/charts/ChartDoughnut.stories.tsx
```diff
@@ -13,4 +13,10 @@ const data = {
   ],
 };
 
-storiesOf("Charts/Doughnut", module).add("default", () => <ChartDoughnut data={data} />);
+storiesOf("Charts/Doughnut", module)
+  .addDecorator(story => <div style={{ width: "25em" }}>{story()}</div>)
+  .add("default", () => (
+    <div>
```
desfero: do we need `div` here?

AlexeyTuboltsev:
visual regression is ok

PR #2180, opened 2018-12-05
Add wallet type to sentry

desfero:


app/modules/auth/sagas.ts
```diff
@@ -171,7 +171,7 @@ function* setUser({ logger }: TGlobalDependencies, action: TAction): Iterator<an
 
   const user = action.payload.user;
 
-  logger.setUser({ id: user.userId, type: user.type });
```
desfero: can be useful to debug information related to wallets

Issue #2042, opened 2018-11-24
[IMPORTANT USABILITY] no way to know when ETO ends


rudolfix:
right now ETO is about to END but it's impossible to know when. I propose to add a counter to the widget - we even have design for it somewhere. it will count
- time until public in whitelist if investment widget visible
- time until signing in public
![image](https://user-images.githubusercontent.com/17202864/48967201-2f967300-efdd-11e8-852b-3037db40eb32.png)

falinebambi:
Count down until end of ETO goes under the 'INVEST NOW' button see design 

<img width="355" alt="screen shot 2018-11-28 at 11 52 47" src="https://user-images.githubusercontent.com/17235402/49147149-2884a300-f304-11e8-85e9-4c2d731daa57.png">

drakkein:
@falinebambi I think yours description and this days left state are enough.  :)

PR #2193, opened 2018-12-06
#2042 Add eto end timer

drakkein:


app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import * as styles from "./EndTimeWidget.module.scss";
+
+interface IExternalProps {
+  endTime: Date | undefined;
+}
+
+const EndTimeWidget: React.SFC<IExternalProps> = ({ endTime }) => {
+  const endMoment = moment(endTime);
+  const endInDays = Math.round(endMoment.diff(moment(), "days", true));
```
AlexeyTuboltsev: maybe get rid of so many moment() function calls:
const endInDays = Math.round(endInHours/24) ?

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
```diff
@@ -0,0 +1,33 @@
...
+import * as React from "react";
+import { FormattedMessage } from "react-intl-phraseapp";
+
+import * as styles from "./EndTimeWidget.module.scss";
+
+interface IExternalProps {
+  endTime: Date | undefined;
+}
+
+const EndTimeWidget: React.SFC<IExternalProps> = ({ endTime }) => {
```
AlexeyTuboltsev: maybe implement a timer here? The way it is now it won't change even if the page stays open for hours

drakkein: The time left is recalculated every time when wallet state is updated. As I remember it happens every 30 seconds. 
image

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
drakkein: This is good idea! Thanks! 👍

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
AlexeyTuboltsev: You're right.  But we def. shouldn't rely on this behavior because it's more of a bug and it will definitely go away when we introduce caching or smth else to optimize the rendering of the app. Then the time will not update anymore. It won't update even now if you have a request error or backend failure.

drakkein: Based on knowledge from @desfero there is mechanism which updates contracts every 1 minute for public and whitelisted ETOs. If contract is updated the remaining time to end is recalculated. I checked it, and it's true. So I guess this is enough. But if you still want me to add some internal update loop to this component, I will no longer resist. 😺

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
```diff
@@ -0,0 +1,33 @@
+import * as moment from "moment";
+import * as React from "react";
+import { FormattedMessage } from "react-intl-phraseapp";
+
+import * as styles from "./EndTimeWidget.module.scss";
+
+interface IExternalProps {
+  endTime: Date | undefined;
```
desfero: hm, `endDate`?

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
```diff
@@ -0,0 +1,33 @@
+import * as moment from "moment";
```
desfero: maybe name the file `EndDateWidget` as it's not only the time

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
desfero: So on my side, we should add counter here (basically reuse what `Counter.tsx` already do by extracting common code to Render props component)

```
<Counter>
{({ daysLeft, hoursLeft }) => (
  <span className={styles.endTime}>
      {daysLeft === 0 ? (
        <FormattedMessage
          id="shared-component.eto-overview.invest.ends-in-hours"
          values={{ endInHours: hoursLeft < 1 ? 0 : Math.round(hoursLeft) }}
        />
      ) : (
        <FormattedMessage
          id="shared-component.eto-overview.invest.ends-in-days"
          values={{ daysLeft }}
        />
      )}
    </span>
  );
)}
</Counter>
```

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.spec.tsx
```diff
@@ -0,0 +1,129 @@
+import { expect } from "chai";
+import { shallow } from "enzyme";
+import * as moment from "moment";
+import * as React from "react";
+import * as sinon from "sinon";
+
+import { FormattedMessage } from "react-intl-phraseapp";
```
desfero: sort imports :)

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
desfero: btw, what if endTime is null? we probably don't even want to render this component so props should be as `endTime: Date` only

app/components/eto/overview/EtoOverviewStatus/EtoOverviewStatus.stories.tsx
```diff
@@ -26,6 +28,19 @@ const eto = {
   contract: {
     timedState: EETOStateOnChain.Whitelist,
     totalInvestment: { totalInvestors: new BigNumber("123"), totalTokensInt: new BigNumber("234") },
+    startOfStates: [
+      undefined,
+      new Date(),
```
desfero: use then always `moment` for consistency if use already use it below

app/components/eto/overview/EtoOverviewStatus/InvestmentWidget/InvestmentWidget.tsx
```diff
@@ -111,6 +122,16 @@ const InvestmentWidget = compose<TInvestWidgetProps, IInvestmentWidgetProps>(
       startInvestmentFlow: () => dispatch(actions.investmentFlow.startInvestment(props.eto.etoId)),
     }),
   }),
+  withProps<IInvestmentWidgetWithProps, IInvestmentWidgetProps>(({ eto }) => {
+    const nextState =
```
desfero: we can probably use here `selectEtoOnChainNextStateStartDate` as it's almost the same logic under the hood

app/components/eto/overview/EtoOverviewStatus/EndTimeWidget.tsx
desfero: btw @AlexeyTuboltsev @drakkein we should also check react-intl `FormattedPlural` component, maybe this component will even further simplify code

AlexeyTuboltsev: oh i'm not sure. It's selfupdating, it's too much magic for my taste. If it's an intl library, let it stay an intl lib, e.g, deliver static strings according to locale.

PR #2198, opened 2018-12-07
At/#2103 swiper refactor

AlexeyTuboltsev:
- refactor swiper to a lightweight component, remove ugly old huge library
- add swiper states to storybook

this PR also adds a very nice viewport addon to storybook and adds a couple of custom breakpoints to play with


AlexeyTuboltsev:
visual regression is ok. Screenshots of slider are wrong cause it's only slider here, without the whole layout. This should be fixed too but in a separate ticket. I checked all states of slider in a real layout, it's all fine

Issue #1035, opened 2018-07-12
presenting manual verification to failed IDnow


shanile:
We need to add a possibility for users to choose verificaiton methon once he droped out of first verification attempt. 

Back end endpoints are already in place. 

Now: if users started with IDnow they can not switch to Manual Verification. 
We need to change this by giving the option. 
See mockup. 
Two links. One back to IdNow, second to Manual Verification.
![verif](https://user-images.githubusercontent.com/2290474/42823697-781fbbaa-89de-11e8-814f-ced76e4286a1.png)

rudolfix:
BUMP! let's do it next week. Proposed solution is simple
when we are in IDNOW we are always in outsourced state. Now when we are in this state it should be able to go back DRAFT from the user interface.
- if we do not have reject reason from IDNOW we can let user "CANCEL IDNOW" that will switch state to DRAFT and move user to KYC form where he can edit data and choose manual/IDNOW again
- if we have reject reason we should show a message "YOUR IDENTIFICATION BY IDNOW FAILED" and we proceed same as above (maybe we should have different text on the button like TRY AGAIN or as Shani says we can put two buttons that take user to different forms)

Moejoe90:
This can be quickly implemented, once the endpoints are provided by backend

shanile:
Please check if this was done

Moejoe90:
@huesforalice 
Can you provide the end points, please?

glupiawymowka:
this is implemented but needs logics & wording fix (see #2408 )

PR #2200, opened 2018-12-07
#1035 Adds cancel IDNow feature

Moejoe90:
Related #1035 

Canceling ID now works only the first time. Once ID is cancelled there the backend doesn't store the user as outsourced anymore. 

Check failed e2e test for exact details
```
KYCFlowIndividualIDnow.spec.e2e.ts
```
Once this is fixed on backend test should pass


Moejoe90:
@huesforalice please check up

Issue #1998, opened 2018-11-22
Success tick rendering size is broken


drakkein:
We have some problems with correct SuccessTick rendering size

![image](https://user-images.githubusercontent.com/10601291/48904292-2fbc3480-ee5e-11e8-8cc4-47424c5b6730.png)

PR #2219, opened 2018-12-09
#1998 Fix success tick box size

drakkein:
Looks like now it's correct. Also position is correctly set. But without margin tick is slightly bigger because of pseudoelements size.  
image


app/components/shared/SuccessTick.module.scss
```diff
@@ -1,11 +1,12 @@
...
 
 .sa {
-  @include rect(80px);
+  @include rect(90px);
 
   background-color: #fff;
   transform: scale(.54);
-  margin-top: -20px;
+  transform-origin: center;
+
```
AlexeyTuboltsev: 👍

Issue #2232, opened 2018-12-11
Fix deprecation warning during `yarn start`


desfero:
```DEPRECATION WARNING on line 2, column 8 of /Users/work/WebstormProjects/platform-frontend/app/styles/global.scss:
Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.
Use a custom importer to maintain this behaviour. Check your implementations documentation on how to create a custom importer```

PR #2234, opened 2018-12-11
#2232 Fix node-sass deprecation warning

desfero:


app/styles/global.scss
```diff
@@ -1,5 +1,4 @@
 @import "~react-id-swiper/src/styles/scss/swiper";
-@import "~react-datetime/css/react-datetime";
```
desfero: moved for consistency to the same place where component is used

package.json
```diff
@@ -143,7 +143,7 @@
     "mini-css-extract-plugin": "^0.4.1",
     "mocha": "^4.1.0",
     "mockdate": "^2.0.2",
-    "node-sass": "^4.9.3",
+    "node-sass": "^4.11.0",
```
desfero: a simple update was enough to fix warnings. 

From node-sass changelog
- Removes noisey deprecation warning for @import'ing .css files

desfero:
visual regression ok right now

package.json
Moejoe90: :+1:

Issue #2225, opened 2018-12-10
Second Round ETO Teaser Cards 


falinebambi:
![screen shot 2018-12-10 at 15 10 49](https://user-images.githubusercontent.com/17235402/49737703-d17cb780-fc8d-11e8-84c4-e20869cbef37.png)

ZEPLIN links with full res assets here: 
https://zpl.io/2ylLWwG
zpl://screen?sid=5c0e6dc9ace637b6e7800e71&pid=5a8a92c89c1a166a6a6e8f37

GIF for Curf Technology on hover can be found here: https://drive.google.com/drive/folders/1Q_4hNykPwBanlXRouNXR5SpPcMKiEvlW 

Final cards in google drive here: 
AirProfile Teaser Card Designs https://drive.google.com/drive/folders/1pOVq-BaSGXZNejnBIELk3S6GUABxOJbu
CURF Teaser Card Designs https://drive.google.com/drive/folders/1Q_4hNykPwBanlXRouNXR5SpPcMKiEvlW
Agora Teaser Card Designs https://drive.google.com/drive/folders/15KcekYs-RmCQpaGqqVNnU1GXhgDohPTk
NGRAVE Teaser Card Designs https://drive.google.com/drive/folders/1OXZ3zx1_6RcqcI-EYY8VBcatNwQUS6EF

PR #2238, opened 2018-12-11
#2225 Add second round of cards

drakkein:
### Waiting for proper images


app/components/landing/Landing.tsx
```diff
@@ -39,17 +39,20 @@ export const LandingComponent: React.SFC = () => (
...
-              className={styles.equityTokenCol}
-              id={`eto-card-${index}`}
-              key={index}
-            >
-              <EtoCard {...e} />
-            </Col>
-          ))}
+          {etoCompaniesCards.map(
+            (e, index) =>
+              !e.data.hidden && (
```
desfero: maybe use `filter` method here is more readable?

app/components/shared/EtoOfferingCard.tsx
```diff
@@ -91,6 +92,7 @@ export class EtoOfferingCardComponent extends React.Component<
               props.teaser && styles.teaser,
               props.isMobile && styles.mobile,
             )}
+            rel="noopener"
```
desfero: +1

app/data/etoCompanies.ts
```diff
@@ -1,11 +1,22 @@
 import { IEtoOfferingProps } from "../components/shared/EtoOfferingCard";
 
+import * as agora from "../assets/img/eto_offers/agora.jpg";
```
desfero: we need to think about lazy loading all images as this is really huge overhead for bundle size

PR #2241, opened 2018-12-11
[Hotfix] Broken lints and panel on eto view

drakkein:
Fixing broken panel styles and lint error, after component rename.
Summoning @trival because I wasn't able to find reason for setting this height, so I'm not sure if this is not breaking anything else. 
image


app/components/shared/Panel.module.scss
```diff
@@ -4,7 +4,6 @@
   display: flex;
   flex-direction: column;
   padding: px-to-rem(30px) px-to-rem(25px) px-to-rem(25px);
-  height: 100%;
```
trival: please keep or refactor the height. It is necessary for the flex layout in the eto issuer dashboard, to keep all box height aligned

PR #2246, opened 2018-12-12
#2209 add DocumentLayout to storybook

AlexeyTuboltsev:
add large state chunks to fixtures


test/fixtures.ts
```diff
@@ -780,3 +781,169 @@ export const testEto = {
     },
   },
 };
+
+export const etoFilesData = {
+  allTemplates: {
```
trival: could we add Typing here?

app/components/documents/Documents.tsx
```diff
@@ -94,7 +95,7 @@ const DocumentsLayout: React.SFC<IProps> = ({
   const { allTemplates, stateInfo } = etoFilesData;
   const generalUploadables = stateInfo ? stateInfo.uploadableDocuments : [];
   return (
-    <LayoutAuthorized>
+    <>
```
desfero: hm, do we then need `Fragment`? I think it's possible to remove fragment and reduce jsx nesting

app/components/documents/Documents.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+  etoDocuments: etoDocuments,
+  documentTitles: documentTitles,
+  isRetailEto: false,
+  generateTemplate: () => {},
+  downloadDocumentByType: () => {},
+};
+
+storiesOf("ETO/Documents", module).add("default", () => {
+  return (
+    <Container>
```
desfero: btw, do we need `Container`? Does the screenshot layout look ugly without? I'm asking because it's better to keep stories without any additional wrappers in most cases

app/lib/api/eto/EtoFileApi.interfaces.ts
```diff
@@ -34,13 +34,19 @@ export enum EEtoDocumentType {
 
 export type TEtoFormType = "document" | "template";
 
+export enum TEtoDocumentLanguage {
```
desfero: typo: `EEtoDocumentLanguage`

PR #2247, opened 2018-12-12
#2225 Add additional teaser fixes

drakkein:
Additional fixes for gif's based teaser cards


app/data/etoCompanies.ts
```diff
@@ -453,6 +453,8 @@ export const etoCompaniesCards: Array<EtoData> = [
         alt: "",
         height: 270,
         width: 530,
+        preserveOriginalRatio: true,
+        forceBg: "black",
```
AlexeyTuboltsev: Doesn't it make sense to use our color variables here?
Pawel defined a bunch of them recently:
platform-frontend/app/styles/variables/colors.scss

drakkein: It's temporray, I will do refactor of landing page during next week. Mainly I will get rid of gifs, so there will be no more need fo this fix.

PR #2250, opened 2018-12-12
Update webpack and related dependencies (eg. loaders)

desfero:


webpack/webpack.config.prod.js
```diff
@@ -1,4 +1,4 @@
-const UglifyJsPlugin = require("uglifyjs-webpack-plugin");
+const TerserPlugin = require("terser-webpack-plugin");
```
desfero: `uglify-es` is abandoned and webpack now recommend to use terser

PR #2256, opened 2018-12-13
Update typescript - new features in typescript, requires approval by all devs

desfero:
Please read Typescript changelog for typescript 3.2 and 3.1 to know about new features that we may use in future: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript


app/components/eto/overview/EtoOverviewStatus/CampaigningWidget/CampaigningActivatedInvestorApprovedWidgetLayout.tsx
```diff
@@ -80,7 +80,7 @@ const CampaigningActivatedInvestorApprovedWidgetLayout: React.SFC<
           </div>
         </div>
       ) : (
-        <Formik
+        <Formik<{ amount: number | "" }>
```
AlexeyTuboltsev: this doesn't look good: why would we allow invalid properties? If it's a number, default value should be 0, undefined or null (we always use undefined). But not an empty string. This might be harmless in this particular case, but it caused us problems in many other places and it's invalid. If this input needs an empty string as starting val, let's convert default value to empty string, and convert empty value to undefined/0/null back inside the input component, so that the outside world never sees an invalid value.

app/components/settings/icbm-wallet-widget/CheckYourICBMWalletWidget.tsx
```diff
@@ -62,7 +62,10 @@ export const CheckYourICBMWalletWidgetComponent: React.SFC<IDispatchProps> = ({
       <p>
         <FormattedMessage id="check-your-icbm-wallet-widget.notice" />
       </p>
-      <Formik initialValues={{ address: "" }} onSubmit={values => loadICBMWallet(values.address)}>
+      <Formik<{ address: string }>
+        initialValues={{ address: "" }}
```
AlexeyTuboltsev: the same as CampaigningActivatedInvestorApprovedWidgetLayout.tsx. 
Empty string is an invalid value. Let's give it an undefined, formik is ok with it.

PR #2262, opened 2018-12-14
Import only english mnemonics from bitcore-mnemonic

desfero:
Reduces gzipped vendor bundle size for 75kb

Before:
image

After:
image


app/utils/englishMnemonics.ts
```diff
@@ -0,0 +1,4 @@
+// We should not import whole `bitcore-mnemonic` as the package size is enormous, we need only english mnemonics.
+import * as englishMnemonics from "bitcore-mnemonic/lib/words/english";
```
Moejoe90: Why don't we just copy the dictionary file and get rid of the library completely.

Bitcore mnemonic is still in the lightwallet btw but that's another discussion

PR #2265, opened 2018-12-14
Fix eto start date e2e test

desfero:


app/components/modals/tx-sender/eto-flow/SetDateSummary.tsx
```diff
@@ -58,7 +58,7 @@ const SetEtoDateSummaryComponent: React.SFC<IProps> = ({
         <FormattedHTMLMessage
           tagName="p"
           id="eto.settings.eto-start-date-summary.dates-description"
-          values={{ timeToChange: changableTill.fromNow(true) }}
+          values={{ timeToChange: changableTill.from(moment().startOf("day"), true) }}
```
desfero: @trival it happened because `fromNow` method compares also time and in some cases it's was either 19 days or 20 days

Issue #2189, opened 2018-12-06
Refactor Documents Section


Moejoe90:
Documents section needs some refactoring. 
We are adding more features and it's becoming a bit harder to maintain. 

1 - All components Should be Split into separate components and files 
2 - There should be a clear separation between Document as a Type and Template
3 - There is some extra code that can get taken off

PR #2276, opened 2018-12-18
#2189 refactor document section

AlexeyTuboltsev:
refactor Documents.tsx, Document.tsx, styles, add typings


app/components/shared/Document.module.scss
```diff
@@ -83,10 +94,25 @@
   background-color: transparent;
   border-radius: px-to-rem(5px);
   border: 0;
+}
+
+.sub-title-download {
+  width: 100%;
+  color: $color-black;
+  display: flex;
```
desfero: hm, i do think more appropriate here is to use `display: block` and then we can remove `justify-content` and `text-align`

app/components/shared/Document.module.scss
```diff
@@ -83,10 +94,25 @@
...
+  justify-content: center;
+  text-align: center;
+  font-size: px-to-rem(8px);
+  padding: 0;
+  margin-bottom: 0;
+  background-color: transparent;
+  border: 0;
+  cursor: pointer;
```
desfero: all `button` are with `cursor: pointer`, see `button.scss`

app/components/shared/Document.module.scss
```diff
@@ -83,10 +94,25 @@
...
+
+.sub-title-download {
+  width: 100%;
+  color: $color-black;
+  display: flex;
+  justify-content: center;
+  text-align: center;
+  font-size: px-to-rem(8px);
+  padding: 0;
+  margin-bottom: 0;
```
desfero: do we need to reset margin?

app/components/shared/Document.module.scss
```diff
@@ -83,10 +94,25 @@
...
+  background-color: transparent;
+  border: 0;
+  cursor: pointer;
 
   &:active,
   &:hover,
   &:focus {
-    box-shadow: 0 0 px-to-rem(5px) px-to-rem(1px) $gray10;
+    text-decoration: underline;
```
desfero: +1

app/components/documents/Documents.module.scss
```diff
@@ -1,32 +1,69 @@
...
+
+  .document-section {
+    grid-template: auto 1fr / repeat(3, 1fr);
+  }
+}
+
+@media(min-width: 900px) {
+  .layout {
+    grid-template: auto 1fr 1fr / 2fr 1fr;
+    padding: 0 15px;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: that's why i left it here. There is a 15px padding on .container and 15 px padding on .layout. I only refactor things up to .layout level, I don't want to touch the container now, cause it's used everywhere and  I'm afraid to break something. So it's kinda backwards compatibility for now

app/components/documents/Documents.tsx
```diff
@@ -93,93 +134,69 @@ const DocumentsLayout: React.SFC<IProps> = ({
   const { allTemplates, stateInfo } = etoFilesData;
   const generalUploadables = stateInfo ? stateInfo.uploadableDocuments : [];
   return (
-    <Row data-test-id="eto-documents">
-      <EtoFileIpfsModal />
-      <Col xs={12} lg={8}>
-        <SectionHeader className="my-4">
+    <>
+      <div data-test-id="eto-documents" className={styles.layout}>
+        <SectionHeader className={cn(styles.header)}>
```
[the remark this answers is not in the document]
AlexeyTuboltsev: well, why? 
It's just a layout for the page, to place other modules in it, with respect to page width. It won't be reused anywhere else.
i'd rather refactor the Profile to use grid, cause now it uses Cols/Rows

app/components/shared/SingleColDocumentWidget.tsx
```diff
@@ -37,12 +35,12 @@ const SingleColDocumentsLayout: React.SFC<IProps> = ({
...
   return (
-    <Panel className={className}>
-      <Col className={styles.groupName}>{title}</Col>
+    <div className={className}>
+      <h3 className={styles.groupName}>{title}</h3>
       <div className={styles.group}>
         {documents.map(({ ipfsHash, mimeType, documentType }, i) => {
           return (
-            <Col xs={12} className={styles.document} key={i}>
+            <div className={styles.document} key={i}>
```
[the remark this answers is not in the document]
AlexeyTuboltsev: 👍 👍 👍

Issue #2220, opened 2018-12-10
Refactor Campaign e2e test


SergejSoft:
Add dynamic assertion.

PR #2280, opened 2018-12-18
#2220 Refactor campaigning e2e tests

drakkein:


app/test-e2e/eto/Campaigning.spec.e2e.ts
```diff
@@ -101,14 +113,22 @@ describe("Eto campaining state", () => {
...
 
             confirmAccessModal();
 
-            cy.get(tid("eto-bookbuilding-amount-backed"));
-            cy.get(tid("eto-bookbuilding-remaining-slots"));
+            cy.get(tid("campaigning-your-commitment")).then($element => {
+              cy.wrap($element).contains(`€${formatThousands(CHANGED_AMOUNT)}`);
+            });
+            cy.get(tid("eto-bookbuilding-remaining-slots")).then($element => {
+              assert.isBelow(
```
Moejoe90: why not calculate the exact amount instead of asserting that the number is getting lower. 
Maybe its not easy to do

PR #2287, opened 2019-01-06
[#2205] Add dashboard stories

drakkein:


app/components/dashboard/Dashboard.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+import * as React from "react";
+
+import { testEto } from "../../../test/fixtures";
+import { withStore } from "../../utils/storeDecorator";
+import { Dashboard } from "./Dashboard";
+
+storiesOf("Dashboard", module)
+  .addDecorator(
+    withStore({
+      web3: {
```
desfero: @Moejoe90 do you think it's ok to add such components to storybook? For me, this looks like a nightmare to maintain as it's too couple with store shape. To add more maintenance complexity, store is not used anywhere in `Dashboard` but in child components those making catching errors here more difficult.

AlexeyTuboltsev: @desfero, @Moejoe90 : I actually had a couple of them in tickets in the tech run before Xmas. I didn't do them, cause it looks to me to be a wrong way to tests this stuff. There should be some way to test the general layout, but maybe with some dummy placeholders instead of real widgets. But the layout components and all styling should be build in the right way for this. Like this:

styles.dummy { background: gray, width: 100%, height:  100%}
container classes basically only say, how much grid column span there is, plus some padding.

`const Dummy = ({containerClass}) => <div cn(styles.dummy, containerClass) />` 

real world:
`<LayoutComp>
<RealWidget1 containerClass={containerClass1} />
<RealWidget2 containerClass={containerClass2}/>
</LayoutComp>`

test: 
`<LayoutComp>
  <Dummy containerClass={containerClass1}/>
  <Dummy containerClass={containerClass1}/>
</LayoutComp>`

Moejoe90: @AlexeyTuboltsev I don't agree with adding dummy components. 

@desfero I understand that adding the store like this can be too much coupled but on the other hand once we have full views we can easily check visual regression on full views including layouts. this is very powerful when it comes to refactoring components and checking changes not only on a component level but a full view 

With that said the least worse solution would be having a mocked store object in one place and wrapping any component that needs store around it

drakkein: Ok then, looks like it's ready to merge

Issue #2197, opened 2018-12-07
Terms of Use. Updated


SergejSoft:
Problem: Since we updated terms of use users will see it again. 
<img width="1234" alt="screen shot 2018-12-07 at 11 26 36" src="https://user-images.githubusercontent.com/2290474/49642323-00d3c000-fa13-11e8-8a6a-4b349be86bbc.png">

The mission: if users have already read T&C once, to show the same modal with the different text:

"Please read our updated Terms of Use" and pls. make the word updated <strong>

glupiawymowka:
*tested on .net*
It's correct (title with "updated" for old users, without it for new ones), but why is the document's name [object Object] now? 
Please fix it 

![image](https://user-images.githubusercontent.com/18315589/51542100-bbe8d900-1e5a-11e9-8bb1-273396f1eeb3.png)

PR #2291, opened 2019-01-08
#2197 Wording of updated ToS modal

AlexeyTuboltsev:
change wording of ToS modal if ToS are updated


app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -41,11 +43,15 @@ export class AcceptTosModalInner extends React.Component<IStateProps & IDispatch
   };
 
   render(): React.ReactNode {
```
drakkein: I know that this is not part of your recent work, but can we change this class component to functional component with recomposes `withState` and `withHandlers`?
Summoning @desfero and @Moejoe90 to hear their opinion.

Rest is good 👍

AlexeyTuboltsev:
Kirstyn said no need to change buttons, so it's good to go, i guess

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
AlexeyTuboltsev: yep, but  i'd do it later, need to get it go

PR #2293, opened 2019-01-09
Fix withdraw on safari

desfero:


app/components/modals/tx-sender/withdraw-flow/Withdraw.tsx
```diff
@@ -87,7 +85,7 @@ const WithdrawComponent: React.SFC<TProps & IIntlProps> = ({
   validationState,
   intl,
 }) => (
-  <div>
+  <section>
```
desfero: for correct semantic when below we have any kind of `h*` tag we should use `section` on wrapper

app/components/modals/tx-sender/withdraw-flow/Withdraw.tsx
```diff
@@ -100,64 +98,61 @@ const WithdrawComponent: React.SFC<TProps & IIntlProps> = ({
...
-                      onValidateHandler(values.value, e.target.value);
-                    }}
-                  />
-                </Col>
-              </Row>
-              <Row>
-                <Col xs={12} className="mb-3">
-                  <FormField
-                    name="value"
-                    type="number"
```
desfero: ohh, sorry removing `return ...` caused a lot of lines to be marked as changed. Here is a real change, `type="number"` is not consistent in different browsers so it's better to use `type="text"` and provide formik validation

Moejoe90: ah so this was the bug

PR #2303, opened 2019-01-10
[WIP] New layout for private key view

desfero:


app/assets/img/inline_icons/add_file.svg
```diff
@@ -1,9 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <svg width="73px" height="78px" viewBox="0 0 73 78" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
-    <!-- Generator: Sketch 48.2 (47327) - http://www.bohemiancoding.com/sketch -->
-    <title>file</title>
```
desfero: Generally, `title` should not be provided in almost all cases. That's because screen reader check for `title` on `svg` and often it doesn't make sense to read `file` for e.g.

app/components/shared/buttons/ButtonGroup.module.scss
```diff
@@ -0,0 +1,8 @@
+.button-group {
+  margin: -10px -15px;
+
+  > * {
```
desfero: anyone know how can I target in CSS modules, specific class? As here it would be better to get class name of `Button` but it's probably not possible with css-modules AFAIK

app/components/shared/InlineIcon.tsx
```diff
@@ -8,21 +11,24 @@ interface IProps {
   height?: string;
   className?: string;
   onClick?: () => void;
```
desfero: btw, we should remove `onClick` here as it's attached to `span` which is not correct and wrap `InlineIcon` in `ButtonIcon` where this props is used

app/components/shared/InlineIcon.tsx
```diff
@@ -8,21 +11,24 @@ interface IProps {
...
+  <>
     <span
       className={className}
       style={{ width, height }}
       dangerouslySetInnerHTML={{ __html: svgIcon }}
       {...props}
     />
-  );
-};
+    {alt && <span className="sr-only">{alt}</span>}
```
desfero: This change is connected with cleaning `title` from svg's. If you need to provide title (e.g. for clipboard button you can now use `alt` attribute

app/components/shared/SectionHeader.tsx
```diff
@@ -7,7 +7,6 @@ import * as styles from "./SectionHeader.module.scss";
 
 export enum ESectionHeaderSize {
   SMALL = styles.small,
-  MEDIUM = styles.medium,
```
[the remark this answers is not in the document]
desfero: it's a default size

app/test-e2e/utils/index.ts
```diff
@@ -144,6 +152,10 @@ export const typeLightwalletRecoveryPhrase = (words: string[]) => {
   cy.get(tid("btn-send")).awaitedClick();
 };
 
+export const assertLockedAccessModal = () => {
+  cy.get(tid("access-light-wallet-locked")).should("exist");
+};
+
 export const confirmAccessModal = (password: string = DEFAULT_PASSWORD) => {
   cy.get(tid("access-light-wallet-password-input")).type(password);
   cy.get(tid("access-light-wallet-confirm")).awaitedClick(1500);
```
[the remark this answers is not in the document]
desfero: yeap, agree there are a couple of different domains mixed so it's easy to extract as a separate files

PR #2306, opened 2019-01-12
E2E tests for critical error page

desfero:


app/config/.env.e2e
```diff
@@ -10,5 +10,3 @@ NF_WITHDRAW_ENABLED=1
 NF_CHECK_LOCKED_WALLET_WIDGET_ENABLED=1
 NF_EQUITY_TOKEN_OFFERINGS_VISIBLE=1
 NF_ETO_LOAD_REAL_DATA=1
-
-NF_SENTRY_DSN=https://1a01753f602c4bbcad39de77a992fc8d@sdsn.neufund.io/7
```
desfero: it's not possible to mock sentry for `CriticalError` page test as under the hood it uses Beacon API to send errors, so need to disable logging for e2e tests.

PR #2310, opened 2019-01-14
bundlewatcher

AlexeyTuboltsev:
add bundlewatcher config, ci config, setup script
fix image added to bundle twice


.circleci/config.yml
```diff
@@ -138,19 +156,22 @@ workflows:
...
+      - test:
+          requires:
+            - lint
+      - test-e2e:
+          requires:
+            - lint
+      - danger-ci:
+          requires:
+            - lint
+      - bundlewatcher:
```
desfero: @AlexeyTuboltsev do we need to watch bundle on nightly build?

app/components/shared/HiResImage.tsx
```diff
@@ -20,19 +20,14 @@ export const HiResImage: React.SFC<IHiResImageProps> = ({
   }
 
   try {
-    // @note: thanks to wonders of static analysis + webpack all images matching these paths will be part of the bundle
-    const image = require("!file-loader?publicPath=/!../../assets/img/" + partialPath + ".png");
-    const image2x = require("!file-loader?publicPath=/!../../assets/img/" +
-      partialPath +
-      "@2x.png");
+    const image = require("../../assets/img/" + partialPath + ".png");
```
desfero: are this one really inside `main` bundle? Or they are loaded on demand

webpack/webpack.config.dev.js
```diff
@@ -127,7 +127,7 @@ module.exports = merge(configCommon, {
             exclude: paths.inlineIcons,
             options: {
               limit: 5000,
-              name: "images/[name].[hash:8].[ext]",
+              name: "images/[hash].[ext]",
```
desfero: 👍 for name, but is there any reason to make hash longer than 8 characters?

webpack/webpack.config.prod.js
```diff
@@ -125,11 +125,11 @@ module.exports = merge(configCommon, {
           },
           {
             test: /\.(jpg|png|svg|gif)$/,
-            loader: "url-loader",
+            loader: "file-loader",
```
desfero: why do we want to put all images in the mail bundle? It will increase bundle a lot `url-loader` is a lot better here as it will put images that are less than `limit` inside main bundle but all other heavy images will be loaded by url

webpack/webpack.config.dev.js
desfero: same to `webpack.prod.config`

.circleci/config.yml
Moejoe90: This is a good point we don't need it actually in the nightly build

webpack/webpack.config.prod.js
Moejoe90: :+1:

app/components/shared/HiResImage.tsx
AlexeyTuboltsev: they all go to /images/ folder now

PR #2315, opened 2019-01-15
At/#2143/translations

AlexeyTuboltsev:
cherry-picked from at/#2143_translations

refactor all strings to stringifiable translation objects, 
pass around only the translation object
get the actual translation (FormattedMessage) only at rendering site


app/components/modals/walletAccess/AccessWalletModal.tsx
```diff
@@ -73,18 +73,22 @@ export const AccessWalletContainerComponent: React.SFC<IStateProps & IDispatchPr
         </div>
       </div>
     )}
-    {errorMessage && (
-      <p className={cn("mt-3", styles.error)}>{getMessageTranslation(errorMessage)}</p>
-    )}
+    {errorMessage && <p className={cn("mt-3", styles.error)}>{errorMessage}</p>}
```
Moejoe90: We dont need 'getMessageTraslation' here?

AlexeyTuboltsev: all the messages for this component get resolved on the lines 82-92

PR #2316, opened 2019-01-15
Fix notification DOM error

desfero:
Fixes:
image


app/components/shared/notification-widget/Notification.tsx
```diff
@@ -1,88 +1,75 @@
...
+          onClick={onClick}
+        >
+          <NotificationContent text={text} type={type} />
+        </Button>
+      ) : (
+        <NotificationContent text={text} type={type} />
+      )}
+
+      {onClick && (
+        <ButtonClose className={styles.close} data-test-id="notification-close" onClick={onClick} />
```
desfero: Still, the implementation here is not correct as `ButtonClose` should close notification without any external action dispatched (like go to settings) but this is a bigger change in notification module, let's keep that in mind

Issue #2307, opened 2019-01-14
Scroll and focus first invalid input


desfero:
**Background:**
It's a usual behavior that UI scroll and focus first invalid input for better usability/accessibility.

glupiawymowka:
*tested on .io, ISSUER FORMS*

focus correctly moves to the first invalid input 
![image](https://user-images.githubusercontent.com/18315589/51915296-0af2b900-23db-11e9-89e7-ad6c8a177b58.png)

PR #2321, opened 2019-01-17
#2307 Scroll to first invalid field

desfero:


desfero:
Visual regression is ok

Issue #1928, opened 2018-11-20
migrated locked account agreement download - use correct lockAccount instance


rudolfix:
currently always euroLock is used. we should use proper instance
```
const [, , agreementUrl] = yield contractsService.euroLock.currentAgreement();
  const fileUri = agreementUrl.replace("ipfs:", "");
```
it should be as follows
1. upgrade wallet - should be on the wallet instance that is being upgraded
2. migration tool - always on ether wallet

PR #2330, opened 2019-01-20
#1928 Fix Agreement Download to use correct LockedAccount instance

drakkein:


app/modules/tx/transactions/upgrade/sagas.ts
```diff
@@ -42,6 +42,7 @@ export function* generateEuroUpgradeTransaction({
   const txDetails: ITxData = {
     ...txInitialDetails,
     gas: addHexPrefix(estimatedGasWithOverhead),
+    tokenType: ETokenType.EURO,
```
Moejoe90: Why do you need to include `tokenType` here what is the purpose?

drakkein: I needed the type of wallet during the wallet upgrade process. If there is possibility to detect it with 100% certainty, for example based hex prefix in value or something, then this is unnecessary.

PR #2341, opened 2019-01-22
Blacklist query params that we send to sentry

desfero:


app/lib/dependencies/logger/SentryLogger.ts
```diff
@@ -0,0 +1,115 @@
...
+  withScope,
+} from "@sentry/browser";
+import { injectable } from "inversify";
+
+import { ErrorArgs, ILogger, LogArg, TUser } from "./ILogger";
+
+/**
+ * Query params that should never be send to sentry for security reasons.
+ */
+const BLACKLISTED_QUERY_PARAMS = ["email", "salt"];
```
Moejoe90: :+1:

PR #2351, opened 2019-01-24
[WIP] Add storybook info addon

desfero:
This will add some useful info about component so it's easier to use storybook as a pattern library.
There is a downside as the first build will take longer than usual because addon needs to analyze typings.

Example:
image


desfero:
Closing for now, will open later after build is stable

Issue #1529, opened 2018-10-10
Research and implement batching for web3 requests


rudolfix:
We are doing tens of requests on each page. It is possible to batch requests going to Parity.
https://www.jsonrpc.org/specification#batch

those guys succesfully implemented it and report 10x boost
https://cloud.google.com/blog/products/data-analytics/ethereum-bigquery-how-we-built-dataset

IMO there are two ways of doing it
1. proxy and delay - we do our requests in bursts so we write a proxy that gathers requests for XXX miliseconds and then does single request and fulfills all promises after that. In that case we do not need to modify other code
2. explicit batching. we batch in each place that makes many calls in a row - explicitly in sagas etc.

PR #2355, opened 2019-01-25
#1529 Add Web3 auto request batching mechanism

desfero:
Some observation/thoughts:
The number of node requests was reduced from 146 to 17 on the dashboard for `.io`.
We get really huge perf boost on initial load by loading contracts and platform terms asynchronously.
For e.g. saga `populatePlatformTermsConstants` previously took 6s (because of 18 synchronous requests to node) by loading all contract constants asynchronously we decrease time to **500ms** (without batching) and to 400ms after batching. Similar benefits we get also while loading asynchronously contracts.

Things to do:

- [x] Add unit tests
- [x] Move console.log to our logger

@rudolfix @Moejoe90 let me know what do you think, if it's ok I will then write some unit tests.


app/lib/web3/Web3Batch.ts
```diff
@@ -0,0 +1,55 @@
...
+      // execute batch request on the next event loop
+      setTimeout(this.execute, 0);
+    }
+
+    this.web3Batch.add(request);
+  }
+
+  private execute = () => {
+    // tslint:disable-next-line:no-console
+    console.log("Number of request batched", this.web3Batch.requests.length);
```
desfero: Just temporary to test benefits

app/lib/web3/Web3Batch.ts
```diff
@@ -0,0 +1,55 @@
...
+  // TODO: Add correct typings when available in `web3-typescript-typings`
+  web3Batch: any;
+
+  constructor(private web3: Web3) {}
+
+  add(request: any): any {
+    if (this.web3Batch === undefined) {
+      this.web3Batch = (this.web3 as any).createBatch();
+      // execute batch request on the next event loop
+      setTimeout(this.execute, 0);
```
Moejoe90: What happens if we increase the timeout from `0` let's say to `1` milliseconds?
I think we need to give a small breathing space for the requests to happen 
But if the requests are really down from 147 to 17 then this is already a good push

app/lib/web3/Web3Batch.ts
```diff
@@ -0,0 +1,55 @@
+import * as Web3 from "web3";
+
+import { SelectPropertyNames } from "../../types";
+
+/**
+ * Wrapper on top of web3 Batch API to execute batch request on the next event loop cycle
+ */
+class Web3AutoExecuteBatch {
+  // TODO: Add correct typings when available in `web3-typescript-typings`
```
Moejoe90: https://github.com/0xProject/web3-typescript-typings is deprecated btw 
this is where they put the latest files 
https://github.com/0xProject/0x-monorepo/blob/55dbb0ece06d17a9db7b93a0ffa274ff65298002/packages/typescript-typings/types/web3/index.d.ts 
but still no typings for batching :(

app/lib/web3/Web3Batch.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Extends Web3 by auto batching common ethereum RPC requests
+ */
+class Web3Batch extends Web3 {
+  batch: Web3AutoExecuteBatch = new Web3AutoExecuteBatch(this);
+
+  constructor(...args: any[]) {
+    super(...args);
+
+    // List of web3.eth methods which calls should be batched in one request when possible
+    const methods: Web3EthMethodNames[] = ["call", "getCode", "getBalance"];
```
Moejoe90: Why not batch everything? 
Really everything as far as i know everything can be batched

app/lib/web3/Web3Batch.ts
desfero: yes :(

app/lib/web3/Web3Batch.ts
desfero: sure, we can add more methods, but for security (as we now monkey patch web3) i don't want to modify in any kind globally methods that modify chain (like `sendTransaction`)

app/test-e2e/utils/ethRpcUtils.ts
```diff
@@ -3,7 +3,7 @@ export enum ETransactionStatus {
   REVERTED = "0X0",
 }
 
-export const requestFromWeb3Node = (
+const requestFromWeb3Node = (
```
desfero: should not be used outside of this file

app/test-e2e/routing/EthererumNodeRPCBatching.spec.e2e.ts
```diff
@@ -0,0 +1,21 @@
...
+      method: "POST",
+      url: process.env.NF_RPC_PROVIDER as string,
+    }).as("node");
+
+    createAndLoginNewUser({ type: "investor" }).then(() => {
+      goToDashboard();
+      assertDashboard();
+
+      cy.get("@node").should((xhr: any) => {
+        expect(xhr.requestBody).to.be.an("array");
```
desfero: It's very naive but should work as there is no way to get all request by `alias` only last one can be retrieved

app/lib/web3/ContractsService.ts
```diff
@@ -89,21 +89,31 @@ export class ContractsService {
...
+      this.etherLock,
+      this.icbmEuroLock,
+      this.icbmEtherLock,
+      this.rateOracle,
+      this.identityRegistry,
+      this.platformTerms,
+      this.euroToken,
+      this.etherToken,
+      this.feeDisbursal,
+    ] = await Promise.all<any>([
```
desfero: need to use `any` here as there are no typings (at least correct one for our use case) when more then 10 items in a tuple

app/lib/web3/Web3Batch.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+import * as Web3 from "web3";
+
+import { createMock } from "../../../test/testUtils";
+import { noopLogger } from "../dependencies/logger";
+import { Web3AutoExecuteBatch } from "./Web3Batch";
+
+const createJsonRPCRequest = (method: string, params: any[] = []) => ({
+  params,
+  method,
+  id: random(1, 1000),
```
Moejoe90: why random?
and not an incrementing number. 

as i think it will cause problems when id is the same 
Maybe this is an extreme edge case

PR #2356, opened 2019-01-27
Lock bookbuilding pledge length to 8 digits

drakkein:
By design number type inputs does ignore the maxLength attribute. I tried some of the options ex. writing synthetic length check, but there is a lot of things to validate, ex. paste event. 
After this i came to conclusion that setting type to string is the best option here. 

Need your opinion here @desfero @Moejoe90 @AlexeyTuboltsev


app/components/eto/overview/EtoOverviewStatus/CampaigningWidget/CampaigningActivatedInvestorApprovedWidgetLayout.tsx
```diff
@@ -98,7 +98,7 @@ const CampaigningActivatedInvestorApprovedWidgetLayout: React.FunctionComponent<
                 name="amount"
                 prefix="€"
                 maxLength={8}
-                type="number"
+                type="text"
```
Moejoe90: yes :+1:

app/test-e2e/eto/Campaigning.spec.e2e.ts
```diff
@@ -121,6 +121,7 @@ describe("Eto campaigning state", () => {
 
             confirmAccessModal();
 
+            cy.wait(1000);
```
desfero: Do we need to wait here?

drakkein: Well, sometimes cypress fetches value before it changes, so with this we will eliminate error that `500 !== 499`

PR #2376, opened 2019-01-29
At/promisify: Write an own promisify function and remove Bluebird from dependencies

AlexeyTuboltsev:


AlexeyTuboltsev:
visual regression is ok

PR #2381, opened 2019-01-30
[hotfix] At/#2380 fix translations in displayInfoModalSaga

AlexeyTuboltsev:
fixes translation bugs in displayInfoModalSaga
changes neuCall to simple saga::call since we don't need dep.injection in this saga any more


app/components/modals/walletAccess/AccessWalletModal.tsx
```diff
@@ -83,7 +82,7 @@ export const AccessWalletContainer = appConnect<IStateProps, IDispatchProps, IEx
       : s.accessWallet.modalTitle && getMessageTranslation(s.accessWallet.modalTitle),
     message: external.message
       ? external.message
-      : s.accessWallet.modalTitle && getMessageTranslation(s.accessWallet.modalMessage as TMessage),
```
Moejoe90: casting is the root of all evil

Issue #2340, opened 2019-01-22
pending/my assets portfolio UI and UX cleanup


rudolfix:
We have a few simple fixes that we'd like to implement

1. We link pending and my assets to company public profile. Kirstyn please tell us should we use a link or the whole area is clickable?
2. We drop the NEU Reward column from MY ASSETS. REWARD applies only to pending assets
3. We rename column DOCUMENTS to INVESTOR AGREEMENTS. In this column we have a single button/link to download documents
4. When link/button is clicked
- in case of NEU we go to commitment page (as it is now! - we do not have NEU documents download implemented yet)
- in case of any other (equity) token we show a modal where you can download all three investor documents separately (so the links you see now are going to this modal).

Steps
1. @falinebambi  does the designs
2. @drakkein will do implementation

deadline on production is 2nd of Feb

falinebambi:
https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/dashboard?seid=5c49eda033fdc7bf07c0e0b6
![portfolio document download](https://user-images.githubusercontent.com/17235402/51694527-4b2ff100-2001-11e9-9163-177ce4410d8d.png)
![updated portfolio](https://user-images.githubusercontent.com/17235402/51694528-4bc88780-2001-11e9-995e-bda1f795f8b6.png)

rudolfix:
I've just noticed that we never defined what should go into MY ASSETS so let's reiterate
QUANTITY - token.balanceOf(user wallet address)
CURRENT PRICE
- for NEU we take it from price oracle - this is already implemented OK
- for equity token we need to take current market price from the token controller (right now we use same price as in RESERVED ASSETS which is wrong)
CURRENT VALUE
- multiply QUANTITY * CURRENT PRICE

How to get market price from token controller:
- instantiate IEquityToken artifact from equity token address (alternatively you can use EquityToken which implements this interface)
- get two values: IEquityToken::tokensPerShare() and IEquityToken::tokenController()
- instantiate IControllerGovernance artifact from token controller address
- call IControllerGovernance::shareholderInformation() which returns
```
function shareholderInformation()
        public
        constant
        returns (
            uint256 totalCompanyShares,
            uint256 companyValuationEurUlps,
            ShareholderRights shareholderRights
        );
```
now the market token price is (market valuation / existing shares) / tokens per share
please use big number for calculations. final price is in wei (10**18) EUR

check platform-contracts repo for all the smart contracts mentioned above

PR #2387, opened 2019-01-31
#2340 Portfolio assets cleanup

drakkein:


app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+                  }
+                >
+                  <DocumentTemplateButton
+                    layout={EButtonLayout.SECONDARY}
+                    title={getDocumentTitles(isRetailEto)[document.documentType]}
+                  />
+                  <InlineIcon
+                    className={styles.downloadIcon}
+                    svgIcon={downloadIcon}
+                    height={"18px"}
```
Moejoe90: why not put height and width in the scss file?

app/modules/public-etos/sagas.ts
```diff
@@ -4,6 +4,7 @@ import { compose, keyBy, map, omit } from "lodash/fp";
 import { delay } from "redux-saga";
 import { all, fork, put, race, select } from "redux-saga/effects";
 
+import BigNumber from "bignumber.js";
```
desfero: move above

app/modules/public-etos/sagas.ts
```diff
@@ -335,6 +338,46 @@ function* downloadTemplateByType(
...
+  { contractsService }: TGlobalDependencies,
+  action: TActionFromCreator<typeof actions.publicEtos.loadTokensData>,
+): any {
+  const myAssets = yield select(selectMyAssets);
+
+  for (const eto of myAssets) {
+    const equityTokenAddress = eto.contract.equityTokenAddress;
+
+    const equityToken = yield contractsService.getEquityToken(equityTokenAddress);
+    const balance = yield equityToken.balanceOf(action.payload.walletAddress);
```
desfero: batch requests here (from 351 to 355), otherwise it will block ui longer than needed (use `all` from redux-saga)

app/components/portfolio/PortfolioAssetAction.tsx
```diff
@@ -14,42 +18,60 @@ type TExternalProps = {
...
           <FormattedMessage id="portfolio.section.reserved-assets.claim-tokens" />
         </Button>
       );
-    // case ETOStateOnChain.Refund:
-    //   return (
-    //     <Button layout="simple" svgIcon={arrowIcon} iconPosition="icon-after">
-    //       <FormattedMessage id="portfolio.section.reserved-assets.refund" />
-    //     </Button>
-    //   );
+    case EETOStateOnChain.Refund:
```
Moejoe90: since it's not developed and now you uncommented it  I would add a `disabled` here

app/modules/public-etos/sagas.ts
```diff
@@ -335,6 +338,46 @@ function* downloadTemplateByType(
...
+    const tokenController = yield equityToken.tokenController;
+
+    const controllerGovernance = yield contractsService.getControllerGovernance(tokenController);
+    const [
+      totalCompanyShares,
+      companyValuationEurUlps,
+      ,
+    ] = yield controllerGovernance.shareholderInformation();
+
+    const tokenPrice = new BigNumber(
```
desfero: Probably we should not convert it back to `BigNumber` as there is a force from @AlexeyTuboltsev to use plain types inside `store`

app/modules/reducer.ts
```diff
@@ -63,4 +64,5 @@ export const appReducers = {
   investmentFlow: investmentFlowReducer,
   videoModal: videoModalReducer,
   personProfileModal: personProfileModalReducer,
+  portfolioDownloadAgreementsModal: portfolioDownloadAgreementsModalReducer,
```
desfero: @drakkein for future reference it's better to use `GenericModal` as this complicates reducer a lot with unneeded states

app/modules/public-etos/actions.ts
```diff
@@ -49,4 +49,14 @@ export const etoActions = {
     (previewCode: string, data: IEtoContractData) => ({ previewCode, data }),
   ),
   setEtoWidgetError: createActionFactory("PUBLIC_ETOS_SET_ETO_WIDGET_ERROR"),
+  loadTokensData: createActionFactory("PORTFOLIO_LOAD_TOKENS_DATA", (walletAddress: string) => ({
```
desfero: Move to `public actions` as this is not state mutation

app/components/portfolio/PortfolioMyAssets.tsx
```diff
@@ -0,0 +1,184 @@
...
+        neuValue: multiplyBigNumbers([myNeuBalance, neuPrice]),
+        myAssets: selectMyAssetsWithTokenData(state)!,
+      };
+    },
+    dispatchToProps: dispatch => ({
+      showDownloadAgreementModal: (etoId: string, isRetailEto: boolean) => {
+        dispatch(actions.portfolio.showDownloadAgreementModal(etoId, isRetailEto));
+      },
+      loadTokensData: (walletAddress: string) => {
+        dispatch(actions.publicEtos.loadTokensData(walletAddress));
```
desfero: btw, why this `loadTokensData` is not inside `InvestorTickets` as this is strictly related to porfolio only? Or not?

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+  selectDownloadAgreementModalIsRetailEto,
+  selectDownloadAgrementModalIsOpen,
+} from "../../portfolio/selectors";
+import { EButtonLayout } from "../../shared/buttons";
+import { DocumentTemplateButton } from "../../shared/DocumentLink";
+import { SectionHeader } from "../../shared/SectionHeader";
+import { ModalComponentBody } from "../ModalComponentBody";
+
+import * as downloadIcon from "../../../assets/img/inline_icons/download.svg";
+import { InlineIcon } from "../../shared/InlineIcon";
```
desfero: move above

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+            }, eto.templates)}
+            <p className="mb-0" />
+          </section>
+        )}
+      </ModalComponentBody>
+    </Modal>
+  );
+};
+
+const DownloadTokenAgreementModal = compose<any, any>(
```
desfero: `any`, `any`?

desfero: If you do not expect any props probably you should use `{}` so it's clear for reader

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+  appConnect<IStateProps, IDispatchProps>({
+    stateToProps: state => {
+      const etoId = selectDownloadAgreementModalEtoId(state);
+      const eto = etoId ? selectEtoWithCompanyAndContractById(state, etoId) : undefined;
+      return {
+        isOpen: etoId ? selectDownloadAgrementModalIsOpen(state) : false,
+        isRetailEto: etoId ? selectDownloadAgreementModalIsRetailEto(state)! : false,
```
desfero: do we need to `!` at the end?

app/components/portfolio/PortfolioAssetAction.tsx
```diff
@@ -14,42 +18,60 @@ type TExternalProps = {
...
+          svgIcon={arrowRight}
+          innerClassName={"text-uppercase"}
+        >
+          <FormattedMessage id="portfolio.section.reserved-assets.refund" />
+        </Button>
+      );
     case EETOStateOnChain.Signing:
       return (
-        <Button layout={EButtonLayout.SECONDARY} disabled>
+        <Button layout={EButtonLayout.SIMPLE} disabled innerClassName={"text-uppercase"}>
```
Moejoe90: why did you remove the `disabled` button this is also an unfinished feature

app/lib/web3/ContractsService.ts
```diff
@@ -95,6 +99,26 @@ export class ContractsService {
     this.etoCommitmentCache[etoId] = contract;
     return contract;
   }
+
+  async getEquityToken(equityTokenAddress: string): Promise<IEquityToken> {
+    if (this.equityTokensCache[equityTokenAddress])
+      return this.equityTokensCache[equityTokenAddress];
+
+    const web3 = this.web3Manager.internalWeb3Adapter.web3;
```
AlexeyTuboltsev: hey, this statement repeats 4 times! (lines 49, 97, 107, 117). Maybe just add it as a private member to this class?

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+                  <InlineIcon
+                    className={styles.downloadIcon}
+                    svgIcon={downloadIcon}
+                    height={"18px"}
+                    width={"18px"}
+                  />
+                </div>
+              ) : null;
+            }, eto.templates)}
+            <p className="mb-0" />
```
desfero: Do we need empty `p`?

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.module.scss
```diff
@@ -0,0 +1,25 @@
...
+
+  &:nth-of-type(2n) {
+    background-color: transparent;
+  }
+
+  .download-icon {
+    margin-left: auto;
+    margin-right: 15px;
+
+    svg {
```
desfero: do we need this if you provide size as a prop?

app/components/shared/NewTable.tsx
```diff
@@ -63,21 +76,28 @@ const NewTable: React.FunctionComponent<TProps> = ({
...
           <thead className={styles.header}>
             <tr>
-              {titles.map((title, index) => (
-                <th className={styles.cell} key={index}>
-                  {title}
-                </th>
-              ))}
+              {titles.map((title, index) => {
+                return typeof title === "object" && title && "width" in title ? (
+                  <th className={styles.cell} key={index} style={{ width: title.width }}>
```
AlexeyTuboltsev: better would be
return title && typeof title === "object" && "width" in title ?

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.module.scss
```diff
@@ -0,0 +1,25 @@
...
+      width: 18px;
+      height: 18px;
+    }
+
+    path {
```
desfero: maybe it's possible to set `color` to `.download-icon` as not you break coupling because you assume that component inside will have `path` node

app/components/portfolio/PortfolioAssetAction.tsx
```diff
@@ -14,42 +18,60 @@ type TExternalProps = {
...
       );
-    // case ETOStateOnChain.Refund:
-    //   return (
-    //     <Button layout="simple" svgIcon={arrowIcon} iconPosition="icon-after">
-    //       <FormattedMessage id="portfolio.section.reserved-assets.refund" />
-    //     </Button>
-    //   );
+    case EETOStateOnChain.Refund:
+      return (
+        <Button
```
desfero: But refund is not ready yet?

desfero: It even don't have any `onClick` attached so it's useless button to show

app/components/portfolio/PortfolioLayout.tsx
```diff
@@ -89,224 +64,8 @@ const PortfolioLayout: React.FunctionComponent<TPortfolioLayoutProps> = ({
...
-                      ) : null;
-                    }, templates)}
-                  </>
-                </NewTableRow>
-              );
-            },
-          )}
-        </NewTable>
-      </Col>
-    </Row>
```
Moejoe90: :+1:

app/components/portfolio/PortfolioAssetAction.tsx
```diff
@@ -14,42 +18,60 @@ type TExternalProps = {
...
+          layout={EButtonLayout.SIMPLE}
```
desfero: Simple button should be only used when you want to build your own custom button as it doen't have any behavior attached for e.g. `:hover`, why we can not stick with `SECONDARY`?

app/components/portfolio/PortfolioMyAssets.tsx
```diff
@@ -0,0 +1,184 @@
...
+                value={neuValue}
+                currency={ECurrency.EUR}
+                currencySymbol={ECurrencySymbol.SYMBOL}
+              />
+              <Money
+                value={neuPrice}
+                currency={ECurrency.EUR}
+                currencySymbol={ECurrencySymbol.SYMBOL}
+              />
+              <ButtonLink
```
Moejoe90: There is a new component that got merged lately for blank targets can you please use it

app/components/portfolio/PortfolioMyAssets.tsx
```diff
@@ -0,0 +1,184 @@
...
+                innerClassName={"text-uppercase"}
+              >
+                <FormattedMessage id="portfolio.section.my-assets.download-agreements" />
+              </ButtonLink>
+            </NewTableRow>
+          ) : null}
+
+          {myAssets &&
+            myAssets
+              .filter(v => v.tokenData)
```
Moejoe90: I'm not sure but be a bit cautious here as `tokenData` are all bigNumbers. So I don't see much use for checking if they will be available.

app/components/portfolio/PortfolioMyAssets.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+const PortfolioMyAssets = compose<TComponentProps, IExternalProps>(
+  appConnect<IStateProps, IDispatchProps>({
+    stateToProps: state => {
+      const neuPrice = selectNeuPriceEur(state);
+      const myNeuBalance = selectNeuBalance(state.wallet);
+
+      return {
+        myNeuBalance,
+        neuPrice: new BigNumber(neuPrice).toFixed(8),
```
Moejoe90: this looks like a good use case for a selector

app/components/shared/NewTable.tsx
```diff
@@ -63,21 +76,28 @@ const NewTable: React.FunctionComponent<TProps> = ({
...
+                    {title.title}
```
AlexeyTuboltsev: now we have either a title (string/TTranslated smth) or an object with many fields unrelated to title. This all gets too convoluted. I suggest to leave the titles to be simply an array of titles (e.g. strings/translations) and if you need to force width on some columns, do it separately. Like add another field to NewTable -- Columns, an array of column widths  (like[null, 5px, null, 20px]).

or, maybe it's easier
convert the titles to on object everywhere, call it columnSettings or smth like this, and pack all this stuff in this object. Like ColumnSettings{title?:string|translation, width?:number, customCSS?:string } etc

app/components/portfolio/PortfolioMyAssets.tsx
```diff
@@ -0,0 +1,184 @@
+import * as cn from "classnames";
+import * as React from "react";
+import { FormattedMessage } from "react-intl-phraseapp";
+import { Col, Row } from "reactstrap";
+import { compose, lifecycle } from "recompose";
+
+import BigNumber from "bignumber.js";
```
Moejoe90: import order please

app/modules/public-etos/sagas.ts
```diff
@@ -335,6 +338,46 @@ function* downloadTemplateByType(
...
+export function* loadTokensData(
+  { contractsService }: TGlobalDependencies,
+  action: TActionFromCreator<typeof actions.publicEtos.loadTokensData>,
+): any {
+  const myAssets = yield select(selectMyAssets);
+
+  for (const eto of myAssets) {
+    const equityTokenAddress = eto.contract.equityTokenAddress;
+
+    const equityToken = yield contractsService.getEquityToken(equityTokenAddress);
```
Moejoe90: This can all be packed into one `Promise.all`
have a look at this
https://github.com/Neufund/platform-frontend/blob/71f8594dba75274b312fcc0862e0b1fc1d57704b/app/modules/wallet-selector/ledger-wizard/sagas.ts#L55
Or this example from a WIP branch
https://github.com/Neufund/platform-frontend/pull/2355/files#diff-453914c9822223caa684238d4e1cd5f2R88

app/components/portfolio/reducer.ts
```diff
@@ -0,0 +1,39 @@
...
+      return {
+        isOpen: true,
+        etoId: action.payload.etoId,
+        isRetailEto: action.payload.isRetailEto,
+      };
+    }
+    case actions.portfolio.hideDownloadAgreementModal.getType(): {
+      return {
+        isOpen: false,
+      };
```
AlexeyTuboltsev: maybe return initial state then? Or remove etoId and isRetailEto from initial state, for consistency's sake

app/components/appRoutes.ts
```diff
@@ -26,5 +26,5 @@ export const appRoutes = {
 
   etoRegister: "/eto/registration",
 
-  portfolio: "/portfolio",
```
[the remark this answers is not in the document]
drakkein: woah, that was Webstorms refactoring tool

app/components/portfolio/PortfolioReservedAssets.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    <Row>
+      <Col className={styles.reservedAssets}>
+        <NewTable
+          keepRhythm={true}
+          panelStyle={styles.panel}
+          placeholder={
+            <FormattedMessage id="portfolio.section.reserved-assets.table.header.placeholder" />
+          }
+          titles={[
```
AlexeyTuboltsev: see comment below (app/components/shared/NewTable.tsx). This is too messy. Need to separate formatting (forcing width etc) from content (titles)

app/components/portfolio/PortfolioLayout.tsx
```diff
@@ -89,224 +64,8 @@ const PortfolioLayout: React.FunctionComponent<TPortfolioLayoutProps> = ({
...
+    <PortfolioReservedAssets pendingAssets={pendingAssets} />
+    <PortfolioMyAssets isRetailEto={isRetailEto} walletAddress={walletAddress} />
```
AlexeyTuboltsev: 👍

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+              <FormattedMessage id="portfolio.section.my-assets.modal.header" />
+            </SectionHeader>
+            {map((document: IEtoDocument) => {
+              return [EEtoDocumentType.SIGNED_INVESTMENT_AND_SHAREHOLDER_AGREEMENT].includes(
+                document.documentType,
+              ) ? (
+                <div
+                  key={document.documentType}
+                  className={styles.document}
+                  onClick={() =>
```
desfero: onClick on div is not correct

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.module.scss
```diff
@@ -0,0 +1,25 @@
+.document {
+  display: flex;
+  align-items: center;
+  height: 46px;
+  background-color: $neu-pale-blue;
+  cursor: pointer;
```
desfero: cursor: pointer should not be added to not focusable elements

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
```diff
@@ -0,0 +1,146 @@
...
+  onClose,
+  generateTemplateByEtoId,
+  downloadDocument,
+  eto,
+  isRetailEto,
+}) => {
+  return (
+    <Modal isOpen={isOpen} toggle={onClose}>
+      <ModalComponentBody onClose={onClose}>
+        {eto && (
```
AlexeyTuboltsev: what happens if there's no eto? Maybe add a branch with loading state or some error message?

app/modules/public-etos/sagas.ts
Moejoe90: :+1: 
For sure we need to only put plain types inside store 
We already have a case where date is in the store 
https://redux.js.org/faq/organizing-state#can-i-put-functions-promises-or-other-non-serializable-items-in-my-store-state

app/components/portfolio/PortfolioLayout.module.scss
```diff
@@ -7,14 +7,30 @@
...
+  text-transform: uppercase;
+  white-space: nowrap;
+}
+
 .document-link {
   margin-bottom: px-to-rem(16px);
   text-align: left;
 }
+
+.reserved-assets,
```
desfero: We should not change styles for tables only in portfolio as this is inconsistent and will lead to some weird UX. If design team wants to change table then it should change all of them

app/components/portfolio/PortfolioMyAssets.tsx
rudolfix: equity token data is not part of InvestorTickets. you can easily have equity token that you bought on secondary market and never invested in it via ETO. so you can have a token in MY ASSETS that you do not have and never had in PENDING ASSETS. you could just got it via `transfer` function from other investor.
So having separate state for Tokens Data is good

rudolfix:
I've checked price calculations in MY ASSETS and it's good. thanks @drakkein

app/modules/reducer.ts
drakkein: Lets leave it this way, I'll create ticket to refactor this

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
drakkein: 👍

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.module.scss
drakkein: it wont work :(

app/components/portfolio/PortfolioAssetAction.tsx
drakkein: Lets disable it

app/components/portfolio/PortfolioAssetAction.tsx
drakkein: ok, but there is need to inject additional styles via `innerClassName` with `padding: 0 !important;` to achieve correct look

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
drakkein: changed to be anchor

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
drakkein: It's set anyway to override inline svgs size, I'll leave it only in scss

app/components/shared/NewTable.tsx
drakkein: Let's make additional issue for `NewTable` improvements

app/components/portfolio/reducer.ts
drakkein: initialState is best option i suppose

app/components/modals/download-token-agreements-modal/DownloadTokenAgreementModal.tsx
drakkein: this condition is only for initial state, there should be no case when eto is missing and modal is open

app/components/portfolio/PortfolioReservedAssets.tsx
drakkein: Like previously, let's make additional issue for `NewTable`

Issue #1733, opened 2018-11-05
Portfolio Payout flow.


SergejSoft:
There is a new design for Portfolio functions that reached the stage of PayOut. 

User case: User invested in the ETO and after a successful campaign, the user is able to claim equity tokens and rewards in NEU.

**Desings**:
List: https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5c51c8019c3e07307b056e1a
No payouts: https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5c5320a1881152354d831cb7
Accept All Summary: https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5c4eff5466e3b3bee976ec3f
Accept nEur summary: https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5c4eff5487c26e79e4575c65
Accept eth summary:
https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5c4eff56c50a21377f681154
Accept success: https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5c51c8029c3e07307b056f3e
Redistribute summary: zpl.io/b6jWdEN
Redistribute confirm: zpl.io/VkwPLm5
Redistribute success: zpl.io/V1zLg8Z

**Progress:**

- [x] Display payout
- [x] Approve payout
- [x] Redistribute payout
- [x] Add NEU icon to header (would be done as a part of #2411)
- [ ] E2E tests for accept payout
- [ ] E2E tests for redistribute payout

**Testing notes**
Please refer to the above progress checklist to see what's already merged to master (`.io`).
To test payout you need to have some payouts so correct fixture should be used. Fortunately, a lot of fixtures are already with available payouts like INV_ICBM_EUR_M_HAS_KYC, INV_HAS_EUR_HAS_KYC and INV_EUR_ICBM_HAS_K.

rudolfix:
Attaching UI to smart contracts:
You can find full information in **FeeDisbursal** contract with source code here: https://github.com/Neufund/platform-contracts/blob/master/contracts/FeeDisbursal/FeeDisbursal.sol
It's pretty well documented.

1. our proRata token is always NEU (Neumark). you can assume proRataToken with NEU address everywhere. We'll have bigger redesign of all dashboards so I would not waste time for anything universal.

2.
[...]

PR #2389, opened 2019-01-31
#1733 Display payouts

desfero:
NEU icon is not displayed on the left side from `Overview of pending payouts from NEU` as I need to make some refactoring to `SectionHeader` to make this happen. Will do it as a separate pull requests.

Plus there is inconsistency in naming. Frontend use `currency` and contracts use `token` that need be discussed later within the team and standardised


app/components/shared/AssetPortfolio.tsx
```diff
@@ -1,78 +1,132 @@
...
+    case ECurrency.EUR_TOKEN:
+      return <img src={nEurIcon} alt="NEur token" />;
+    case ECurrency.ETH:
+      return <img src={ethIcon} alt="Ether token" />;
+    default:
+      throw new Error(`Icon for currency ${currency} not found`);
+  }
+};
+
+const AssetPortfolioLayoutNoPayouts: React.FunctionComponent = () => (
```
drakkein: 👍

Issue #2393, opened 2019-02-01
Remove Cypress test routes from prod build


desfero:
**Background**
Currently `IS_CYPRESS` constant use `window.Cypress` check. This works but we land shipping in the production build all test routes. We should introduce env variable `IS_CYPRESS` for all cypress related npm scripts and replace `window.Cypress` by `process.env.IS_CYPRESS`.

**Component**
`constants.ts`

**Acceptance criteria**
Given I build bundle in production mode when I look in the source code then all routes related to e2e tests should be removed.

**Testing notes**
Not changes in the UI. If e2e tests pass then we are good to go.

PR #2399, opened 2019-02-01
adds isCypress ENV

Moejoe90:
also routing e2e test
https://github.com/Neufund/platform-frontend/issues/2393


app/config/constants.ts
```diff
@@ -72,4 +72,4 @@ export const BROWSER_WALLET_RECONNECT_INTERVAL = 1000;
 /*
  * General constants
  */
-export const IS_CYPRESS = typeof window === "undefined" ? false : !!(window as any).Cypress;
+export const IS_CYPRESS = typeof window === "undefined" ? false : !!process.env.IS_CYPRESS;
```
desfero: @Moejoe90 are you sure it will correctly test routes from production build? Why simply not `process.env.IS_CYPRESS`

Moejoe90: I'll change it accordingly your comment makes sense

Issue #2403, opened 2019-02-02
[QUICK FIX] when downloading personalized agreements in token claim we wait too long without indication


rudolfix:
related to https://github.com/Neufund/platform-frontend/issues/2102

when user clicks to download document: both when claiming tokens and then in MY ASSETS we make two calls to backend: resolve placeholders and then generate personalized PDF.

on the backend there's libre office doing word -> pdf. it's slow. user waits for 2-3 seconds without a clue what is going on so the user clicks again.

quick fix:
* change mouse cursor to WAIT
* prevent clicks until document starts downloading (or error!)

**Zepplin**
Quick fix - not needed

**Component**
Portfolio - Token Claim process

**Acceptance criterea**
- cursor changes or other indication making user wait when request is being processed
- cannot download same document when request is being processed

**Testing notes**
First click makes you wait for a long time. next clikcs on the same documents are cached. we need branch testing

glupiawymowka:
the cursor indicates now, that some action was taken and the user should wait.

PR #2404, opened 2019-02-03
#2403 Lock duplicated document download

drakkein:


app/modules/eto-documents/sagas.ts
```diff
@@ -38,14 +40,13 @@ export function* generateDocumentFromTemplate(
...
-      },
+      ipfsHash: templates.ipfs_hash,
+      mimeType: templates.mime_type,
+      placeholders: templates.placeholders,
       asPdf: false,
     });
     yield call(downloadLink, generatedDocument, document.name, ".doc");
+    yield put(actions.immutableStorage.downloadImmutableFileDone(document.ipfsHash));
   } catch (e) {
     logger.error("Failed to generate ETO template", e);
```
Moejoe90: Watch out because you have 
```
yield put(actions.immutableStorage.downloadDocumentStarted(document.ipfsHash));
```
If the flow is good then it reaches
```
    yield put(actions.immutableStorage.downloadImmutableFileDone(document.ipfsHash));
```
But if an error while downlaoding the file happens the button is gonna be disabled forever

app/modules/eto-documents/sagas.ts
```diff
@@ -84,6 +87,7 @@ export function* generateDocumentFromTemplateByEtoId(
       asPdf: true,
     });
     yield call(downloadLink, generatedDocument, document.name, extension);
+    yield put(actions.immutableStorage.downloadImmutableFileDone(document.ipfsHash));
   } catch (e) {
     logger.error("Failed to generate ETO template", e);
```
Moejoe90: Same as the comment mentioned above

app/modules/immutable-file/actions.ts
```diff
@@ -1,8 +1,21 @@
...
-  downloadImmutableFile: (immutableFileId: ImmutableFileId, fileName: TMessage | string) =>
-    createAction("IMMUTABLE_STORAGE_DOWNLOAD_FILE", { immutableFileId, fileName }),
+  downloadImmutableFile: createActionFactory(
+    "IMMUTABLE_STORAGE_DOWNLOAD_FILE",
+    (immutableFileId: ImmutableFileId, fileName: TMessage | string) => ({
+      immutableFileId,
+      fileName,
+    }),
+  ),
+  downloadDocumentStarted: createActionFactory(
```
Moejoe90: `downloadImmutableFile` can be the indicator that the download started instead of having to have its own action `downloadDocumentStarted` 
But its ok both ways

app/modules/immutable-file/reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+    case actions.immutableStorage.downloadImmutableFile.getType(): {
+      return {
+        ...state,
+        pendingDownloads: {
+          ...state.pendingDownloads,
+          [action.payload.immutableFileId.ipfsHash]: true,
+        },
+      };
+    }
+    case actions.immutableStorage.downloadDocumentStarted.getType(): {
```
Moejoe90: whats the difference now between 
`actions.immutableStorage.downloadImmutableFile.getType()`
and
`actions.immutableStorage.downloadDocumentStarted.getType():`
Both alter the state in the same way

app/modules/immutable-file/sagas.ts
```diff
@@ -17,7 +18,13 @@ export function* downloadFile(
     const downloadedFile = yield apiImmutableStorage.getFile(immutableFileId);
     const extension = immutableFileId.asPdf ? ".pdf" : ".doc";
     yield call(downloadLink, downloadedFile, action.payload.fileName, extension);
+    yield put(
```
Moejoe90: What triggers this one here?
as you have only `actions.immutableStorage.downloadImmutableFileDone`

app/modules/immutable-file/reducer.ts
drakkein: For download documents I had to use document ipfsHash which is known to component instead of template ipfsHash which is passed into immutable storage, so this is reason for use of `actions.immutableStorage.downloadDocumentStarted()`

app/modules/eto-documents/sagas.ts
drakkein: good catch, thanks 👍

PR #2421, opened 2019-02-05
#1733 Approve payout

desfero:


desfero:
Visual regression ok

app/components/modals/tx-sender/investor-payout/Summary.tsx
```diff
@@ -0,0 +1,97 @@
...
+      </InfoList>
+      <section className="text-center">
+        <a
+          className="d-inline-block mb-3"
+          href={withParams(externalRoutes.commitmentStatus, { walletAddress })}
+        >
+          <FormattedMessage id="investor-payout.summary.neu-tokenholder-agreement" />
+        </a>
+        <p className={cn(styles.hint)}>
+          To process this payout we will need to trigger and ethereum transaction which will incure
```
desfero: Add to transaltions

app/modules/investor-portfolio/types.ts
```diff
@@ -4,7 +4,7 @@ import { ECurrency } from "../../components/shared/Money";
 import { IEtoTokenData, TEtoWithCompanyAndContract } from "../public-etos/types";
 
 export interface ITokenDisbursal {
-  currency: ECurrency;
+  token: ECurrency;
   amountToBeClaimed: string;
```
AlexeyTuboltsev: i think amountToBeClaimed should be a Bignumber,
the next line as well

app/components/modals/tx-sender/investor-payout/Summary.tsx
AlexeyTuboltsev: please move this to translations

app/modules/tx/transactions/payout/saga.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  const estimatedGasWithOverhead = yield web3Manager.estimateGasWithOverhead(txInitialDetails);
+
+  return {
+    ...txInitialDetails,
+    gas: estimatedGasWithOverhead,
+  };
+}
+
+export function* generatePayoutTokenTransaction(
```
AlexeyTuboltsev: these two sagas have almost the same code. Maybe make one of them with an additional argument txInitialDetails?

app/modules/investor-portfolio/types.ts
desfero: Discussed with @AlexeyTuboltsev, we should save strings to store

app/modules/tx/transactions/payout/saga.ts
desfero: maybe leave this as it's now because generally all transaction sagas look the same so we should prepare some kind of general saga `generateTransactionData` that we can use everywhere.

app/components/portfolio/AssetPortfolio.tsx
```diff
@@ -89,42 +108,76 @@ const AssetPortfolioLayout: React.FunctionComponent<ILayoutProps> = ({ tokensDis
...
+              size={ButtonSize.SMALL}
+              onClick={() => acceptPayout(tokenDisbursal)}
+              layout={EButtonLayout.SECONDARY}
+            >
+              <FormattedMessage id="portfolio.asset.payouts-from-neu.accept-payout" />
+            </Button>
+          </NewTableRow>
+        ))}
+        <NewTableRow>
+          <></>
```
Moejoe90: what is the benefit of this?

Desperado: basically cosmetical. more common words

PR #2430, opened 2019-02-07
Refactor ButtonIcon component

desfero:


app/components/shared/InlineIcon.stories.tsx
```diff
@@ -88,68 +37,19 @@ import * as nEUR_icon from "../../assets/img/nEUR_icon.svg";
 import * as token_icon from "../../assets/img/token_icon.svg";
 
 import { InlineIcon } from "./InlineIcon";
+
+// Load all inline icons
+const context = require.context("../../assets/img/inline_icons/", true, /\.(svg)$/);
+const allInlineIcons: ReadonlyArray<string> = context.keys().map<string>(context);
```
desfero: it's better to load all `inline_icons` folder then each icon separately

app/components/shared/PeopleSwiperWidget.tsx
```diff
@@ -218,9 +217,10 @@ class PeopleSwiperWidgeLayout extends React.PureComponent<IOwnProps & IDispatchP
     return (
       <div className={styles.swiperMain}>
         {showArrows && (
-          <InlineIcon
+          <ButtonIcon
```
desfero: we should not use `InlineIcon` with `onClick` attached, it's not semantically valid

desfero:
Visual regression ok

app/components/shared/buttons/ButtonLink.tsx
```diff
@@ -31,7 +31,7 @@ type TButtonDispatchProps = {
   navigate: () => void;
 };
 
-type TButtonWithoutOnClick = Omit<IButtonProps, IGeneralButton>;
+type TButtonWithoutOnClick = OmitKeys<IButtonProps, "onClick">;
```
Moejoe90: I would add onClick as a constant instead of a direct literal but non blocking

app/components/shared/InlineIcon.stories.tsx
```diff
@@ -88,68 +37,19 @@ import * as nEUR_icon from "../../assets/img/nEUR_icon.svg";
...
-        <span style={{ display: "inline-block", height: "2em", width: "3em" }}>
-          <InlineIcon svgIcon={svg} />
-        </span>
-      ))}
-      <br />
-    </>
-  ))
+  .add("InlineIcon", () =>
+    allInlineIcons.map((svg, i) => (
+      <span key={i} style={{ display: "inline-block", height: "2em", width: "3em" }}>
```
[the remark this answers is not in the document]
desfero: i don't really know, but it doesn't matter here as it's stories

app/components/shared/PeopleSwiperWidget.tsx
desfero: yeap, I went through all  usages, there is not `onClick` anymore

app/components/shared/buttons/ButtonLink.tsx
desfero: hm, i don't think it will make code easier to understand as i need to introduce in type that will only contain `"onClick"` because OmitKeys expect typescript type not constant. If i do that then also type name `TButtonWithoutOnClick` should be also changed as which keys are ommited is no longer hardcoded to type `OmitKeys<IButtonProps, "onClick">`

PR #2435, opened 2019-02-10
Download agreements e2e tests

drakkein:


app/modules/immutable-file/sagas.ts
```diff
@@ -17,6 +19,12 @@ export function* downloadFile(
     const immutableFileId = action.payload.immutableFileId;
     const downloadedFile = yield apiImmutableStorage.getFile(immutableFileId);
     const extension = immutableFileId.asPdf ? ".pdf" : ".doc";
+
+    if (IS_CYPRESS) {
+      // required for cypress to be able to simulate generating time
```
Moejoe90: Interesting trick how did you think of it?

app/components/portfolio/PortfolioMyAssets.tsx
```diff
@@ -107,6 +107,8 @@ const PortfolioMyAssetsComponent: React.FunctionComponent<TComponentProps> = ({
                 iconPosition="icon-after"
                 svgIcon={arrowRight}
                 size={ButtonSize.SMALL}
+                data-test-id="portfolio-my-assets-neu-agreements"
+                data-test-wallet-address={walletAddress}
```
Moejoe90: is this used anywhere?
`data-test-wallet-address`

app/test-e2e/portfolio/PortfolioDownloadDocuments.spec.e2e.ts
```diff
@@ -0,0 +1,48 @@
...
+          const downloadSelector = tid(
+            `modals.portfolio.portfolio-assets.download-agreements-${PUBLIC_ETO_ID}.download`,
+          );
+          cy.clock().then(clock => {
+            cy.get(downloadSelector)
+              .first()
+              .parent()
+              .click()
+              .then(() => {
+                cy.wait(100);
```
Moejoe90: no need to for `wait` here

app/modules/immutable-file/sagas.ts
drakkein: welll, `cy.clock` is taking control of every js time function includes `setTimeout` and `delay` is just `setTimeout` wrapped into `Promise`

app/components/portfolio/PortfolioMyAssets.tsx
drakkein: no, it can be removed

PR #2444, opened 2019-02-11
Small portfolio fixes

desfero:


app/components/portfolio/AssetPortfolio.tsx
```diff
@@ -191,6 +202,9 @@ const AssetPortfolio = compose<ILayoutProps & IDispatchToProps, IExternalProps>(
     ({ tokensDisbursal }) => tokensDisbursal.length === 0,
     renderComponent(AssetPortfolioLayoutNoPayouts),
   ),
+  withProps<IWithProps, IExternalProps>(props => ({
+    allowedToPayout: !props.isVerifiedInvestor,
```
AlexeyTuboltsev: looks like allowedToPayout means notAllowedToPayout. Something is wrong with the naming here?
I suggest to call it   !eligibleForPayout or smth like this

Issue #2395, opened 2019-02-01
[PORTFOLIO] please fix the view for smaller screens 


glupiawymowka:
On smaller screens, some elements overlap

To get this issue check portfolio on mobile or simply shrink the browser window

**Screenshots**
![image](https://user-images.githubusercontent.com/18315589/52121640-96a95700-2620-11e9-83da-65b74cb8732d.png)

glupiawymowka:
view fixed 

![image](https://user-images.githubusercontent.com/18315589/53807282-776b5580-3f4f-11e9-834f-4680d7ff9590.png)

PR #2449, opened 2019-02-12
#2395 Fix reserved assets columns width

drakkein:
For now this is the quickest and best solution.

snap


app/components/portfolio/PortfolioAssetAction.module.scss
```diff
@@ -0,0 +1,7 @@
+.investment-progress {
```
desfero: We don't need these styles. Now you basically hide another issue in the code because SVG should automatically take all available width.

Issue #2409, opened 2019-02-04
nEUR Step 1 - Bank Verification 


falinebambi:
**Background**
nEUR Step 1 - Bank Verification 
On the platform, the user can use both ETH or nEUR. nEUR can be obtained through Quintessence PLC by linking the profile on the platform with the user's bank account.
The bank account gets linked by sending 1€ to the Quintessence account as verification (user gets 1nEUR during this action)

**PRD** https://docs.google.com/document/d/1tNrQGkVPDdSbSDaoJMNevPfQQtLjXc8l8Z_MZYA8MYo/edit

**Zeplin**
https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/dashboard?seid=5c585eeb4a2d813589e7c55e

**Acceptance criterea**
- [x] nEUR and ETH balance widgets should be split in the Wallet tab (please mind the new widget names)
- [x] Linked Bank Account widget should appear in Profile tab
- [x] for unverified user, nEUR balance widget should contain an information that no bank account is verified 
- [x] for unverified user, Linked Bank Account widget should contain a button that starts verification process
- [x] for unverified user, clicking PURCHASE, REDEEM and VERIFY BANK ACCOUNT buttons should start bank account verification process
- [x] user should be only available to continue the process when he accepts both terms of use 
- [x] user should receive an email with the transaction summary matching the data on the screen (iban, reference number etc) when the user's transfer was confirmed and his account credited with 1nEUR, nEUR
- [x] Linked Bank Account widget and nEUR balance widget should change after successful account verification

**Testing notes**
check if the correct bank account nr is shown in Linked Bank Account widget
check if link different account link goes to new verification process

huesforalice:
To retrieve the current banking details of a user from the backend, there is an endpoint on the kyc api (you can check the swagger file). Right now it will only give you the last 4 digits of the account no we think you have. If we don't have anything on file, you will receive a 404. When my banking branch is merged, there will also be another field with the name of the account holder, which can be slightly different from what we have from the kyc. Please ask me directly if you have questions.

rudolfix:
## technical notes
**when to enter this flow**
REDEEM button is inactive if user does not have `isVerified` flag set in identityRegistry OR it has `isFrozen` flag set
When (1) user is verified AND (2) user has bank account in API (status 200) AND (3) user has `hasBankAccount` flag in identityRegistry we go to redeem flow (#2334)

[...]

PR #2459, opened 2019-02-14
#2409 Load bank details from api

desfero:


app/components/wallet/views/start/Start.tsx
```diff
@@ -120,7 +120,11 @@ export const WalletStartComponent: React.FunctionComponent<TProps> = ({
 
 export const WalletStart = compose<React.FunctionComponent>(
   onEnterAction({
-    actionCreator: dispatch => dispatch(actions.wallet.loadWalletData()),
+    actionCreator: dispatch => {
+      dispatch(actions.wallet.loadWalletData());
+
```
Moejoe90: is this new line of purpose?

app/lib/api/client/IHttpClient.ts
```diff
@@ -1,12 +1,14 @@
...
 import { Dictionary } from "../../../types";
+import { Schema } from "../../yup-ts";
 
 export type HttpMethod = "GET" | "POST" | "PUT" | "PATCH" | "DELETE";
 
 export interface IHttpRequestCommon {
   url: string;
   baseUrl?: string;
-  responseSchema?: Yup.Schema<any>;
+  responseSchema?: Yup.Schema<any> | Schema<any>;
```
Moejoe90: in the future we can think about changing this to a generic and get of `any`

desfero: hm, i'm just thinking whether we can get rid of `YupTs` in future and generate type information from `Yup.Schema` in this way `T extends Yup.Schema<infer R> ? R : never`

app/components/wallet/views/start/Start.tsx
desfero: yeah, this is on purpose as we will use in next pr data they are downloaded by this action

app/lib/api/client/IHttpClient.ts
Moejoe90: If can be done then perfect

app/components/wallet/views/start/Start.tsx
Moejoe90: I meant the empty line between them

desfero: ah, it's just for readability, do you think it's not needed?

app/modules/kyc/reducer.ts
```diff
@@ -54,6 +55,9 @@ export interface IKycState {
 
   // contract claims
   claims: TClaims | undefined;
+
+  // api bank details
+  bankAccount: TBankAccount | undefined;
```
Moejoe90: why not `bankAccount?: TBankAccount;`

app/modules/kyc/selectors.ts
```diff
@@ -117,3 +117,5 @@ export const selectIsAccountFrozen = createSelector(selectClaims, claims => {
 
   return false;
 });
+
+export const selectBankAccount = (state: IAppState) => state.kyc.bankAccount;
```
Moejoe90: can you add the return value here its better for readability

app/lib/api/client/IHttpClient.ts
desfero: if you can imagine then it can be done 🙏

app/lib/api/KycApi.ts
```diff
@@ -270,4 +269,23 @@ export class KycApi {
...
+      url: BANK_ACCOUNT_PATH,
+      responseSchema: KycBankAccountSchema,
+      allowedStatusCodes: [404],
+    });
+
+    if (response.statusCode === 404) {
+      throw new BankAccountNotFound();
+    }
+
+    return response;
```
Moejoe90: I think it's better to return the body here directly to get rid of 
`result.body`

app/modules/kyc/sagas.ts
```diff
@@ -578,6 +580,35 @@ function* submitBusinessRequest(
...
+  logger,
+  notificationCenter,
+}: TGlobalDependencies): Iterator<any> {
+  try {
+    const result: IHttpResponse<TKycBankAccount> = yield apiKycService.getBankAccount();
+
+    yield put(
+      actions.kyc.setBankAccountDetails({
+        hasBankAccount: true,
+        details: result.body,
```
Moejoe90: mentioned above

app/modules/kyc/reducer.ts
desfero: as this works in a different way in typescript. In reducer, i want to force this value to be set explicitly by a developer in initialState (see `kycInitialState`). If we use `?` then typescript will not warn us that property should be set in `kycInitialState`.

app/modules/kyc/selectors.ts
desfero: fixed

app/lib/api/KycApi.ts
desfero: fixed

app/modules/kyc/sagas.ts
desfero: fixed

app/modules/kyc/reducer.ts
Moejoe90: :+1:

PR #2463, opened 2019-02-14
Puts permissions into an enum

Moejoe90:
This PR got reduced to some small refactoring


app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -61,7 +61,7 @@ export function* deleteMyPledge(
   try {
     yield neuCall(
       ensurePermissionsArePresent,
-      [DO_BOOK_BUILDING],
+      [EJwtPermissions],
```
desfero: @Moejoe90 does `ensurePersimmsionsArePresent` support while enum as a first array item?

Moejoe90: good catch

PR #2469, opened 2019-02-15
#2409 NEUR Purchase

desfero:


app/utils/invariant.ts
```diff
@@ -2,8 +2,11 @@ export class InvariantError extends Error {
   type = "InvariantError";
 }
 
-export function invariant(value: any, message: string): any {
+export function invariant(value: any, message: string): null {
```
Moejoe90: if only the tsc gets a bit smarter when it comes to invariants it can be a really cool tool for type guards

app/config/.env.example
```diff
@@ -19,7 +19,6 @@ NF_CHECK_LOCKED_WALLET_WIDGET_ENABLED=1
 NF_WITHDRAW_ENABLED=1
 NF_GAS_PRICE_OVERHEAD=0.2 # adds 20% to estimated gas price
 NF_GAS_LIMIT_OVERHEAD=0.2 # adds 20% to estimated gas limit
-NF_HOURS_TO_DISABLE_BANK_TRANSFER_INVESTMENT=72 # default 3 days
```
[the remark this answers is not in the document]
desfero: but if you want to invest with NEUR then for sure you need to have this NEUR tokens available in your wallet otherwise investment is going to be blocked with this wallet

app/utils/invariant.ts
desfero: yeah, agree. I even tried to implement proper type guard `function invariant<T extends false | true>(value: T, message: string)): T extends false ? never : null` but this is not working yet

PR #2472, opened 2019-02-15
[BUGFIX] decouples token data state from investor ticket state

rudolfix:
When FORCE tokens are transferred, it's possible that token holder didn't participate in the ETO. currently such tokens will not be displayed in MY ASSETS because selector depends on investor ticket from ETO.

this PR disables this coupling however there are two things that I could not fix

1. now when you claim your tokens, do the Ethereum transaction and click VIEW portfolio, new token does not appear in MY ASSETS, you need to reload the page. There's some trickery I do not understand

2. When claiming tokens, I must press checkbox that I've read agreement. This dumps some huge react warning to the console.

please fix both before merging


app/modules/investor-portfolio/selectors.ts
```diff
@@ -51,15 +51,11 @@ export const selectEtoWithInvestorTickets = (
...
+export const selectMyAssets = (state: IAppState): TEtoWithCompanyAndContract[] | undefined => {
+  const etos = selectPublicEtos(state);
 
   if (etos) {
-    return etos.filter(
-      eto =>
-        eto.contract!.timedState === EETOStateOnChain.Claim ||
-        eto.contract!.timedState === EETOStateOnChain.Payout,
-    );
+    return etos.filter(eto => eto.contract && eto.contract.timedState === EETOStateOnChain.Payout);
```
drakkein: Since we are hiding tokens with 0 quantity there is no reason to select ETO's with `EETOStateOnChain.Claim` state.

app/components/portfolio/Portfolio.tsx
```diff
@@ -39,6 +40,21 @@ export const Portfolio = compose<TPortfolioLayoutProps, {}>(
       tokensDisbursal: selectTokensDisbursal(state),
       isVerifiedInvestor: selectIsVerifiedInvestor(state),
     }),
+    dispatchToProps: dispatch => ({
+      loadTokensData: (walletAddress: string) => {
+        dispatch(actions.publicEtos.loadTokensData(walletAddress));
+      },
+    }),
+  }),
+  lifecycle<any, any>({
```
desfero: `any` can be replaced by correct types

app/modules/investor-portfolio/selectors.ts
desfero: 👍

PR #2474, opened 2019-02-18
Ask to confirm navigation when form is dirty

desfero:


desfero:
closing for now, don't have time to find why danger is failing

PR #2499, opened 2019-02-22
At/document tiles - additional states

AlexeyTuboltsev:
fixes some usability issues with document tiles:
adds busy/loading state to tile
adds busy state to download link
makes investment agreement tile only active in signing state
forces investment agreement tile to be only active if there's no pending transaction


app/components/shared/Document.tsx
```diff
@@ -44,7 +53,7 @@ export const Document: React.FunctionComponent<IDocumentProps> = ({ extension, b
   return (
     <svg
       viewBox="0 0 61 73"
-      className={cn("document-icon", styles.document, !blank && computedExtension)}
+      className={cn("document-icon", styles.document, !blank && computedExtension, extraCss)}
```
drakkein: won't the @desfero magic work here, i mean this `{ [computedExtension]: !blank }`

app/components/shared/Document.tsx
```diff
@@ -67,69 +76,98 @@ export const DocumentTile: React.FunctionComponent<IDocumentProps & IDocumentTil
   blank,
   onlyDownload,
   active,
+  busy = false,
```
drakkein: I'm not sure here, shouldn't we set default values via `defaultProps` or `defaultProps()` from `recompose`? I mean this is correct from ES6 perspective, but in the react way we should use built-in mechanisms. Summoning @desfero and @Moejoe90

app/components/shared/Document.tsx
```diff
@@ -67,69 +76,98 @@ export const DocumentTile: React.FunctionComponent<IDocumentProps & IDocumentTil
...
 }) => {
   return (
-    <div className={cn(styles.tile, active && styles.active, !blank && styles.enabled, className)}>
-      <Document extension={extension} blank={blank} />
+    <div
+      className={cn(
+        styles.tile,
+        busy && styles.busy,
+        !blank && styles.enabled,
```
drakkein: The same way with @desfero magic `{ [styles.enabled]: !blank, [styles.active]: active }`

app/components/shared/Document.tsx
```diff
@@ -67,69 +76,98 @@ export const DocumentTile: React.FunctionComponent<IDocumentProps & IDocumentTil
...
+    >
+      <LoadingIndicator className={busy ? styles.documentBusy : styles.documentNotBusy} />
+
+      <Document extension={extension} blank={blank} extraCss={busy && styles.documentIconBusy} />
       <p
         className={cn(
           styles.title,
           blank && styles.blankTitle,
           !onlyDownload && !active && styles.disabledTitle,
+          busy && styles.textBusy,
```
drakkein: Also here

app/components/shared/Document.tsx
```diff
@@ -67,69 +76,98 @@ export const DocumentTile: React.FunctionComponent<IDocumentProps & IDocumentTil
...
         )}
       >
         {title}
       </p>
       {!onlyDownload &&
         blank &&
         active && (
-          <p className={cn(styles.subTitle)}>Drag and drop or Click to upload high quality PDF</p>
+          <p className={cn(styles.subTitle, busy && styles.textBusy)}>
```
drakkein: And here

PR #2516, opened 2019-02-26
At/#2470 skip resolving templates in on chain state

AlexeyTuboltsev:
when issuer tries to download templates from the document section and ETO state is ON_CHAIN, skip resolving templates and download them from IPFS directly

### DIFF
--- a/app/modules/eto-documents/sagas.ts
+++ b/app/modules/eto-documents/sagas.ts
@@ -3,16 +3,19 @@ import { call, fork, put, select } from "redux-saga/effects";
 
 import { EtoDocumentsMessage, IpfsMessage } from "../../components/translatedMessages/messages";
 import { createMessage } from "../../components/translatedMessages/utils";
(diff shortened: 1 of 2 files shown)


app/modules/eto-documents/sagas.ts
```diff
@@ -24,27 +27,32 @@ export function* generateDocumentFromTemplate(
...
+        },
+        {},
+      );
+    }
+
     const generatedDocument = yield apiImmutableStorage.getFile({
-      ipfsHash: templates.ipfs_hash,
-      mimeType: templates.mime_type,
-      placeholders: templates.placeholders,
+      ipfsHash: resolvedTemplate ? resolvedTemplate.ipfs_hash : document.ipfsHash,
```
rudolfix: isn't `document` the same type as `resolvedTemplate`? if so maybe you can override `document` reference with `resolvedTemplate` and you do not need this ternary operator here

PR #2519, opened 2019-02-26
#2409 Update bank verification flow to match latest requirements

desfero:


app/components/modals/bank-transfer-flow/purchase/BankTransferPurchaseSummary.tsx
```diff
@@ -72,6 +72,10 @@ const BankTransferPurchaseLayout: React.FunctionComponent<IProps> = ({
           />
         }
       />
+      <InfoRow
+        caption={<FormattedMessage id="bank-transfer.summary.purchase-price" />}
+        value="€1 = 1 nEUR"
```
AlexeyTuboltsev: maybe add it to translations too? (not sure... hardcoded string looks bad though)

desfero: hm, but there's nothing we can translate here?

PR #2520, opened 2019-02-26
Update ledger Provider

Moejoe90:
Fixes Ledger not connecting problem. 

To be done: 
Exchange timeout for Transport now is hardcoded for 30 seconds we either fork it and make a PR for them or patch it


banciur:
@Moejoe90 
I'm observing following chain of events when I'm running application locally
- I'm logging with my private ledger (updated software)
- Them I'm going to my profile and trying to perform email verification
- When there is step to confirm message on ledger I decline it.
- I see error
- Still blue toaster that e-mail was updated is being shown.

app/lib/web3/ledger-wallet/LedgerConnector.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+    const web3Adapter = new Web3Adapter(this.web3);
+
+    // note: in future we may want to solve locking issue on a lower level
+    const address = await web3Adapter.getAccountAddress();
+
+    return new LedgerWallet(web3Adapter, address, this.ledgerInstance, derivationPath);
+  }
+
+  public async getMultipleAccountsFromHdPath(
```
banciur: I really don't like those two getMultipleAccounts... functions. Probably during refactor we could get rid of that. But for now for sure it can stay.

app/lib/web3/ledger-wallet/errors.ts
```diff
@@ -0,0 +1,25 @@
...
+export class LedgerConfirmationRejectedError extends LedgerError {}
+export class LedgerContractsDisabledError extends LedgerError {}
+export class LedgerLockedError extends LedgerError {}
+export class LedgerNotAvailableError extends LedgerError {}
+export class LedgerNotSupportedVersionError extends LedgerError {}
+export class LedgerInvalidDerivationPathError extends LedgerError {}
+export class LedgerTimeoutError extends LedgerError {}
+export class LedgerUnknownError extends LedgerError {}
+
+export function parseLedgerError(error: any): LedgerError {
```
banciur: Here probably we also can live without those. But we should check if oldest ledgers return those errors and new ones different then this function will get complicated. But we should handle both errors old and new ones documented here https://github.com/Neufund/ledger-experiments#ledger-exceptions

app/modules/auth/jwt/sagas.ts
```diff
@@ -42,22 +42,22 @@ export async function obtainJwtPromise(
   const address = selectEthereumAddressWithChecksum(state);
 
   const salt = cryptoRandomString(64);
-
-  /* tslint:disable: no-useless-cast */
-  const signerType = web3Manager.personalWallet!.getSignerType();
-  /* tslint:enable: no-useless-cast */
+  if (!web3Manager.personalWallet) {
+    throw new Error();
```
banciur: here probably we should have more meaningful error. But we can live with this for now.

app/modules/wallet-selector/ledger-wizard/sagas.ts
```diff
@@ -39,40 +38,45 @@ export function* loadLedgerAccounts({
...
-  );
-
-  const balancesNEU: string[] = yield Promise.all(
-    derivationPathsArray.map(dp =>
-      contractsService.neumark.balanceOf(dp.address).then(bn => bn.toString()),
-    ),
-  );
+  try {
+    const derivationPathToAddressMap = advanced
+      ? yield ledgerWalletConnector.getMultipleAccountsFromDerivationPrefix(
```
banciur: `yield` magic. I believe you ;)

app/lib/web3/ledger-wallet/ledgerUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+};
+
+const getLedgerConfig = async (ledgerInstance: ILedgerCustomProvider): Promise<ILedgerConfig> => {
+  try {
+    const Transport = await ledgerInstance.getTransport();
+    const ethInstance = new Eth(Transport);
+    const configration = await ethInstance.getAppConfiguration();
+    Transport.close();
+    return configration;
+  } catch (e) {
```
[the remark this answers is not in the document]
Moejoe90: Yes correct

app/lib/web3/ledger-wallet/ledgerUtils.ts
```diff
@@ -0,0 +1,132 @@
...
+    throw new LedgerNotAvailableError();
+  }
+};
+
+export const createWeb3WithLedgerProvider = async (
```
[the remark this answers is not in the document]
Moejoe90: Take note here that `TransportU2F.create() ` is not created its only passed as a reference we don't need for anything at this point.

app/lib/web3/ledger-wallet/LedgerConnector.ts
Moejoe90: Yeah I agree with you I don't like them as well somehow they should be moved to the sagas instead of here anyway as they are platform specific functions IMO

app/lib/web3/ledger-wallet/errors.ts
Moejoe90: done new errors added

app/modules/wallet-selector/ledger-wizard/sagas.ts
Moejoe90: :)

app/modules/auth/jwt/sagas.ts
Moejoe90: This is done in a way that looks like an `invariant` its just a way to tell tsc that for sure personal data when it reaches this point `personalWallet` is already initialized

PR #2527, opened 2019-02-27
#2409 Get bank account from api / Inform backend about purchase request / Replace verification flow icons

desfero:


app/components/modals/bank-transfer-flow/purchase/BankTransferPurchaseSummary.tsx
```diff
@@ -78,15 +72,15 @@ const BankTransferPurchaseLayout: React.FunctionComponent<IProps> = ({
       />
       <InfoRow
         caption={<FormattedMessage id="bank-transfer.summary.recipient" />}
-        value={<CopyToClipboardLabel label={recipient} />}
+        value={<CopyToClipboardLabel label={quintessenceBankAccount!.name} />}
```
AlexeyTuboltsev: I suggest to rule out the quintessenceBankAccount === undefined in the appConnect, as a branch or as an invariant. Than you don't need a casting here

app/lib/api/KycApi.ts
```diff
@@ -283,4 +284,17 @@ export class KycApi {
 
     return response.body;
   }
+
+  public async nEurPurchaseRequest(amount: string, purpose: string): Promise<{}> {
+    const response = await this.httpClient.post<{}>({
+      baseUrl: BASE_PATH,
```
AlexeyTuboltsev: is it possible to type the response? Why is it a {} ?

app/modules/bank-transfer-flow/sagas.ts
```diff
@@ -111,8 +115,16 @@ export function* getRedeemData({ contractsService }: TGlobalDependencies): any {
   yield put(actions.bankTransferFlow.setRedeemData(bankFeeUlps, minEuroUlps));
 }
 
+export function* completeBankTransfer({ apiKycService }: TGlobalDependencies): any {
+  const reference: string = yield select(selectBankTransferFlowReference);
+
+  // TODO: replace by correct amount when implemented
+  apiKycService.nEurPurchaseRequest("1", reference);
```
AlexeyTuboltsev: shouldn't it have some error handling? it's a network call

app/lib/api/KycApi.ts
desfero: hm, the response is `{}` exactly :D

PR #2529, opened 2019-02-27
Fix redeem input behavior

drakkein:


app/components/modals/tx-sender/redeem/BankTransferRedeemInit.module.scss
```diff
@@ -19,3 +19,9 @@
   justify-content: space-between;
   margin-bottom: 1rem;
 }
+
+.invest-input {
```
desfero: don't think we need it

PR #2538, opened 2019-03-01
Bank transfer flow e2e tests

desfero:


app/modules/kyc/sagas.ts
```diff
@@ -649,20 +649,10 @@ export function* loadKycRequestData(): any {
 
   yield put(actions.kyc.kycLoadClaims());
 
-  // we block init until all requests are done. This avoids flickering of various elements in the app.
-  yield loadForOneOfTheKYCRequestsToLoad();
```
desfero: this one looks really strange to call function two times, replaced with `all` call.

app/modules/sagasUtils.ts
```diff
@@ -61,12 +61,18 @@ export function* neuTakeUntil(
...
  */
-export function* neuTakeOnly<T extends TActionType>(type: T, payload: TActionPayload<T>): any {
+export function* neuTakeOnly<T extends TActionType>(
+  type: T,
+  payload: Partial<TActionPayload<T>>,
+): any {
   while (true) {
     const takenAction = yield take(type);
-    if (isEqual(takenAction.payload, payload)) return takenAction;
+    if (isMatch(payload, takenAction.payload)) {
```
desfero: @Moejoe90 you are the original creator of `neuTakeOnly`, do you think it's good to use `isMatch` here over `isEqual`?

app/components/modals/bank-transfer-flow/verify/BankTransferVerifySummary.tsx
```diff
@@ -13,18 +12,15 @@ import {
...
 import { InfoRow } from "../../tx-sender/shared/InfoRow";
 
-import * as styles from "../../tx-sender/investment-flow/Summary.module.scss";
-
 interface IStateProps {
   referenceCode: string;
   minAmount: string;
-  quintessenceBankAccount: KycBankQuintessenceBankAccount | undefined;
+  quintessenceBankAccount: KycBankQuintessenceBankAccount;
 }
```
AlexeyTuboltsev: 👍

app/modules/kyc/sagas.ts
AlexeyTuboltsev: 👍 👍

Issue #974, opened 2018-07-02
[SECURITY ISSUE] when activating e-mail in context of another logged user I land in his dashboard


rudolfix:
I activated via following link
`https://platform.neufund.net/email-verify?code=629a56e5-d100-42e9-9806-596008103b86&email=rudolfix%2Bissuer%40possij.pl`
However I was logged into following account `rudolfix+reco@possij.pl`:
```
address
:
"0x429123b08df32b0006fd1f3b0ef893a8993802f3"
email
:
"rudolfix+reco@possij.pl"
salt
:
"dt4g+D4qCgT33S+ldVlVROrHGWfmc5MgAZq1m5IIc50="
vault
:
[...]

PR #2542, opened 2019-03-03
#974 Logout different user on email verification

drakkein:


app/index.tsx
```diff
@@ -1,3 +1,10 @@
+import "reflect-metadata";
+
+import "./components/translatedMessages/yupLocales.sideEffect";
+
+import "./styles/bootstrap.scss";
+
+// tslint:disable-next-line:ordered-imports
```
drakkein: These imports have to be in specific order to work correctly. Also the `tslint:disable:ordered-imports` is quite not intuitive https://github.com/palantir/tslint/issues/4372#issuecomment-446484566

desfero: 🎖

app/test-e2e/utils/index.ts
```diff
@@ -93,6 +93,14 @@ export const closeModal = () => {
   cy.get(tid("modal-close-button")).click();
 };
 
+export const getLatestVerifyUserEmailLink = () => {
+  return cy.request({ url: mockApiUrl + "sendgrid/session/mails", method: "GET" }).then(r => {
+    const activationLink = get(r, "body[0].personalizations[0].substitutions.-activationLink-");
+    // we need to replace the loginlink pointing to a remote destination with one pointing to our local instance
+    return activationLink.replace("platform.neufund.io", "localhost:9090");
```
desfero: hm, here i would prefer to make this just relative url (like just remove `https://platform.neufund.io` from url. As this will break tests when we change port number

app/test-e2e/verify/WalletVerify.spec.e2e.ts
```diff
@@ -26,4 +36,43 @@ describe("Invest with change", () => {
       });
     });
   });
+
+  it("should logout previous user when email activation occurs", () => {
+    const emailSeed = "wallet-e2e@test.com";
+    const email = convertToUniqueEmail(emailSeed);
```
desfero: maybe just `generateRandomEmailAddress`? Same in line 54

app/test-e2e/verify/WalletVerify.spec.e2e.ts
```diff
@@ -1,9 +1,19 @@
...
+  convertToUniqueEmail,
+  getLatestVerifyUserEmailLink,
+  registerWithLightWallet,
+} from "../utils/index";
 import { tid } from "../utils/selectors";
 import { createAndLoginNewUser } from "../utils/userHelpers";
 
-describe("Invest with change", () => {
-  it("do", () => {
+describe("Verify Wallet", () => {
```
desfero: 👍

app/test-e2e/verify/WalletVerify.spec.e2e.ts
```diff
@@ -26,4 +36,43 @@ describe("Invest with change", () => {
...
+      const newEmail = convertToUniqueEmail(emailSeed);
+      registerWithLightWallet(newEmail, password);
+      assertUserInDashboard();
+
+      assertUserInDashboard();
+
+      // try to activate previous user when second one is logged in
+      cy.visit(activationLink);
+
+      cy.get(tid("light-wallet-login-with-email-email-field")).then(activationEmailNode => {
```
desfero: `with-email-email-field`

desfero: plus we don't need `then` here, should replace with `should('eq'...`.
```cy.get(tid("light-wallet-login-with-email-email-field")).should('eq`, email)```

File: app/test-e2e/verify/WalletVerify.spec.e2e.ts [hunk 6]
```diff
@@ -26,4 +36,43 @@ describe("Invest with change", () => {
...
+        const activationEmail = activationEmailNode.text();
+        expect(activationEmail).to.equal(email);
+
+        cy.get(tid("light-wallet-login-with-email-password-field")).type(password);
```
desfero: maybe for consistency let's use `fillForm` util?

app/test-e2e/verify/WalletVerify.spec.e2e.ts
```diff
@@ -26,4 +36,43 @@ describe("Invest with change", () => {
...
+        cy.get(tid("wallet-selector-nuewallet.login-button")).click();
+
+        // email should be verified
+        assertVerifyEmailWidgetIsInVerfiedEmailState();
```
desfero: are not all these assertion mutual exclusive? so if `assertVerifyEmailWidgetIsInVerfiedEmailState` pass all below are not possible anyway?

app/test-e2e/verify/WalletVerify.spec.e2e.ts
drakkein: this is light wallet login > login with email > email field selector :D

app/test-e2e/verify/WalletVerify.spec.e2e.ts
drakkein: you are right there is no need for such of precise assertion here,
but `        cy.get(tid("profile.verify-email-widget.verified-email")).contains(email);` should stay in order to check if correct email has been validated

Issue #2487, opened 2019-02-21
hack nEUR wallet in investment flow


rudolfix:
**Background**
1. We want to run e2e test on nEUR investment flow
2. We want to run manual tests to actually use nEUR in your wallet (on io)
We do it before #2452 will not be visible on production (no ETOs)

**Designs**
We copy existing designs for LockedAccount nEUR investment. Check fixture user:
0x86894df4dae0eec4d7d13d08c32d92e614161790accdf1820981b45e6a74f07a

**Acceptance criteria**
1. User sees nEUR wallet in investment modal.
2. Can select wallet if nEUR balance > 0
3. Can follow identical investment flow as with Enabled ICBM Wallet nEUR
4. Sees NEU reward > 0
5. Sees gas costs (it's not zero contrary to direct investment)

**Technical notes**
Investment transaction is done via calling 3 parameter version of `transfer` function on EuroToken. Mind that typechain does not support overloads.
In gas checks and regarding gas - you are using your ETH wallet - same situation as LockedAccount investment

**Testing notes**
See acceptance criteria. Testing flow can include:
INV_HAS_EUR_HAS_KYC - already has nEUR in the wallet
any other fixture account: in that case you need a fake bank transaction to increase balance of nEUR first. it's end to end scenario

glupiawymowka:
- [x] option to invest with nEUR available only when nEUR balance > 0 
- [x] The progress bar is updated correctly 
![image](https://user-images.githubusercontent.com/18315589/54595622-50bf1b80-4a33-11e9-8ed4-6bd97cd5ccc3.png)
- [x] correct NEU reward counted 

TO DO
- [ ] no error message when I want to invest with nEUR and don't have ETH for GAS
![image](https://user-images.githubusercontent.com/18315589/54594786-aa264b00-4a31-11e9-9493-9b4dd97dbebe.png)

- [ ] My investment with nEUR is not recorded in Reserved Assets, but my investment with ETH updates there correctly. Here, more than 200k nEUR is invested together with~8 ETH. Only tokens bought with ETH are visible  
![image](https://user-images.githubusercontent.com/18315589/54595780-9d0a5b80-4a33-11e9-8613-2013b9e2e914.png)

glupiawymowka:
all fine on .io, still need to test "no ETH" scenario

glupiawymowka:
The "not enough ETH for GAS" case is also fixed 
[...]

PR #2554, opened 2019-03-05
#2487 add nEur investment type

desfero:


app/modules/investment-flow/reducer.ts
```diff
@@ -33,7 +29,7 @@ export const investmentFlowInitialState: IInvestmentFlowState = {
   etoId: "",
```
AlexeyTuboltsev: I know there are different opinions, but it looks sooo wrong to have invalid values in initial state. Or anywhere...

desfero: yeah, agree with you :(. It's on my todo list to get rid of empty strings in state

Issue #2414, opened 2019-02-04
Change email feature


Moejoe90:
**Background**
Change email widget is an already working. Now its time to finalize it. Maintaining separate widget does not make sense anymore. Merge it into `EMAIL VERIFICATION` widget
- [x] if there's verified e-mail, add button 'CHANGE EMAIL ADDRESS`
- [x] when clicked allow to enter new e-mail, we have state for it
![image](https://user-images.githubusercontent.com/17202864/54088326-59419300-435d-11e9-8203-3fd7655a5d0e.png)

- [x] If the `new email` equals `current email` verified and unverified show error toast component with user has the same email.
- [x] If entered email is unavailable return an error [ALREADY DONE]
- [x] If entered email is available add new email [ALREADY DONE]
- [x] change existing toaster text after adding e-mail from `Your email address has been successfully added.` to `Activation link was sent to your email.`
- [ ] cleanup code, remove change e-mail widget and `PERSONAL SETTINGS` section, remove setting to show e-mail widget `NF_FEATURE_EMAIL_CHANGE_ENABLED`
- [x] Add e2e tests for all flows

All calls are done via `user` api 

**Zepplin**
not needed
**Component**
VerifyEmailWidget
**Acceptance criteria**
* initial e-mail activation must work correctly and without any changes - we are adding a new state after it's completed
* flow above must work. user should be able to change -> enter -> activate e-mail
* user should be able to cancel verification ONLY if there's already verified e-mail
* in case of LightWallet, **login link must be sent to new e-mail**
* all e-mail templates must make sense so full test on stage is required

**Testing notes**

glupiawymowka:
- [ ] the buttons should be aligned (means, "update email address" should go lower)

![image](https://user-images.githubusercontent.com/18315589/55330470-6439a000-5491-11e9-8da3-472ac71a9d98.png)

- [ ] this is the email that I get when I change the address using light wallet. Things that are wrong here: 
- I didn't use the recovery phase, a separate template is needed here 
- The same sentence used twice (before each button) @vizmab please also take a look 

![image](https://user-images.githubusercontent.com/18315589/55330838-57697c00-5492-11e9-8a63-27f1c33166d2.png)

- layout used for buttons is different from the one we have when we register with light wallet 

![image](https://user-images.githubusercontent.com/18315589/55331231-1de54080-5493-11e9-8046-65d918f94d38.png)


[...]

PR #2565, opened 2019-03-06
#2414 Add email change tests

drakkein:
Added verification for email if it's not used already as verified/unverified email.

Waiting for @andywalton to confirm proper translations.


app/modules/auth/jwt/sagas.ts
```diff
@@ -104,7 +104,7 @@ export function* ensurePermissionsArePresentAndRunEffect(
     if (error instanceof MessageSignCancelledError) {
       logger.info("Signing Cancelled");
     } else {
-      throw new Error(error);
+      throw error;
```
drakkein: In this case we were wrapping `Error` into `Error` loosing his instance data, it should be throwed directly. So for example we were getting just `Error` instead of instance of `EmailAlreadyExists`.

app/config/.env.e2e
```diff
@@ -4,7 +4,7 @@ NF_REMOTE_NODE_PROXY_ROOT=http://localhost:8545
 
 NF_CONTRACT_ARTIFACTS_VERSION="localhost"
 
-NF_FEATURE_EMAIL_CHANGE_ENABLED=0
+NF_FEATURE_EMAIL_CHANGE_ENABLED=1
```
Moejoe90: I would say we don't need this flag anymore since the component got deleted 
Or?

app/modules/auth/jwt/sagas.ts
Moejoe90: :+1:

app/test-e2e/settings/ChangeEmail.spec.e2e.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    clearEmailServer();
+
+    registerWithLightWallet(email, DEFAULT_PASSWORD);
+    assertUserInDashboard();
+    verifyLatestUserEmail();
+
+    goToProfile();
+
+    cy.get(tid("verify-email-widget.change-email.button")).click();
```
Moejoe90: would be cool to have this as a `util` function 
For example `assertSomething`

Also do you know what happens if the assertion fails before the tests even start?

intl/locales/en-en.json
```diff
@@ -731,10 +730,12 @@
...
   "modules.auth.sagas.verify-user-email-promise.failed-email-verify": "We failed to verify your email.",
   "modules.settings.sagas.add-new-email.add-title": "Submit your email",
-  "modules.settings.sagas.add-new-email.confirm-description": "Please confirm that you would like to update your email.",
+  "modules.settings.sagas.add-new-email.confirm-description": "Your email address has been successfully added. Activation link was sent to your email.",
   "modules.settings.sagas.add-new-email.error": "Failed to send email",
-  "modules.settings.sagas.add-new-email.new-email-added": "Your email address has been successfully added.",
+  "modules.settings.sagas.add-new-email.new-email-added": "Your email address has been successfully added. Activation link was sent to your email.",
   "modules.settings.sagas.add-new-email.update-title": "Update email address",
+  "modules.settings.sagas.change-email.unverified-exists": "New email could not be the same as currently pending verification email.",
+  "modules.settings.sagas.change-email.verified-exists": "New email could not be the same as currently verified.",
```
Moejoe90: `New email could not be the same as currently verified.`
Should make a bit more sense

app/test-e2e/settings/ChangeEmail.spec.e2e.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    fillForm({
+      email: newEmail,
+      "verify-email-widget-form-submit": { type: "submit" },
+    });
+
+    acceptWallet();
+
+    // assert if new email is pending for verification
+    cy.get(notificationTid("profile-email-change-success")).should("exist");
```
Moejoe90: This can be a util function with `assertSomething`
These three assertions are used more than once

app/test-e2e/settings/ChangeEmail.spec.e2e.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    // fill again with the same email
+    fillForm({
+      email: newEmail,
+      "verify-email-widget-form-submit": { type: "submit" },
+    });
+
+    // assert if error message pooped in
+    cy.get(notificationTid("profile-email-change-unverified-exists")).should("exist");
+  });
```
Moejoe90: Please make util functions from all assertions this will make it easier for reusability

app/config/.env.e2e
drakkein: yup, you are right

intl/locales/en-en.json
drakkein: This will be the same message as for `modules.settings.sagas.change-email.verified-exists`

app/test-e2e/settings/ChangeEmail.spec.e2e.ts
drakkein: The test will fail if `beforeEach` fails.

intl/locales/en-en.json
Moejoe90: Do you mean it will be deleted or what do you mean?

PR #2567, opened 2019-03-07
Fix PersonalAccountDetails header icon

desfero:


app/components/settings/personal-account-details/PersonalAccountDetails.tsx
```diff
@@ -63,14 +62,7 @@ class AccountDetailsComponent extends React.Component<IStateProps, IOwnState> {
...
         headerText={<FormattedMessage id="settings.account-details.title" />}
-        rightComponent={
-          <ResponsiveImage
-            theme="transparent"
-            srcSet={{ "1x": personImage }}
-            alt="person image"
-            className={styles.image}
-          />
-        }
+        rightComponent={<img src={personImage} className={styles.image} alt="" />}
```
drakkein: can we not leave empty alt?

desfero: why do you want alt here? It's a representational icon that doesn't have any meaning (meaning was already provided by heading text that is on the right side). 'alt=""` should be used on 90% cases as often there is some more meaning text so there is no reason to bloat screen reader speech with `Account details info person image image` IMO. What do you think?

drakkein: To remove it completely or is it required by linting?

desfero: See https://davidwalsh.name/accessibility-tip-empty-alt-attributes. `img` without `alt` is not valid.

Issue #2560, opened 2019-03-05
REDEEM - redeem whole balance sets incorrect value


rudolfix:
**Description**
My balance is 29.083350986799034666 but when I use redeem whole balance, the value is set to rounded up wallet balance which is of course more than real balance

**Environment**
Production - 0x64Be85331949105Eb0459B936abF474710B4601b

**Actual behaviour**
Rounded up wallet balance gets copied

**Expected behavior**
What is expected is that after `withdraw` function is called on `EuroToken` contract, the balance is exactly 0. This means that in my case we should withdraw exactly 29.083350986799034666 - so wallet balance with full amount.
We should however do not scare people with such high precision numbers so unfortunately when you use this option you should show **wallet balance rounded down** but withdraw actual high precision balance!
In other words - we display human readable string but withdraw actual full precision balance

Once we do it, we'll apply this behaviour in other places - investment flow, ETH withdraw (this is not part of this ticket)

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5a72da804b5806bc2bc52ecc/f2cd985e-9e93-479c-88c8-9f41b554d443)

glupiawymowka:
*tested on .io*

Issue #2559, opened 2019-03-05
nEUR REDEEM - balance check incorrect


rudolfix:
**Description**
My actual holding of nEUR is 29.083350986799034666. Now I want to withdraw below that amount but application does not let me. Examples of amounts below that do not work:
29.080000001
29.083
29.081

**Environment**
Production - 0x64Be85331949105Eb0459B936abF474710B4601b

**Actual behaviour**
Hard to guess. It compares with something but hard to say what is this value or if we compare string or numbers

**Expected behavior**
All values below or equal actual balance with full precision should work. So in my case I should be fully able to withdraw 29.083350986799034666

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5a72da804b5806bc2bc52ecc/f177c802-ef3f-4d6d-9a3c-7d8053cda565)

Desperado:
tested on .io with other fixes.

PR #2569, opened 2019-03-07
#2560 #2601 Fix redeem whole nEUR balance

drakkein:
Right now balance is rounded down, and if user withdraws whole balance, the maximum precision number is passed to transaction instead of the rounded one.


app/modules/tx/transactions/redeem/saga.ts
```diff
@@ -34,9 +39,21 @@ export function* startNEuroRedeemGenerator(_: TGlobalDependencies): any {
...
+  const nEURBalanceUlps = yield select((state: IAppState) =>
+    selectLiquidEuroTokenBalance(state.wallet),
+  );
+  const nEURBalance = formatMoney(nEURBalanceUlps, MONEY_DECIMALS, 2, ERoundingMode.DOWN);
+  const selectedAmountUlps = convertToBigInt(txDataFromUser.value);
+
+  // Whole precision number should be passed when there is whole balance redeemed
+  const redeemAmountUlps =
+    compareBigNumbers(selectedAmountUlps, convertToBigInt(nEURBalance)) === 0
+      ? nEURBalanceUlps
```
AlexeyTuboltsev: here you convert value to BN with some precision, then to string  (formatMoney, convertToBigInt) and then back to a BN many times.
I suggest to simply use (new BigNumber(value)).toDigits() with given rounding mode and precision

app/modules/tx/transactions/redeem/saga.ts
```diff
@@ -1,22 +1,27 @@
...
 import { actions } from "../../../actions";
 import { selectStandardGasPriceWithOverHead } from "../../../gas/selectors";
 import { neuCall } from "../../../sagasUtils";
+import { selectLiquidEuroTokenBalance } from "../../../wallet/selectors";
 import { selectEthereumAddressWithChecksum } from "../../../web3/selectors";
 
 export function* generateNeuWithdrawTransaction(
   { contractsService, web3Manager }: TGlobalDependencies,
-  amount: string,
+  amountUlps: string,
```
AlexeyTuboltsev: is it really an Ulps? I'd rather leave it simply as amount
(https://neufund.slack.com/archives/C1X7PTD5Z/p1507214466000737)

drakkein:
#2559 is connected to this one

PR #2580, opened 2019-03-11
At/#2493 double signature request

AlexeyTuboltsev:
this pr
1. fixes the multiple sign request on metamask
2. adds a button to retry signing after rejecting
3. fixes retrying of connection approval (didn't work in the latest metamask)
4. adds one more sentence to the explanation what to do if metamask unavailable. It needs a graphic though:
Screen Shot 2019-03-11 at 15 11 01


falinebambi:
[Metamask_login_icons.zip](https://github.com/Neufund/platform-frontend/files/2953147/Metamask_login_icons.zip)
Screenshot 2019-03-11 at 17 27 04

app/components/wallet-selector/WalletMessageSigner.tsx
```diff
@@ -51,6 +53,14 @@ export const MessageSignerComponent: React.FunctionComponent<IStateProps & IDisp
         <LoadingIndicator className={styles.spinner} />
       )}
       <Row>
+        {!isLightWallet &&
```
desfero: I didn't check but what if we call `retrySigning` when trying to login with ledger? As above is only check for `isLightWallet` but `retrySigning` dispatch browser wallet specif action `tryConnectingWithBrowserWallet`

intl/locales/en-en.json
```diff
@@ -1016,6 +1017,7 @@
   "wallet-selector.browser.register-prompt": "Register with MetaMask",
   "wallet-selector.browser.steps.2": "1. Turn on the MetaMask extension in your browser.",
   "wallet-selector.browser.steps.3": "2. Login to your wallet.",
+  "wallet-selector.browser.steps.4": "2. Reload this page for changes to take effect.",
```
desfero: can you sync steps please. I don't see step 1 in keys and step 2 in values is repeated twice

app/components/wallet-selector/WalletMessageSigner.tsx
AlexeyTuboltsev: nice catch. I'll change it to be called only if walletType === browser for now. Need to investigate the ledger flow, will do it later

app/components/wallet-selector/WalletMessageSigner.tsx
```diff
@@ -51,6 +54,14 @@ export const MessageSignerComponent: React.FunctionComponent<IStateProps & IDisp
         <LoadingIndicator className={styles.spinner} />
       )}
       <Row>
+        {walletType === EWalletType.BROWSER &&
```
Moejoe90: This is slowly getting convoluted we probably should think of splitting to different components depending on wallet type
All components are viewed from here of course 

This is a future note

app/modules/wallet-selector/browser-wizard/reducer.spec.ts
```diff
@@ -28,7 +28,7 @@ describe("Wallet selector > Browser wizard > reducer", () => {
     );
 
     expect(state).to.be.deep.eq({
-      isLoading: true,
+      isLoading: false,
```
Moejoe90: Are you sure here the initial state is not `isLoading=true`?

app/modules/wallet-selector/browser-wizard/sagas.ts
```diff
@@ -1,28 +1,45 @@
...
+import { neuCall, neuTakeEvery } from "../../sagasUtils";
 import { mapBrowserWalletErrorToErrorMessage } from "./errors";
 
-export function* tryConnectingWithBrowserWallet({
-  browserWalletConnector,
-  web3Manager,
-  logger,
-}: TGlobalDependencies): any {
+export function* browserWalletConnectionWatcher(): any {
+  yield fork(neuCall, tryConnectingWithBrowserWallet);
```
Moejoe90: It's better to yield this directly as forking it will cause unexpected problems. It's a separate process
`yield neuCall(tryConnectingWithBrowserWallet)`

app/modules/wallet-selector/browser-wizard/sagas.ts
```diff
@@ -1,28 +1,45 @@
...
+
+  const connectionRetry = yield takeLatest(
```
Moejoe90: its better to use `try` `catch` instead of like this as it makes it easier to understand what happens when an error occurs 
also not all errors are bound by `BROWSER_WALLET_CONNECTION_ERROR`

app/modules/wallet-selector/browser-wizard/sagas.ts
```diff
@@ -1,28 +1,45 @@
...
+    "BROWSER_WALLET_CONNECTION_ERROR",
+    neuCall,
+    tryConnectingWithBrowserWallet,
+    BROWSER_WALLET_RECONNECT_INTERVAL,
+  );
+  yield take(["@@router/LOCATION_CHANGE", "WALLET_SELECTOR_CONNECTED"]);
+  yield cancel(connectionRetry);
```
Moejoe90: This won't stop all the sagas as `takeLatest` spawns not forks

app/modules/wallet-selector/browser-wizard/sagas.ts
```diff
@@ -1,28 +1,45 @@
...
+  yield cancel();
+}
```
Moejoe90: The best case here is to change the way this saga works into a while loop so that the whole process can be encapsulated into one process instead of forking and spawning different sagas

app/modules/wallet-selector/browser-wizard/sagas.ts
AlexeyTuboltsev: but it forks, see docs ))

app/modules/wallet-selector/browser-wizard/reducer.spec.ts
AlexeyTuboltsev: initial state is true, but I set it to false if signing is rejected. So it's ok here

app/modules/wallet-selector/browser-wizard/sagas.ts
AlexeyTuboltsev: solved

app/components/wallet-selector/WalletMessageSigner.tsx
AlexeyTuboltsev: I just have simplified it. It's looks very neat now

app/components/wallet-selector/WalletMessageSigner.tsx
AlexeyTuboltsev: ok this is all solved

app/modules/wallet-selector/browser-wizard/sagas.ts
Moejoe90: Ok then there is some misalignment in the documentation since it forks then it can be cancelled indeed

Desperado:
@AlexeyTuboltsev @Moejoe90 testing on .io with Opera Mobile and can't somehow see how can I sign any transaction. Everything is turned on. Page pops me up to reload the browser. Is it because of the proxy?

PR #2581, opened 2019-03-11
(2500 , 2597) - Automatically logout user when JWT expires

Moejoe90:
This also restarts all user related sagas after logout


app/utils/JWTUtils.ts
```diff
@@ -9,15 +9,25 @@ interface IJwt {
...
+/**
+ * Gets expiration date
+ */
+export function getJwtExpiryDate(token: string): Moment.Moment {
+  try {
+    const parsedJwt = parseJwt(token);
+    return Moment.unix(parsedJwt.exp);
+  } catch (e) {
+    throw new Error(`Cannot parse JWT token: ${token}`);
+  }
```
AlexeyTuboltsev: hm why would you catch it here? How would your handleJwtTimeout saga work in this case?

app/utils/JWTUtils.ts
```diff
@@ -9,15 +9,25 @@ interface IJwt {
...
  */
 export function isJwtExpiringLateEnough(token: string): boolean {
   try {
-    const parsedJwt = parseJwt(token);
-    const expirationDate = Moment.unix(parsedJwt.exp);
+    const expirationDate = getJwtExpiryDate(token);
     const expirationDiff = expirationDate.diff(Moment(), "minutes");
-
     return expirationDiff >= MAX_EXPIRATION_DIFF_MINUTES;
   } catch (e) {
```
AlexeyTuboltsev: why catch here? Looks dangerous ))

app/modules/auth/sagas.ts
```diff
@@ -93,10 +96,24 @@ function* verifyUserEmail(): Iterator<any> {
...
+/**
+ * JWT Timeout AutoLogout
+ */
+function* handleJwtTimeout(): Iterator<any> {
+  const jwt: string = yield select(selectJwt);
+
+  const ExpiryDate = getJwtExpiryDate(jwt);
+
+  // Automatically logout after ExpiryDate
+  yield delay(calculateTimeLeft(ExpiryDate, true) * 1000);
```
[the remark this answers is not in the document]
Moejoe90: but `delay` counts in milliseconds hence why i multiply

app/modules/auth/sagas.ts
```diff
@@ -93,10 +96,24 @@ function* verifyUserEmail(): Iterator<any> {
...
+  actions.auth.logout();
```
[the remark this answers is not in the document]
Moejoe90: good catch

app/modules/auth/sagas.ts
```diff
@@ -93,10 +96,24 @@ function* verifyUserEmail(): Iterator<any> {
   yield put(actions.routing.goToProfile());
 }
 
+/**
+ * JWT Timeout AutoLogout
+ */
+function* handleJwtTimeout(): Iterator<any> {
+  const jwt: string = yield select(selectJwt);
+
+  const ExpiryDate = getJwtExpiryDate(jwt);
```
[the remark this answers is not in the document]
Moejoe90: done

app/utils/JWTUtils.ts
Moejoe90: It catches it then throws again so i would say that is fine

app/utils/JWTUtils.ts
Moejoe90: Also here its fine IMO

app/modules/auth/sagas.ts
```diff
@@ -93,10 +96,24 @@ function* verifyUserEmail(): Iterator<any> {
   yield put(actions.routing.goToProfile());
 }
 
+/**
+ * JWT Timeout AutoLogout
+ */
+function* handleJwtTimeout(): Iterator<any> {
+  const jwt: string = yield select(selectJwt);
```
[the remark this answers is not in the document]
Moejoe90: good catch

app/modules/auth/sagas.ts
```diff
@@ -93,10 +97,28 @@ function* verifyUserEmail(): Iterator<any> {
...
+function* handleJwtTimeout({ logger }: TGlobalDependencies): Iterator<any> {
+  try {
+    const jwt: string | undefined = yield select(selectJwt);
+    if (!jwt) throw new JwtNotAvailable();
+    const expiryDate = getJwtExpiryDate(jwt);
+    // Automatically logout after ExpiryDate
+    yield delay(calculateTimeLeft(expiryDate, true) * 1000);
+    yield put(actions.auth.logout());
+  } catch (e) {
+    logger.error("Failed to Auto Handle JWT AutoLogout");
```
desfero: this will then be logged as `[undefined]` to sentry because `Error` is not provided

typings/modules.d.ts
```diff
@@ -25,3 +25,5 @@ declare module "react-to-print";
 
 declare module "qrcode";
 declare module "proper-url-join/lib";
+
+declare module "jsonwebtoken";
```
desfero: why not https://www.npmjs.com/package/@types/jsonwebtoken?

app/modules/auth/sagas.ts
```diff
@@ -93,10 +97,28 @@ function* verifyUserEmail(): Iterator<any> {
...
 }
 
+/**
+ * JWT Timeout AutoLogout
+ */
+function* handleJwtTimeout({ logger }: TGlobalDependencies): Iterator<any> {
+  try {
+    const jwt: string | undefined = yield select(selectJwt);
+    if (!jwt) throw new JwtNotAvailable();
+    const expiryDate = getJwtExpiryDate(jwt);
```
desfero: Do we refresh expiry date when user is using platform? Or it's a constant value that backend set during login process?

app/modules/sagas.ts
```diff
@@ -36,34 +37,35 @@ import { web3Sagas } from "./web3/sagas";
...
-    effects.fork(remoteFileSagas),
-    effects.fork(txValidatorSagasWatcher),
-    effects.fork(txTransactionsSagasWatcher),
-    effects.fork(gasApiSagas),
-    effects.fork(etoDocumentsSagas),
-    effects.fork(txMonitorSagas),
-    effects.fork(tokenPriceSagas),
-    effects.fork(investmentFlowSagas),
     effects.fork(routingSagas),
+    // Sagas that should be restarted immediately when logout occurs
```
desfero: @Moejoe90 maybe we even can restart whole root saga after logout?

app/modules/sagas.ts
```diff
@@ -36,34 +37,35 @@ import { web3Sagas } from "./web3/sagas";
...
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", bankTransferFlowSaga),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", kycSagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", investorTicketsSagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", profileSagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", web3Sagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", walletSagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", icbmWalletGetDataSagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", etoFlowSagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", immutableFileSagas),
+    effects.fork(neuRestartIf, "AUTH_LOGOUT", etoSagas),
```
desfero: Check please `watchEtoSetAction` and `watchEtosSetAction` they use `neuFork` which under the hood is based on `spawn`.  We should IMO replace in `neuFork` `spawn` by `fork` and then eto watching should be also cancelled

app/modules/auth/sagas.ts
```diff
@@ -93,10 +97,28 @@ function* verifyUserEmail(): Iterator<any> {
...
+    // Automatically logout after ExpiryDate
+    yield delay(calculateTimeLeft(expiryDate, true) * 1000);
```
desfero: Maybe we can refactor `calculateTimeLeft` to work on milliseconds as it's really weird that it returns time left in seconds because milliseconds is standard IMO

Moejoe90: You have a point. 
But i wouldn't do it in this PR

app/modules/sagas.ts
Moejoe90: Not a good idea as you have some sagas that should keep running regardless 
For example `routingSagas` or `initSagas`

app/utils/JWTUtils.ts
```diff
@@ -9,15 +9,25 @@ interface IJwt {
...
 
+/**
+ * Gets expiration date
+ */
+export function getJwtExpiryDate(token: string): Moment.Moment {
+  try {
+    const parsedJwt = parseJwt(token);
+    return Moment.unix(parsedJwt.exp);
+  } catch (e) {
+    throw new Error(`Cannot parse JWT token: ${token}`);
```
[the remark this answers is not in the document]
Moejoe90: I would say If its a token that can't be parsed then we should see how the token looks like 
In all cases the JWT can't be used 

Or what do you think?

typings/modules.d.ts
Moejoe90: Ah i didn't know there were types for this will change

app/modules/auth/sagas.ts
Moejoe90: We don't refresh as to refresh the expiry date you need to sign again. IMO this will create a complicated flow the best approach is to log the user out and then they can obtain a new JWT

app/modules/sagas.ts
Moejoe90: good point 
I changed `neuFork` to use fork instead there shouldn't be any problems but lets see. 
IMO spawn should never be used

app/modules/sagasUtils.ts
```diff
@@ -101,3 +101,18 @@ export function* neuRepeatIf(
     }
   }
 }
+
+/**
+ *  Executes the generator and restarts running job if a specific actions is fired
+ */
+export function* neuRestartIf(
```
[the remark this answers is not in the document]
Moejoe90: good idea

app/components/shared/utils.ts
```diff
@@ -34,3 +34,5 @@ export {
   weekdayUTC,
   calculateTimeLeft,
 };
+
+// TODO: Move whole file to general app utils folder
```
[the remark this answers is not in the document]
Moejoe90: :vulcan_salute:

app/modules/auth/sagas.ts
Moejoe90: done

PR #2582, opened 2019-03-11
[HOTFIX] Load reserved assets on view init

drakkein:


drakkein:
Something is wrong with this branch, each run different test is failing.

PR #2591, opened 2019-03-13
Fix SMALL INLINE Button when inside Panel

desfero:


app/components/shared/Panel.module.scss
```diff
@@ -1,8 +1,6 @@
 .panel {
   background-color: $neu-white-two;
   box-shadow: $shadow-bottom-black;
-  display: flex;
```
desfero: no needed

app/components/shared/Panel.module.scss
```diff
@@ -16,9 +14,3 @@
   justify-content: space-between;
   margin-bottom: $spacing-vertical-sm-1;
 }
-
-.content {
-  display: flex;
```
desfero: We should not force `flex` for children because in case there is some inline content (like `a` or `button`) it will cause unneeded side effects like button taking all available space because of `flex: 1 1 100%`.

desfero:
visual regression is ok

Issue #2501, opened 2019-02-22
If node fails, other requests also fail.


Desperado:
**Description**
When we have invalid JSON rpc response, we need to ignore it. Make sure that the process will retry the request.

**Environment**
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36

**Steps To Reproduce**
Sentry issue

**Actual behaviour**

**Comments, context, logs**
Error: Invalid JSON RPC response: ""
https://sentry.neusrv.net/sentry/production/issues/3052/

**Screenshots**

Moejoe90:
This Ticket will be split into two tickets 

- [ ] Add retry on RPC fail for Batched methods 

- [ ] Assert that if a web3 method is not using batch if yes inject a separate retry on rpc layer abstraction

- [ ]

pstroinski:
Will be fixed with the web3 upgrade

PR #2604, opened 2019-03-19
[WIP] - #2501 - Retries RPC call when node fails

Moejoe90:


app/lib/web3/Web3Batch/Web3Batch.ts
```diff
@@ -26,9 +29,19 @@ class Web3AutoExecuteBatch {
...
-
-    this.web3Batch.execute();
-    this.web3Batch = undefined;
+    for (let retries = 0; retries < NUMBER_OF_ALLOWED_RETIRES; retries++) {
+      try {
+        this.web3Batch.execute();
+        this.web3Batch = undefined;
+        return;
+      } catch (e) {
+        delay(FAILED_REQUEST_WAIT_TIME);
```
desfero: can we add a log here? Can be useful in some cases for debugging and monitoring how often retries are called
```
this.logger.info(`${retries} retry call to node with ${this.web3Batch.requests.length}` to batch);
```
and then move log from line 31 below line 35 so structure is consistent

app/lib/web3/Web3Batch/Web3Batch.ts
```diff
@@ -26,9 +29,19 @@ class Web3AutoExecuteBatch {
 
   private execute = () => {
     this.logger.info(`Number of web3 node rpc request batched: ${this.web3Batch.requests.length}`);
-
-    this.web3Batch.execute();
-    this.web3Batch = undefined;
+    for (let retries = 0; retries < NUMBER_OF_ALLOWED_RETIRES; retries++) {
```
desfero: can we add unit tests fo this?

app/lib/web3/Web3Batch/Web3Batch.ts
```diff
@@ -26,9 +29,19 @@ class Web3AutoExecuteBatch {
...
+        this.web3Batch.execute();
+        this.web3Batch = undefined;
+        return;
+      } catch (e) {
+        delay(FAILED_REQUEST_WAIT_TIME);
+        continue;
+      }
+    }
+    throw new NodeNotRespondingError(
+      `Failed to Connect to RPC Node after ${NUMBER_OF_ALLOWED_RETIRES} retries`,
```
desfero: hm, can we also log the original error in some way? Like move this to `catch` and add original error message?

Moejoe90:
This will get closed until further investigation

Issue #2586, opened 2019-03-12
Remove the graphic pattern from the background of the platform


Louisswryan:
Hi,

This is a change to be made across the board on both issuer and investor side.

Let's remove the graphic that sits towards the bottom of the background of the platform.

This has been signed off by all designers. We will be introducing more of our unique Neufund graphics into the platform and this background will make it too visually complex.

![Screenshot 2019-03-12 at 13.31.47.png](https://images.zenhubusercontent.com/5c5184c7df0733a30553c211/83137078-e7b3-44ce-86d3-b6d410ef490e)

glupiawymowka:
*tested on .io*
as agreed with Louis, this is only about the dashboard. 
The pattern on the landing page can stay

PR #2609, opened 2019-03-20
#2586 Remove footer pattern from dashboard

drakkein:


app/styles/global.scss
```diff
@@ -48,7 +48,6 @@ body {
   padding-left: $authorized-layout-horizontal-padding-sm;
   padding-right: $authorized-layout-horizontal-padding-sm;
   -webkit-overflow-scrolling: touch; // sass-lint:disable-line no-vendor-prefixes
-  background: url("../assets/img/footer-dots.png") repeat-x bottom;
```
desfero: pleas remote this image from assets and reduce bottom padding as there is no longer need for that big bottom padding

drakkein:
Dots should stay on Landing views.

PR #2613, opened 2019-03-20
At/#2493 double signature request

AlexeyTuboltsev:
this branch has already been merged once, but it needed more work to fix small css issues on mobile. 
It's literally 3 lines of code that have to be added


app/components/wallet-selector/browser/WalletBrowser.module.scss
```diff
@@ -5,7 +5,7 @@
   grid-auto-flow: column;
   grid-auto-columns: 1fr;
   align-items:  start;
-  max-width: calc(100% / 1.5);
+  max-width: 100%;
```
desfero: do we then need `max-width`, isn't it 100% by default?

app/components/wallet-selector/browser/WalletBrowser.module.scss
```diff
@@ -22,3 +22,9 @@
   margin-bottom: 1.5rem;
   text-align: center;
 }
+
+@media (min-width: 500px) {
```
desfero: can we use here `@include media-breakpoint-up(sm)` here for consistency? It equals to `min-width: 576` if it's okay. And also please put (if you agree) this under `.step-card-wrapper` that is defined on line `1`, IMO it's easier to read when css is grouped

desfero:
and please assign yourself to pull request 👍

app/components/wallet-selector/browser/WalletBrowser.module.scss
AlexeyTuboltsev: it's not. Or it doesn't work. That's why I had to add this

PR #2617, opened 2019-03-21
Add log level support for logger

desfero:


app/config/.env.example
```diff
@@ -32,3 +32,5 @@ NF_NEURO_WITHDRAW_ENABLED=0
 NF_SENTRY_DSN=
 # Uncomment when want to enable staging sentry logs
 # NF_SENTRY_DSN=https://1a01753f602c4bbcad39de77a992fc8d@sdsn.neufund.io/7
+
+NF_LOG_LEVEL=info
```
desfero: If you just want to disable logging of `info` during development just put `NF_LOG_LEVEL=warn` inside your local `.env`

PR #2622, opened 2019-03-21
Serve local build on the network

AlexeyTuboltsev:
this PR adds a package.json command to be able to serve your local dev build on the network. 
this can be particularly useful if you need to debug on mobile
don't forget to change the node url in Metamask too (e.g https://\<yourIp\>:9090/node)


webpack/loadAppEnv.js
```diff
@@ -34,6 +33,12 @@ module.exports = function loadAppEnv(processEnv) {
       console.error("not overriding NF_UNIVERSE_CONTRACT_ADDRESS");
     }
   }
-
+  // if we want to serve local dev build on the network
```
Moejoe90: Its better to write comments in the passive tense
But not blocking

Issue #1837, opened 2018-11-15
Remove deprecated React lifecycle hooks


desfero:
**Background:** 
Due to the async rendering feature that is going to be shipped with React 17 a couple of hooks are deprecated (`componentWillReceiveProps`, `componentWillMount` and `componentWillUpdate`). Please refer to docs (https://reactjs.org/docs/react-component.html) for details.

drakkein:
Automated tests are passing, but would be nice to verify if there is no any strange behaviour across platform pages.

PR #2635, opened 2019-03-25
fix double sign request 

AlexeyTuboltsev:
this is a small followup for #1837 
fix OnlyPublicRoute to only redirect on mount


app/components/shared/routing/OnlyPublicRoute.tsx
```diff
@@ -34,9 +29,9 @@ export const OnlyPublicRoute = compose<IComponentProps, RouteProps>(
       isAuthorized: selectIsAuthorized(s.auth),
     }),
   }),
-  branch<IStateProps>(
-    state => state.isAuthorized,
-    renderComponent(OnlyPublicRouteRedirectComponent),
-  ),
+  onEnterAction({
+    actionCreator: (dispatch, props) =>
+      props.isAuthorized && dispatch(actions.routing.goToDashboard()),
```
desfero: IMO when we want to do some side-effects it's better to use `if`. with `{ }` as this then easier to read. What do you think?

app/components/shared/routing/OnlyPublicRoute.tsx
```diff
@@ -34,9 +29,9 @@ export const OnlyPublicRoute = compose<IComponentProps, RouteProps>(
       isAuthorized: selectIsAuthorized(s.auth),
     }),
   }),
-  branch<IStateProps>(
-    state => state.isAuthorized,
-    renderComponent(OnlyPublicRouteRedirectComponent),
-  ),
+  onEnterAction({
+    actionCreator: (dispatch, props) =>
```
desfero: can we type `props` as now it's `any`?

app/components/shared/routing/OnlyPublicRoute.tsx
```diff
@@ -15,12 +16,6 @@ interface IComponentProps {
   component: React.ComponentType;
 }
 
-type TProps = RouteProps & IStateProps;
-
-const OnlyPublicRouteRedirectComponent: React.FunctionComponent<TProps> = props => {
```
desfero: What was the exact problem here? Seems like logic is still the same?

AlexeyTuboltsev: yeah I thought it is but it's not. We should only redirect on component mount and now it redirects on props change too. That's what I fixed with this pr

app/components/shared/routing/OnlyPublicRoute.tsx
AlexeyTuboltsev: I think you're right 😄

app/components/shared/routing/OnlyPublicRoute.tsx
drakkein: yeah, looks like you are right @AlexeyTuboltsev

PR #2657, opened 2019-03-28
generate random seed and address for e2e tests

AlexeyTuboltsev:
add an e2e utility function to generate random seed and address for test fixtures
add an cypress promise utility
fix WalletRecover.spec.e2e.ts


app/test-e2e/utils/cyPromise.ts
```diff
@@ -0,0 +1,2 @@
+//an easy way to use cy promise and be able to call async functions in then()
+export const cyPromise = () => cy.wrap(null);
```
desfero: 👍

app/test-e2e/wallets/WalletRecover.spec.e2e.ts
```diff
@@ -8,67 +8,46 @@ import {
...
-    "strike",
-  ];
-
-  const expectedGeneratedAddress = "0x429123b08DF32b0006fd1F3b0Ef893A8993802f3";
-
   it("should recover wallet from saved phrases", () => {
-    const email = "john-smith@example.com";
-    clearEmailServer();
+    cyPromise()
+      .then(async () => await generateSeedAndAddress("m/44'/60'/0'"))
```
desfero: do we need `async/await` here?

app/test-e2e/wallets/WalletRecover.spec.e2e.ts
```diff
@@ -8,67 +8,46 @@ import {
...
+      .then(({ seed: words, address }) => {
+        const expectedGeneratedAddress = address;
```
desfero: do we need this redundant assignment? Why just not rename `address: expectedGeneratedAddre` above?

app/test-e2e/utils/cyPromise.ts
desfero: btw, what do you think to simplify it to so we can use it like
```
cyWrapPromise(thisMethodReturnsPromise());
```

app/test-e2e/utils/generateSeedAndAddress.spec.node.ts
```diff
@@ -0,0 +1,12 @@
+import { expect } from "chai";
+
+import { generateSeedAndAddress } from "./generateSeedAndAddress";
+
+describe("generateSeedAndAddress", () => {
+  it("generates seed and password", async () => {
+    const { seed, address } = await generateSeedAndAddress("m/44'/60'/0'");
+
+    expect(seed.length).to.eq(24);
+    expect(address.length).to.eq(42);
```
Moejoe90: Also, we can check if it's an ethereum address

app/test-e2e/utils/generateSeedAndAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+  deserializeLightWalletVault,
+  getWalletKey,
+} from "../../lib/web3/light-wallet/LightWalletUtils";
+
+interface ISeedAndAddressGenerator {
+  seed: string[];
+  address: string;
+}
+
+export const generateSeedAndAddress = async (
```
Moejoe90: i suggest `generateRandomSeedAndAddress`

app/test-e2e/wallets/WalletRecover.spec.e2e.ts
Moejoe90: I think this depends on the use case but we can make `cyPromise` take an argument so that we use it in the context of 
`cyPromise(()=>await generateSeedAndAddress("m/44'/60'/0'"))).then{etc}`
What do you think

app/test-e2e/wallets/WalletRecover.spec.e2e.ts
```diff
@@ -8,67 +8,46 @@ import {
...
-    "idea",
-    "strike",
-  ];
-
-  const expectedGeneratedAddress = "0x429123b08DF32b0006fd1F3b0Ef893A8993802f3";
-
   it("should recover wallet from saved phrases", () => {
-    const email = "john-smith@example.com";
-    clearEmailServer();
+    cyPromise()
```
Moejoe90: :+1:

app/test-e2e/wallets/WalletRecover.spec.e2e.ts
Moejoe90: Yep its needed like that one of the wonders of cypress :*(

AlexeyTuboltsev: thank you moe :)

AlexeyTuboltsev: done, see my comment on Pawel's first comment

app/test-e2e/utils/cyPromise.ts
AlexeyTuboltsev: cy has it's own promise implementation, I couldn't get it to work without _.then_
so I can't do 
```
const result = cyWrapPromise(thisMethodReturnsPromise());
//do syncronous stuff with result
```
because `result` is `Cypress.Chainable<any>`

what I did is 
```
cyPromise(()=>someFn(arg)).then(
     //our test
)
```

app/test-e2e/utils/generateSeedAndAddress.spec.node.ts
Moejoe90: Done changed to `isAddress`

app/test-e2e/utils/cyPromise.ts
Moejoe90: Yes this is exactly how I envisioned it the `then` has to be there at least for now 
Maybe in future iterations, they will work on this there is also an async/wait library for cypress but only works with chrome as a browser so the idea never materialized i don't why it can't be added in the future though

app/modules/web3/utils.ts
```diff
@@ -30,7 +30,8 @@ export function ethereumNetworkIdToNetworkName(networkId: EthereumNetworkId): st
   }
 }
 
-export const validateAddress = (value: string) => value && isAddress(value.toUpperCase());
+export const validateAddress = (value: string): boolean =>
+  !!value && isAddress(value.toUpperCase());
```
Moejoe90: did typescript force `!!`? as i think it should already return a bool

app/test-e2e/utils/cyPromise.ts
```diff
@@ -1,2 +1,4 @@
-//an easy way to use cy promise and be able to call async functions in then()
-export const cyPromise = () => cy.wrap(null);
+// an easy (and recommended by cypress) way to use cy promise and be able to call async functions in then()
+// @see https://docs.cypress.io/api/utilities/promise.html#Waiting-for-Promises
+export const cyPromise = (fn?: Function) =>
```
Moejoe90: why is `fn` optional it shouldn't be the whole point is we want to resolve a promise in a chainable cypress way

Issue #2606, opened 2019-03-19
Issuer ETO forms - Reordering of Key Individual list items


andywalton:
**Background**
Currently there is a usability issue on the Issuer ETO forms. For example, when adding Key Individuals, you can reorder the individuals added to the various sections without a) editing the fields to correctly represent the order you want or b) deleting them and then adding back information in the order you want. This is tedious and frustrating.

A better solution would be to allow a user to reorder. The simplest solution is to use some reordering arrows up and down arrows.

**Designs**
See attached a screen design. Note to the right of the list item cards there would be a reordering up down arrow. When clicking either direction it should change the order, ie. either up or down.

_Please ask Kirstyn for the vector svgs for the arrow icons if they aren't attached to this ticket before you pick it up_.

**Acceptance criteria**
As a user, I can reorder the following items in the Issuer step Key Individuals sections:
* TEAM
* ADVISORS
* KEY ALLIANCES
* BOARD MEMBERS
* NOTABLE INVESTORS
* KEY CUSTOMERS
* PARTNERS

Reordering will be limited to inside the section. Ie. you won't be able to reorder an item from one section into another (Team to Advisors).

**Technical notes**
If you feel we should use drag and drop for reordering, let's chat then about specific additional designs you'd need. You'll probably need a placeholder card and a drop zone. But I'd recommend do the simplest and for now use the up down arrows to reorder.

**Testing notes**
After reordering. If you revisit the page the order should remain. If you reorder it should update immediately and shouldn't require a page save. The reason, that's how it is when you delete. It deletes an item immediately and you don't need to press save for it to finalise the deletion.

**Attachments**
![Screenshot 2019-03-19 at 17.33.21.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/d9190378-64c8-482c-9a50-5ea37bb206b8)

andywalton:
Hey @AlexeyTuboltsev - let me know if you want to talk through this before you start. 👍

shanile:
Tested and it's working

PR #2675, opened 2019-03-29
#2347 - #2606 reordering key individuals

### LINKED ISSUE #2347: Issuer ETO forms - bug when deleting Key individuals
When you delete a position in the key individuals form after 1 click on"x" the form reloads back to dashboard. There is no way to delete more positions in the form without going to dashboard first
image
.

AlexeyTuboltsev:
this PR 
- adds possibility to move keyIndividuals up and down
- fixes #2347


AlexeyTuboltsev:
FOR QA: please test it from #2347 too

AlexeyTuboltsev:
the .svg arrows are not done quite right yet, I'll swap them on Monday. They don't affect neither code nor css, so you can review now ;))

app/utils/generateKey.ts
```diff
@@ -0,0 +1,37 @@
+// code taken from https://github.com/zzarcon/react-smart-key.
```
Moejoe90: Why is this added it's not used anywhere in this PR?

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -232,4 +258,4 @@ class SocialProfilesEditorLayout extends React.Component<IProps & TFormikConnect
   }
 }
 
-export const SocialProfilesEditor = connect<IProps, any>(SocialProfilesEditorLayout);
+export const SocialProfilesEditor = connect<IProps>(SocialProfilesEditorLayout);
```
Moejoe90: Taking of `any` wouldn't solve the problem as its typed now by default any 
have you tried `connect<IProps,VALUES>` based on the component up it looks that values = `IProps & TFormikConnect`
Check
https://jaredpalmer.com/formik/docs/api/connect#type-signature

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -173,24 +175,48 @@ interface IState {
...
+    const {
+      formik: { values },
+      name,
+    } = this.props;
+    const {
+      formik: { values: oldValues },
+    } = prevProps;
+
+    if (!isEqual(get(oldValues, name), get(values, name))) {
+      this.setSelectedFields(values, name);
```
Moejoe90: I don't think its a good idea to call `this.setSelectedFields` from here 
as it has `setState` 
https://github.com/Neufund/platform-frontend/pull/2675/files?file-filters%5B%5D=.scss&file-filters%5B%5D=.tsx#diff-c733dbc8225e5508acd8dfd1d26b3e76R197

Is there no way out of this shouldn't formik update all values automatically 
Lets discuss this

app/components/eto/registration/pages/KeyIndividuals.tsx
```diff
@@ -50,68 +52,100 @@ interface IIndividual {
...
+          </legend>
+          <SocialProfilesEditor
+            profiles={SOCIAL_PROFILES_PERSON}
+            name={`${member}.socialChannels`}
+          />
+        </fieldset>
+      </FormHighlightGroup>
+      <div>
+        <ButtonIcon
+          onClick={() => swap(index, index - 1)}
```
desfero: can you please add `alt` prop to `ButtonIcon` and pass it back to `InlineIcon`? This is bad for accessibility to have icons without any text so we can add `Move Up` and `Move Down` alternative (`alt`) text

app/components/eto/registration/pages/KeyIndividuals.tsx
```diff
@@ -133,6 +167,17 @@ class KeyIndividualsGroupLayout extends React.Component<IKeyIndividualsGroup & T
...
 
+  shouldComponentUpdate(nextProps: IKeyIndividualsGroup & TFormikConnect): boolean {
+    const {
+      formik: { values },
+      name,
+    } = this.props;
+    const {
+      formik: { values: newValues },
+    } = nextProps;
+    return !isEqual(get(newValues, name), get(values, name));
```
desfero: For some reason this will not help, even when you always put `false` here it will still rerender everything. Just turn on react highlight updates feature to see that there are no benefits here

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -1,5 +1,7 @@
 import * as cn from "classnames";
 import { connect, FieldArray, getIn } from "formik";
+import { isEqual } from "lodash/fp";
+import get from "lodash/get";
```
desfero: move `get` to `lodash/fp` please

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -123,7 +125,7 @@ const SocialMediaTags: React.FunctionComponent<{
           <button
             data-test-id={`social-profiles.profile-button.${name}`}
             onClick={() => onClick(index)}
-            className={cn(Boolean(selectedFields[index]) && "is-selected", styles.tab)}
+            className={cn({ "is-selected": Boolean(selectedFields[index]) }, styles.tab)}
```
desfero: 🎖

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -176,29 +178,54 @@ interface IState {
...
+  setSelectedFields = (values: IProps & TFormikConnect, name: string) => {
+    const {
+      profiles,
+      formik: { setFieldValue },
+    } = this.props;
     const socialMediaValues: TSocialChannelsType = getIn(values, name) || [];
     const selectedFields: boolean[] = [];
 
-    profiles.forEach((profile, index) => {
+    const profilesWithUserData = profiles.map((profile, index) => {
```
desfero: 🎖

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -176,29 +178,54 @@ interface IState {
...
       const previousLink = socialMediaValues.find(v => v.type === profile.name);
       const value: string = previousLink ? previousLink.url || "" : "";
-      setFieldValue(`${name}.${index}`, { type: profile.name, url: value });
       selectedFields[index] = profile.preSelected ? true : !!value;
```
desfero: we should also refactor this, as doing side effects is not acceptable inside `map`

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -176,29 +178,54 @@ interface IState {
...
+    const {
+      formik: { values },
+      name,
+    } = this.props;
+    this.setSelectedFields(values, name);
   }
 
-  toggleProfileVisibility = (index: number): void => {
-    const { selectedFields } = this.state;
+  shouldComponentUpdate(nextProps: IProps & TFormikConnect, nextState: IState): boolean {
```
desfero: save with KeyIndividuals

app/components/shared/SocialProfilesEditor.tsx
```diff
@@ -176,29 +178,54 @@ interface IState {
...
+    return !isEqual(get(newValues, name), get(values, name)) || !isEqual(this.state, nextState);
+  }
 
-    this.setState({ selectedFields });
+  toggleProfileVisibility = (index: number): void => {
+    const {
+      formik: { setFieldValue },
+      name,
+    } = this.props;
+    const newSelectedFields = [...this.state.selectedFields];
```
desfero: hm, can you change `selectedFields` to ReadonlyArray? This is really bad to modify array and then push it to setState. Thanks for fixing it

desfero: btw, we should use function as a `setState` first arg because of `https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous`

app/components/shared/forms/FormSection.module.scss
```diff
@@ -1,24 +1,19 @@
 .form-section {
-  @include make-row;
+  display: grid;
+  grid-template: auto auto / auto;
+  grid-column-gap: 2rem;
+  width: 100%;
```
desfero: do we need `width: 100%`, how `display: grid` affect `width`? I always thought it's the same for container as `display: block`

app/components/shared/forms/FormSection.tsx
```diff
@@ -12,7 +12,7 @@ export const FormSection: React.FunctionComponent<IProps> = ({ title, children }
   return (
     <section className={styles.formSection}>
       <h3 className={styles.title}>{title}</h3>
-      <div className={styles.content}>{children}</div>
+      <div>{children}</div>
```
desfero: hm, that's the reason why i don't like fully using display grid inside css as then in react you will have a lot of plain `div` without any class. And this can be misunderstood by some of the developers and removed in future

app/components/eto/registration/pages/KeyIndividuals.tsx
```diff
@@ -50,68 +52,100 @@ interface IIndividual {
...
+  image: string;
+  key: string;
+}
+
 const getBlankMember = () => ({
   name: "",
   role: "",
   description: "",
   image: "",
+  key: Math.random().toString(),
```
Moejoe90: Math.random() doesn't guarantee the uniqueness of generated values

The probability is indeed low but there. This will introduce an extremely hard to find bug in the future

app/components/eto/registration/pages/KeyIndividuals.tsx
```diff
@@ -250,15 +299,56 @@ const EtoRegistrationKeyIndividuals = compose<React.FunctionComponent>(
...
   }),
 )(EtoRegistrationKeyIndividualsComponent);
 
 export { EtoRegistrationKeyIndividualsComponent, EtoRegistrationKeyIndividuals };
+
+const generateKeys = () => (data: { members: IMemberData[] }) => {
+  const members =
+    data.members &&
+    data.members.map(member => {
+      member.key = Math.random().toString();
```
Moejoe90: See comments up

app/utils/generateKey.ts
Moejoe90: Is this file used in the PR?

app/components/eto/registration/pages/KeyIndividuals.tsx
```diff
@@ -250,15 +299,56 @@ const EtoRegistrationKeyIndividuals = compose<React.FunctionComponent>(
     }),
     dispatchToProps: dispatch => ({
       saveData: (data: TPartialCompanyEtoData) => {
-        dispatch(actions.etoFlow.saveDataStart({ companyData: data, etoData: {} }));
+        const convertedData = convert(data, fromFormState);
```
Moejoe90: What is the exact reason for using `convert` is it just to get rid of the additional key added? 
If that's the case then I suggest creating a utility function that utilizes `convert` is really not clear that the only reason we are doing convert is to get rid of keys

app/components/eto/registration/pages/KeyIndividuals.tsx
```diff
@@ -250,15 +299,56 @@ const EtoRegistrationKeyIndividuals = compose<React.FunctionComponent>(
...
     validationSchema: EtoKeyIndividualsType.toYup(),
-    mapPropsToValues: props => props.stateValues,
+    mapPropsToValues: props => convert(props.stateValues, toFormState),
     handleSubmit: (values, props) => props.props.saveData(values),
   }),
 )(EtoRegistrationKeyIndividualsComponent);
 
 export { EtoRegistrationKeyIndividualsComponent, EtoRegistrationKeyIndividuals };
+
+const generateKeys = () => (data: { members: IMemberData[] }) => {
```
Moejoe90: Both `generateKeys`,`removeKeys` look like utility functions that shouldn't stay here IMO

app/components/shared/SocialProfilesEditor.tsx
AlexeyTuboltsev: solved

app/components/eto/registration/pages/KeyIndividuals.module.scss
```diff
@@ -6,6 +16,7 @@
 
 .add-button {
   padding-top: 0 !important; // sass-lint:disable-line no-important
+  padding-left: 0 !important; // sass-lint:disable-line no-important
```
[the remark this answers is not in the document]
AlexeyTuboltsev: yes, I agree. I don't like this either

app/components/eto/registration/pages/KeyIndividuals.module.scss
```diff
@@ -1,3 +1,13 @@
+.section-wrapper {
+  max-width: initial;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: I have to override the max-width=730px in app/components/eto/registration/Shared.module.scss .
I don't want to make any big refactoring here now. It's out of scope of this pr, easier to override it

app/components/shared/SocialProfilesEditor.tsx
AlexeyTuboltsev: but we don't derive this state from previous state, do we?

app/components/shared/SocialProfilesEditor.tsx
AlexeyTuboltsev: sure you're right, fixed

app/components/shared/forms/FormSection.module.scss
AlexeyTuboltsev: thank you, this one is a wip leftover too

app/components/shared/forms/FormSection.tsx
AlexeyTuboltsev: well they'll see immediately why they shouldn't 
i see your point but I don't know what to do here

app/utils/generateKey.ts
AlexeyTuboltsev: good catch, this is a wip leftover, sorry for this

app/components/eto/registration/pages/KeyIndividuals.tsx
Moejoe90: How do you know that the probability is zero? 

for sure we shouldn't rely on non-deterministic values 
Does key have size limitations? 
If not then why not hash `Math.random().toString()` with `UTC.timestap` we already have hashing functions in the bundle.

app/components/eto/registration/pages/KeyIndividuals.module.scss
```diff
@@ -1,3 +1,13 @@
...
+}
+
+.wrapper {
```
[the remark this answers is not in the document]
desfero: ok, let's then open a ticket

app/components/shared/SocialProfilesEditor.tsx
desfero: but we do `this.state` a lot in `toggleProfileVisibility`. Or maybe i don't understand something

app/components/shared/forms/FormSection.tsx
desfero: they will not see it if `children` provides single element. So the problem can appear after refactoring of `children` for e.g. to use `<>` then it will by committed to DOM as a couple of elements and this can be really hard to catch later

PR #2678, opened 2019-03-31
Fix storybook stories marked as unsafe

desfero:
Stories marked as unsafe where not included in storybook.


tslint.json
```diff
@@ -21,8 +21,14 @@
...
       true,
       {
-        "include": [".*.(tsx)"],
-        "exclude": [".*.(.spec.tsx)", ".*.(.util.tsx)", ".*.(.sideEffect.tsx)", ".*.(.unsafe.tsx)"]
+        "include": [".*.tsx"],
+        "exclude": [
+          ".*.spec.tsx",
+          ".*.util.tsx",
+          ".*.sideEffect.tsx",
+          ".*.stories.tsx",
```
Moejoe90: I don't think we should add this to the `exclusions` folder

PR #2679, opened 2019-03-31
Improve OnEnterAction/OnLeaveAction typings

desfero:


app/utils/OnEnterAction.tsx
```diff
@@ -1,41 +1,36 @@
...
+) =>
+  connect<{}, IOnEnterActionDispatchProps, P>(
     undefined,
     (dispatch, props) => ({
       enterAction: () => options.actionCreator(dispatch, props),
     }),
     undefined,
     {
-      pure: "pure" in options ? options.pure : true,
+      pure: options.pure,
```
Moejoe90: Here i can see the change in logic as it used to assume that `pure` should be `true` if `pure` was `undefined`

PR #2696, opened 2019-04-02
#2414 - Fix buttons position and wording

drakkein:
To have buttons correctly aligned in all possible variations, it requires to move them to bottom. 

Acceptance from @Louisswryan or @falinebambi is required

image

image


shanile:
Tested and working

Issue #2587, opened 2019-03-12
[Investor flow] please adjust investment module visually to the current needs


glupiawymowka:
**Background**
@falinebambi I'm adding you so you can update the designs as well. 
We agreed to drop all the € symbols to be consistent with all the  money/token values in the future, as some of them don't have symbols. Also, "Eth Wallet " and "nEuro Wallet" is not consistent with any name that we use on the platform. 

**Designs**
This whole view needs fixes. What is wrong: 
- header, "SELECT YOUR WALLET TO INVEST" => "Select which balances you’d like to use" (also capital)
- to be consistent : "Invest Your Entire Wallet Balance" => first, it's the normal sentence, no need to start every word with a capital letter, second, "Invest your entire balance"
- Eth Wallet & nEuro Wallet => ETH balance & nEUR balance
- The same here
![image](https://user-images.githubusercontent.com/18315589/54371741-39221480-467a-11e9-87ad-ccbb1d0626e5.png)

- min. amount is 0.43 **EUR** 
![image](https://user-images.githubusercontent.com/18315589/54204031-b52e2800-44d3-11e9-9c4d-227b746bbbe7.png)
- The € symbol in the input should be changed to EUR
- the same problem in the estimated gas footer 
![image](https://user-images.githubusercontent.com/18315589/54204525-ab58f480-44d4-11e9-90ee-bb95a0f931a2.png)

**Acceptance criteria**
- [x] "select your balance" header is updated
- [x] "Invest your entire balance" button is updated
- [ ] balances to choose are named correctly (ETH balance t + nEUR balance) 
- [ ] no € symbol to be seen in the investment module

**Testing notes**
Can be tested on .io with 
INV_HAS_EUR_HAS_KYC fixture
0x0009c1d95c547d53e3b962059be11802b5e85ba3

falinebambi:
@glupiawymowka We are updating designs for this in another ticket, will make sure that those above changes are applied.

falinebambi:
@andywalton can you make phrase app text changes please. @Louisswryan can you double check the copy against the new designs for the enable ICBM wallet ticket

Louisswryan:
This is how the investor flow appears now

Zeplin: https://zpl.io/VkGg6pZ

Every change must be made. They are listed above and below is a screenshot outlining them.

[...]

PR #2700, opened 2019-04-03
#2587 - Fix wording in investment flow

drakkein:


drakkein:
Gas symbols are fixed by #2698

PR #2718, opened 2019-04-06
Adds cypress-retry package

Moejoe90:
This removes our manual script in favour of a better module that should retry each test within `describe` instead of the full test in `describe` this will make our tests faster after a flaky test fails also makes it more stable


Moejoe90:
Will continue working on this in the future

Issue #2405, opened 2019-02-03
Past investments in Portfolio


rudolfix:
when tokens are claimed, investment line disappears from RESERVED ASSETS **forever**, we need to show those somewhere probably where history of transactions was supposed to be shown

See below

Desperado:
@rudolfix do we need some investigation for this, or it should be proper story?

andywalton:
We met and it was agreed we'll do the following:

1) In Wallet - we will have a platform history of transactions (we might need to create a Transaction detail view but we'll see about this when designing it and if we can delay this for simplicity for a later stage).
2) In Portfolio - we'll have a new Past Transactions list in chronological date order (it's basically a timeline feed)

See a working PRD here: https://docs.google.com/document/d/1ileQzHNis9RBHmZc6bHE2cF9kj7PCyUXAZEFYXSQpnY/edit?usp=sharing

falinebambi:
We are splitting this ticket into 2 parts. This ticket will concentrate only on the listing of the 'Past Investment' in the Portfolio view. This solves the problem of the Reserved Assets such as unsuccessful but refunded Tokens disappearing from record. 

The Past Investments are sorted by Date. 
If a field in not applicable leave it blank in the table 

Order of table: 
Date
Token 
Profile 
Quanity
Amount Invested 
Purchase Price 
Neu Reward
Status 

List of Status:
Claimed
Refunded 
No status 


![image.png](https://images.zenhubusercontent.com/57c01da2e40e5714b16dddf9/07ac1347-8bf5-4085-8e09-35788e15a132)

Zeplin link: https://zpl.io/boDnAmg

rudolfix:
**tech details**
this is basically different on the same data you have in RESERVED ASSETS with the following exceptions:
* ETO is in on chain state of REFUND, CLAIM or PAYOUT (so goes out of reserved assets, goes in here)
* we order those by date

Please also note "View Profile" in MY ASSETS, which is active only for Equity Token, not NEU  and leads to public profile of company

falinebambi:
@rudolfix updated to remove view profile on NEU

glupiawymowka:
*this was tested on .io and will be tested live again*

PR #2719, opened 2019-04-07
#2405 - Add Past investments

drakkein:


app/components/portfolio/PortfolioPastInvestments.tsx
```diff
@@ -0,0 +1,137 @@
...
+              const timedState = contract!.timedState;
+              const investmentDate = new Date(contract!.startOfStates[timedState]!);
+
+              return (
+                <NewTableRow
+                  key={etoId}
+                  data-test-id={`portfolio-reserved-asset-${etoId}`}
+                  cellLayout={ENewTableCellLayout.MIDDLE}
+                >
+                  <FormattedDate value={investmentDate} />
```
desfero: probably `FormattedDate` also supports timestamp so there is not to convert date

app/components/portfolio/PortfolioPastInvestments.tsx
```diff
@@ -0,0 +1,137 @@
...
+                      investorTicket.equivEurUlps,
+                    )}
+                    currency={ECurrency.EUR}
+                    currencySymbol={ECurrencySymbol.SYMBOL}
+                    format={EMoneyFormat.FLOAT}
+                    isPrice={true}
+                  />
+
+                  <Money
+                    data-test-id="portfolio-reserved-asset-neu-reward"
```
desfero: maybe we should replace all `tid` here to `portfolio-past-invesment`?

app/modules/investor-portfolio/selectors.ts
```diff
@@ -209,3 +210,28 @@ export const selectIsIncomingPayoutAvailable = (state: IAppState): boolean => {
...
+  const etos = selectEtoWithInvestorTickets(state);
+  if (etos) {
+    return etos
+      .filter(eto => eto.investorTicket.claimedOrRefunded)
+      .filter(eto => isPastInvestment(eto.contract!.timedState))
+      .sort((left, right) => {
+        const timedStateLeft = left.contract!.timedState;
+        const timedStateRight = right.contract!.timedState;
+
+        const investmentDateLeft = new Date(
```
desfero: This seems like a redundant conversion as `startOfStates` is already a date object.

app/modules/investor-portfolio/utils.ts
```diff
@@ -89,3 +90,14 @@ export const getTokenPrice = (equityTokenInt: BigNumber, equivEurUlps: BigNumber
   const equityToken = Q18.mul(equityTokenInt);
   return equivEurUlps.div(equityToken).toString();
 };
+
+export const isPastInvestment = (etoState: EETOStateOnChain) => {
```
desfero: maybe `export const isPastInvestment = includes([EETOStateOnChain.Payout, EETOStateOnChain.Refund, EETOStateOnChain.Claim]);` Where includes is from `lodash/fp`?

app/components/portfolio/PortfolioPastInvestments.tsx
```diff
@@ -0,0 +1,137 @@
+import * as cn from "classnames";
```
desfero: can we add this to storybook?

app/components/portfolio/PortfolioPastInvestments.tsx
```diff
@@ -0,0 +1,137 @@
...
+              return (
+                <NewTableRow
+                  key={etoId}
+                  data-test-id={`portfolio-reserved-asset-${etoId}`}
+                  cellLayout={ENewTableCellLayout.MIDDLE}
+                >
+                  <FormattedDate value={investmentDate} />
+                  <>
+                    <img src={equityTokenImage} alt="" className={cn("mr-2", styles.token)} />
+                    <span className={"d-inline-block"}>
```
desfero: do we need this additional `span` with `inline-block`?

Issue #2654, opened 2019-03-27
ETO company detail page changes


andywalton:
**Background**
As part of the Neufund multi-jurisdiction ETO products epic we need to make changes to the ETO company pages as new Issuer fields exist. For more information read here #2546 

This ticket is about updating the ETO company page.

---

**Tasks**

### Task 1: Update / edit company page fields

**1. Section LEGAL INFORMATION**

<img width="642" alt="Screenshot 2019-03-29 at 12 39 49" src="https://user-images.githubusercontent.com/609657/55237044-846b2400-5231-11e9-9e77-3d236a8c3018.png">

- [x]  Change label NUMBER OF EMPLOYEES to read TEAM SIZE 

**2. Section ETO INVESTMENT TERMS / EQUITY TOKEN SALE**

<img width="764" alt="Screenshot 2019-03-29 at 12 07 24" src="https://user-images.githubusercontent.com/609657/55237069-90ef7c80-5231-11e9-9497-1f2e50abd125.png">

Add a new few field:
- [x]  ETO jurisdiction (Germany / Lichtenstein) <-- Add at the bottom of the current list

**3. Section ETO INVESTMENT TERMS / TOKEN HOLDER RIGHTS**

<img width="844" alt="Screenshot 2019-03-29 at 12 07 12" src="https://user-images.githubusercontent.com/609657/55237597-d1032f00-5232-11e9-9ece-db3fbbd0341e.png">

- [x] Token tradable (Immediately after ETO / In the future) <-- 3rd item - replaces Token transferability
- [x] Token transferable (Yes / No) <-- 4th item
- [x] Asset type (Security / Investment asset) <-- 5th item
- [x] Dividend rights (Yes -- but it's always yes FYI) <-- 7th item (after Voting rights)

---

### Task 2: Highlight the ETO jurisdiction on the page

Implement as per the screen designs. **@falinebambi please add the designs and a Zeplin link here and a screenshot.**

---

**Acceptance criteria & testing notes**
- [x] ETO company pages include the new fields as described above and in the designs.
- [x] The ETO company pages are linked to from the Dashboard ETOs list. They change the URL as per the pattern above. 
- [ ] The legal jurisdiction disclaimer popup should display in situations as per ticket #2655 
- [ ] The legal jurisdiction of the ETO is highlighted on the page as per the designs.

**Technical notes**
None.

falinebambi:
Zeplin link: https://zpl.io/beDXDAq

[...]

PR #2736, opened 2019-04-09
#2654 - Update company details page

drakkein:


app/components/eto/public-view/EtoInvestmentTermsWidget.tsx
```diff
@@ -242,6 +243,29 @@ const EtoInvestmentTermsWidgetLayout: React.FunctionComponent<TExternalProps & T
...
               />
             )}
+            {etoData.product &&
+              !!etoData.product.jurisdiction && (
+                <Entry
+                  label={
+                    <FormattedMessage id="eto.public-view.token-terms.public-eto.product.jurisdiction" />
+                  }
+                  value={
+                    <>
```
Moejoe90: This would be good to be moved to a component that takes Jurisdiction and returns the correct message based on that. But can be done in the future since there will only be two in the near future

Issue #2570, opened 2019-03-07
[REDEEM] please add a tooltip to bank transaction fee


glupiawymowka:
**Background**
We have a bank transaction fee, but it's nowhere said how big it is and why.
Let's add a tooltip explaining it to the user. 

**Designs**

This is how the popup looks atm, we could add a tooltip as in some other places, or design it differently 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/0af2bba6-3809-43c1-8470-c0faa9a9bb8d)

**Wording**

When you redeem nEUR, a 0.5% transaction fee is charged by our banking partner. This fee will be deducted from the amount you've requested to redeem.

falinebambi:
![Redeem](https://user-images.githubusercontent.com/17235402/54127309-c6aafd80-4409-11e9-924f-9dac8ad2ccd9.png)

Updated Design: https://zpl.io/bLKkY4YThis is also a universal styling change to Tooltips which will now be signified by a dotted underline. 

We will follow Material Design logic for the behaviour or tooltips.  
https://material.io/design/components/tooltips.html#behavior
![Screenshot 2019-03-11 at 14 29 17](https://user-images.githubusercontent.com/17235402/54127502-34efc000-440a-11e9-9870-b21e9cfb99df.png)

glupiawymowka:
*tested on stage*

PR #2740, opened 2019-04-10
#2570 - Add Bank transfer fee tooltip

drakkein:


app/components/modals/tx-sender/redeem/BankTransferRedeemInit.unsafe.tsx
```diff
@@ -157,10 +158,27 @@ const BankTransferRedeemLayout: React.FunctionComponent<IProps> = ({
...
+                        new BigNumber(bankFee).mul(100), // to get percentage value we have to multiply ba 100
+                        ECurrency.EUR,
+                        EMoneyFormat.ULPS,
+                        false,
+                        ERoundingMode.HALF_UP,
+                      ),
+                    }}
+                  />
+                }
+                preventDefault={false}
```
desfero: Do you need prevent default here?

app/components/modals/tx-sender/redeem/BankTransferRedeemInit.unsafe.tsx
```diff
@@ -157,10 +158,27 @@ const BankTransferRedeemLayout: React.FunctionComponent<IProps> = ({
...
-              <Heading level={3} decorator={false} size={EHeadingSize.SMALL}>
-                <FormattedMessage id="bank-transfer.redeem.init.redeem-fee" />
-              </Heading>
-
+              <Tooltip
+                content={
+                  <FormattedMessage
+                    id="bank-transfer.redeem.init.redeem-fee.tooltip"
+                    values={{
+                      fee: getFormattedMoney(
```
desfero: can we use our `Money` component here?

app/components/modals/tx-sender/redeem/BankTransferRedeemInit.unsafe.tsx
```diff
@@ -157,10 +158,27 @@ const BankTransferRedeemLayout: React.FunctionComponent<IProps> = ({
...
+                        new BigNumber(bankFee).mul(100), // to get percentage value we have to multiply ba 100
```
desfero: hm, what's the real reason to multiply? It's not clear from the comment in this case. As if you see the message it says `{fee}%` so what's the reason to use `getFormattedMoney`?

app/components/modals/tx-sender/redeem/BankTransferRedeemInit.unsafe.tsx
drakkein: It's not working. Element is not added to translated string :(

app/components/modals/tx-sender/redeem/BankTransferRedeemInit.unsafe.tsx
drakkein: Nope :D

app/components/modals/tx-sender/redeem/BankTransferRedeemInit.unsafe.tsx
drakkein: Added more detailed comment

Issue #2709, opened 2019-04-04
when pending e-mail activation link is invalid/expired - no error message during activation attempt


rudolfix:
**Description**
I have expired (see later) activation link. When I click on it in the e-mail I go to platform site and log in and I'm getting immediately logged out without any toaster or information.

**Environment**
Stagin (and probably all others)

**Steps To Reproduce**
1. create new account (any wallet type) and activate e-mail
2. change e-mail, you should get email with e-mail activation button
3. without activating change, change e-mail again. you will receive another e-mail. ignore it
4. find previous e-mail and click activate link in incognito window
5. log in and you get logged out

**Actual behaviour**
see above

**Expected behavior**
1. user should see activation status (that there was no pending activation or that link was invalid and activation was not possible)
2. we should keep user logged in

Moejoe90:
@rudolfix  If you look at ticket 974 the solution to log out the user was done on the bases that we assume if not the same email then it's not the same user. During that time the case where a user can change their email midway wasn't thought of

I think to solve this we need user id in the activation link so that we can assert if its the same user or not
https://github.com/Neufund/platform-frontend/issues/974

Moejoe90:
As per our discussion during grooming 


**Expected behaviour**
We should log the user out and display a toast component with a correct error message

rudolfix:
You could use following text "Your activation code is invalid. Make sure you used activation link from the most recent e-mail"

glupiawymowka:
This works correctly if the link of another user/account was taken. 

Unfortunately, the page crashes when Metamask + the old link for the same account was used.

**Steps to reproduce**
1. Login with metamask (you can use any fixture or empty account)
2. verify/update email address (with the link)
3. Choose to update email address to another one, but don't use the new link
4. Paste the link to the previous verification

**Actual result**
- The user will be logged out with the message "Your activation code is invalid. Make sure you used activation link from the most recent e-mail"
- Metamask will ask for a signature to login back. If the user signs it, another couple of requests will be sent, then redirect the user to the dashboard, then after some more clicks will finally log him in 

**Expected result**
From my understanding, the user should not be logged out 

> user should see activation status (that there was no pending activation or that link was invalid and activation was not possible)
>we should keep user logged in

**OR** 

> We should log the user out and display a toast component with a correct error message

but without Metamask requests that appear later 

**Comments**
I've recorded the whole process but it's too big to upload it here, please ask me if you'd like to see it

PR #2743, opened 2019-04-10
#2709 - shows notification bar with error message when email activation link is wrong

Moejoe90:
Also refactors the toast component to use `cy.get` assertions


app/test-e2e/link-triggered-flows/WalletVerify.spec.e2e.ts
```diff
@@ -62,6 +62,12 @@ describe("Verify Wallet", () => {
 
       cy.get(tid("light-wallet-login-with-email-email-field")).contains(email);
 
+      // Asserts if error toast shows up
+      // @SEE https://github.com/Neufund/platform-frontend/issues/2709
+      cy.get(tid("modules.auth.sagas.verify-user-email.toast.verification-failed"));
```
desfero: can we use additionally `should('exist`)` as it's more appropriate and easy to figure out that it should exist

app/test-e2e/settings/ChangeEmail.spec.e2e.ts
```diff
@@ -67,8 +66,8 @@ describe("Change Email", () => {
 
     acceptWallet();
 
-    // assert if error message has pooped in
-    assertNotificationExists("profile-email-exists");
+    // assert if error message has popped in
+    cy.get("profile-email-exists");
```
desfero: same here and on line 80 and 106. And please use `tid`

app/test-e2e/utils/assertions.ts
```diff
@@ -135,12 +135,13 @@ export const assertEmailChangeFlow = (): void => {
...
 };
 
+// @deprecated us cy.get()directly
 export const assertNotificationExists = (selector: string): void => {
   cy.get(notificationTid(selector)).should("exist");
 };
 
 export const assertEmailPendingChange = (email: string, newEmail: string): void => {
-  assertNotificationExists("profile-email-change-success");
+  cy.get("profile-email-change-success");
```
desfero: `should('exist')`

app/components/shared/Toast.tsx
```diff
@@ -8,15 +8,20 @@ import "react-toastify/dist/ReactToastify.css";
 
 const ToastContainer = () => <ToastifyContainer autoClose={TOAST_COMPONENT_DELAY} />;
 
-const getOptions = (options?: TDataTestId) =>
-  IS_CYPRESS && options ? { className: options["data-test-id"] } : undefined;
+const ComponentWithTestId = (children: TTranslatedString, options?: TDataTestId) =>
```
desfero: do you think we can refactor this component to:

```
getComponentWIthTestId = (children: ..., dataTestId?: string) => ...
```

app/test-e2e/utils/assertions.ts
Moejoe90: :+1:

Moejoe90:
This couldn't be tested with commit deployment testing because you need to use a link

Issue #2663, opened 2019-03-29
Terms of use change


andywalton:
Don't start on this ticket until it's assigned to you and the blocked label is removed. If in doubt, please ask Andy. 👍  
---

**Background**

The day we go live with the first multi-jurisdiction ETO offering (est. 15 May 2019) we'll need to have our _Terms of use_ online with that release.

This ticket's primary task is to update our Terms of Use that we display in app via a modal when you login to the app (first time) and each time we change our terms and you login the next time thereafter.

* Please ask Andy or Dialika for the latest terms pdf.

<img width="1260" alt="Screenshot 2019-03-29 at 11 26 24" src="https://user-images.githubusercontent.com/609657/55226635-bd959b00-5215-11e9-8413-25e04121d2dd.png">

These terms are also visible on our website via a link in the footer of the site. The Marketing Team will be responsible for getting this file online but feel free to help if you can.

<img width="1469" alt="Screenshot 2019-03-29 at 11 33 39" src="https://user-images.githubusercontent.com/609657/55227885-e9fee680-5218-11e9-838f-286d812d1bf1.png">

**Acceptance criteria & Testing notes**
- [ ] The new Terms of use are downloadable via the Please read our terms modal. They display for all new customers and will display again to existing customers when you login again after the terms have be updated. (If not Moe, see @Moejoe90 comment below).
- [ ] Please check with Marketing if they've updated the terms on our website.

Moejoe90:
@andywalton 
this is taken directly from the smart contract 
from frontend we take the current hash agreement from the smart contract and check if the user signed this hash

andywalton:
@Moejoe90 --> thanks for the extra info!

falinebambi:
![image](https://user-images.githubusercontent.com/17235402/56123799-82190180-5f75-11e9-8044-f9ae9563c8a5.png)

glupiawymowka:
For the new users that did not see any other ToU than the updated ones, please keep the new design and translations for buttons, but use the following header and paragraph: 



> PLEASE READ OUR TERMS OF USE
> 
> We would like you to read and confirm that you agree with our Terms of Use. This is an important step for you to understand our product.

glupiawymowka:
Please use a better image quality so it won't look like that when enlarged
![image](https://user-images.githubusercontent.com/18315589/56904363-64cc5300-6a9e-11e9-8185-15a852695ae0.png)

glupiawymowka:
[...]

PR #2774, opened 2019-04-17
#2663 - Update TOS modal

drakkein:


app/modules/auth/selectors.ts
```diff
@@ -62,7 +62,7 @@ export const selectIsVerifiedInvestor = createSelector(
 // TOS Related Selectors
 
 export const selectIsAgreementAccepted = (state: IAppState): boolean =>
-  !!(state.auth.user && state.auth.user.latestAcceptedTosIpfs !== undefined);
+  !!(state.auth.user && !!state.auth.user.latestAcceptedTosIpfs);
```
desfero: do we need `!!` here? There is `!!` at the begging so it should be enought

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -63,7 +77,11 @@ export const AcceptTosModalInner: React.ComponentType<IStateProps & IDispatchPro
         </Button>
       </div>
       <div>
-        <Button onClick={() => onLogout(userType)} layout={EButtonLayout.SIMPLE}>
+        <Button
+          onClick={() => onLogout(userType)}
+          layout={EButtonLayout.INLINE}
+          size={ButtonSize.SMALL}
```
desfero: This seems strange to provide size for `INLINE` button as this should mimic the look and feel of normal anchor tag

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -34,21 +40,29 @@ export const AcceptTosModalInner: React.ComponentType<IStateProps & IDispatchPro
 }) => {
   return (
     <section className="text-center">
-      <h1>
+      <img src={tosImg} className={styles.image} />
+      <Heading level={2} decorator={false}>
         {agreementChanged ? (
-          <FormattedHTMLMessage tagName="span" id="settings.modal.accept-updated-tos.title" />
```
desfero: 👍

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -34,21 +40,29 @@ export const AcceptTosModalInner: React.ComponentType<IStateProps & IDispatchPro
 }) => {
   return (
     <section className="text-center">
-      <h1>
+      <img src={tosImg} className={styles.image} />
+      <Heading level={2} decorator={false}>
```
desfero: 👍

app/components/modals/accept-tos-modal/AcceptTosModal.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  margin: {
+    top: px-to-rem(10px);
+    bottom: px-to-rem(16px);
+    left: auto;
+    right: auto;
+  }
+}
+
+.download {
+  @include font-label;
```
desfero: hm, we should not change the font size for buttons. Maybe we can reuse of the layout we already have? Or we need to talk with designers

app/components/modals/accept-tos-modal/AcceptTosModal.module.scss
```diff
@@ -0,0 +1,22 @@
+.image {
+  width: 270px;
+  margin-bottom: px-to-rem(28px);
+}
+
+.text {
+  @include font-label;
```
desfero: we should not change the size here as we have a bug with overall paragraph sizes. If you look at zeplin they all are set to 12px so this should be a global change IMO

app/components/modals/accept-tos-modal/AcceptTosModal.tsx
```diff
@@ -34,21 +40,29 @@ export const AcceptTosModalInner: React.ComponentType<IStateProps & IDispatchPro
 }) => {
   return (
     <section className="text-center">
```
desfero: This is a good use case to reuse our `Message` component. So all modals are consistent

drakkein: :+1:

app/components/shared/buttons/Button.module.scss
```diff
@@ -119,6 +119,7 @@ $_active-color: lighten($_hover-color, 20%);
 
     &#{$button}-small {
       @include font-label;
+      font-weight: $font-weight-bold;
```
desfero: are you sure that's it's safe to change font-size of our small button? It's already used in a couple of places. Is this approved by designers?

app/components/modals/accept-tos-modal/AcceptTosModal.module.scss
desfero: up

app/styles/overrides.scss
```diff
@@ -58,6 +58,11 @@ textarea {
   }
 }
 
+// override default paragraph size
+p {
```
desfero: this i think should be done throughout sass variables

app/components/shared/buttons/Button.module.scss
drakkein: It's verified and accepted by @Louisswryan

PR #2775, opened 2019-04-17
#2543 - Add e2e tests for redeem flow

drakkein:


app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
```diff
@@ -1,24 +1,65 @@
...
+  parseAmount,
+} from "../utils/index";
 import { formField, tid } from "../utils/selectors";
 import { createAndLoginNewUser } from "../utils/userHelpers";
 import { assertBankAccountDetails } from "./assertions";
 
-// Requires dedicated fixture with NEUR available to redeem
-describe.skip("Redeem NEUR", () => {
+describe("Redeem NEUR", () => {
+  before(() => {
```
Moejoe90: this looks to me like a test why is it wrapped around `before`?
is there any specific purpose

app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
```diff
@@ -64,4 +105,65 @@ describe.skip("Redeem NEUR", () => {
 
     cy.get(formField("amount")).should("have.value", "123 456 789.99");
```
Moejoe90: hardcoding values will introduce problems in the future `123 456 789.99`
is there no way to get these values deterministically

app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
```diff
@@ -64,4 +105,65 @@ describe.skip("Redeem NEUR", () => {
...
+
+    // continue transaction
+    cy.get(tid("bank-transfer.redeem-summary.continue")).click();
+    acceptWallet();
+
+    // go to wallet
+    cy.get(tid("bank-transfer.redeem.success.go-to-wallet")).click();
+
+    // wait for balance to refresh
+    cy.wait(1000);
```
Moejoe90: this is a bit error-prone. but I can't think of a good solution for this atm
maybe in the future, we implement a watcher

app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
drakkein: Before redeeming nEUR i need to claim it first, so thats why there is claiming flow in `before`. No fixture with claimed nEUR is available AFAIK.

app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
drakkein: hmmm... I can format number value using one of our formatters and then compare with input value. Since inputs are using different method as mask formatter this should be safe.

app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
drakkein: As I'm thinking about it I've got idea that I should use `goToWallet()` here to get rid of wait. There will be unnecessary redirection, but `get` will wait for page with updated values to load.

Moejoe90: that is for sure better than waiting for an arbitrary amount of time

app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
Moejoe90: :+1:

app/test-e2e/bank-transfer/RedeemNEUR.spec.e2e.ts
Moejoe90: do we have a claim test? 
because this seems to me like a good interlude to add a claim test

drakkein: Yes, @desfero created tests for whole claim flow.

Issue #2737, opened 2019-04-09
display nominee name as provided in the backend


rudolfix:
**Background**
We have hardcoded nominee name in the frontned: `Neumini UG`. with the backend SHA available the nominee address may be changed and we need to display correct nominee name. See more info here
https://github.com/Neufund/platform-backend/issues/1133

**Acceptance criteria**
* [x] everywhere we display hardcoded nominee name now you take it from `nominee_display_name` field of `eto_data`
* [x] if nominee is successfully changed in backoffice app, the displayed nominee name should change to selected nominee in backoffice

**Technical notes**
Please note that **all ETO fixtures coming from backend have the same nominee** with the name `Neumini` so update your e2e tests where you have `Neumini UG` hardcoded

**Testing notes**
Nominee may be changed only in DRAFT/PREVIEW so use ISSUER_DRAFT for manual testing

glupiawymowka:
*tested on .io and stage*

PR #2776, opened 2019-04-17
#2737 - Add nominee name to ETO details

drakkein:


app/test-e2e/eto-view/EtoViewForInvestor.spec.e2e.ts
```diff
@@ -38,7 +38,7 @@ describe("Eto Investor View", () => {
     cy.get(tid("eto-public-view-public-eto-duration")).should("contain", "14 Days");
 
     // TOKEN HOLDER RIGHTS section
-    cy.get(tid("eto-public-view-nominee")).should("contain", "Neumini UG");
+    cy.get(tid("eto-public-view-nominee")).should("contain", "Neumini NOMINEE_NEUMINI");
```
Moejoe90: is this now in the tests? 
`NOMINEE_NEUMINI`

drakkein: All of the ETO's are using this fixture `0xCB6470fa4b5D56C8f494e7c1CE56B28c548931a6` as nominee which has name set to `Neumini NOMINEE_NEUMINI`

Moejoe90: ok fair enough
still its a recipe for problems but the only way would be to conduct an API call from cypress and assert we can live without this now

Issue #2760, opened 2019-04-15
Prompt Ledger Nano users with older firmware to upgrade to our supported firmware version


Moejoe90:
**Background**
After the latest Chrome update version `73.0.3683.103` to be precise the old Ledger support was again broken and therefore connecting to the platform with these older Ledgers isn't possible.

We will need to prompt old firmware users to update their Nano Ledgers to the latest version.

We can do this by adding a new check to see if the Ledger firmware matches what we support, which is currently version 1.5.5.

If it doesn't pass this check, we need to display an upgrade your firmware message as per the design below.

**Designs**

<img width="1774" alt="Screenshot 2019-04-17 at 10 54 48" src="https://user-images.githubusercontent.com/609657/56587185-6353e400-65e1-11e9-839c-c5b5f596ff14.png">

__The copy is as follows:__
> Ledger Nano firmware is unsupported

> Looks like you are using a Ledger Nano with outdated firmware. For security reasons we only support devices with the latest firmware version. Please update your Ledger Nano to access Neufund.

> For update instructions, visit Ledger.com.

The Ledger.com text should open in a new tab the following URL: https://support.ledger.com/hc/en-us/articles/360002731113-Update-device-firmware

**Acceptance criteria & testing notes**
- [x] You can connect to the platform with a Ledger Nano that has our supported firmware version (version 1.5.5).
- [x] You are told to upgrade your firmware if your Ledger fails the firmware check (less than version 1.5.5).

**Technical notes**
We need to write a function that checks the users firmware version is supported as part of the check that verifies a Ledger is connected.

If it's not supported, we should prevent them from connecting to the platform and display a view that prompts them to upgraded.

We need to add a regression test to check any link we point externally to Ledger.com doesn't return a 404.

Note the container for the content above is set to `.col-md-9` and the row class that wraps it is set to `.justify-content-center`. If you zoom in on that screengrab you'll see the `html` used to achieve the layout.

andywalton:
![IMG_0241](https://user-images.githubusercontent.com/609657/56204960-d80aa980-6048-11e9-89aa-318ed3a0b684.JPG)
![IMG_0243](https://user-images.githubusercontent.com/609657/56204961-d80aa980-6048-11e9-87b1-11551525d8a4.JPG)
![IMG_0244](https://user-images.githubusercontent.com/609657/56204962-d80aa980-6048-11e9-8ee2-51e347746353.JPG)

Going to add to this ticket to get copy written and some designs.

[...]

PR #2780, opened 2019-04-18
#2760 - Adds ledger not supported message

Moejoe90:


app/components/wallet-selector/ledger/WalletLedgerInitComponent.tsx
```diff
@@ -94,6 +81,42 @@ export const WalletLedgerInitComponent: React.FunctionComponent<
...
+    {errorMessage && (
+      <Row className="justify-content-center">
+        <WarningAlert className="my-4">
+          <FormattedMessage id="wallet-selector.ledger.start.connection-status" />{" "}
+          <span data-test-id="ledger-wallet-error-msg">{getMessageTranslation(errorMessage)}</span>
+        </WarningAlert>
+      </Row>
+    )}
+    {/* If there is a need for more visual cases then we will need to implement a full solution */}
+    {errorMessage && errorMessage.messageType === LedgerErrorMessage.NOT_SUPPORTED ? (
```
Moejoe90: As mentioned in the comment I don't want to over-engineer a solution that might be not needed in this case. 
I'm assuming that this comment should be enough to indicate if there are more conditions we need to show then a better solution should be done

app/components/wallet-selector/ledger/WalletLedgerInitComponent.tsx
```diff
@@ -38,25 +42,8 @@ const InitStep: React.FunctionComponent<IInitStep> = ({ header, img, desc }) =>
...
 
-interface IWalletLedgerInitComponentProps {
-  isInitialConnectionInProgress: boolean;
-  errorMessage?: TMessage;
-}
-
-export const WalletLedgerInitComponent: React.FunctionComponent<
-  IWalletLedgerInitComponentProps & IIntlProps
-> = ({ errorMessage, intl }) => (
+const LedgerConnectionSteps: React.SFC<IIntlProps> = ({ intl }) => (
```
desfero: `React.FunctionComponent` please, `SFC` is deprecated AFAIK

app/components/wallet-selector/ledger/WalletLedgerInitComponent.tsx
```diff
@@ -38,25 +42,8 @@ const InitStep: React.FunctionComponent<IInitStep> = ({ header, img, desc }) =>
...
-      <Row className="justify-content-center">
-        <WarningAlert className="my-4">
-          <FormattedMessage id="wallet-selector.ledger.start.connection-status" />{" "}
-          <span data-test-id="ledger-wallet-error-msg">{getMessageTranslation(errorMessage)}</span>
-        </WarningAlert>
-      </Row>
-    )}
     <Row>
       <InitStep
         header={intl.formatIntlMessage("wallet-selector.ledger.start.step1.header")}
```
desfero: @Moejoe90 can you refactor also `InitStep` component? We should not use `intl` here as we can pass `TTranslateString to `header` and `desc`. 🙏

app/lib/web3/ledger-wallet/ledgerUtils.ts
```diff
@@ -113,7 +113,7 @@ export const createWeb3WithLedgerProvider = async (
 export const connectToLedger = async (): Promise<() => any> => {
   const getTransport = () => TransportU2F.create();
   const ledgerConfig = await getLedgerConfig(getTransport);
-  if (semver.lt(ledgerConfig.version, "1.0.8")) {
+  if (semver.lt(ledgerConfig.version, "1.5.5")) {
```
desfero: btw, just as a nice side example, we can get the same without `semver` library, `ledgerConfig < "1.5.5"`.

intl/locales/en-en.json
```diff
@@ -1109,6 +1111,7 @@
   "wallet-selector.ledger.start.step5.header": "5. Enable \"Contract support\"",
   "wallet-selector.ledger.start.step6.description": "Choose Browser Support and select Yes by pressing both left and right buttons simultaneously.",
   "wallet-selector.ledger.start.step6.header": "6. Enable \"Browser support\"",
+  "wallet-selector.ledger.update-instructions": "<p>For update instructions visit <a href={href} target=\"_blank\">here</a></p>",
```
desfero: Unfortunately, it's bad for security to have an anchor to an external resource without `rel` attribute. Have look here https://github.com/Neufund/platform-frontend/blob/5acb77be4d2a1473de74b2fe7b7dba35fef76dc8/app/components/modals/tx-sender/shared/TxError.tsx#L72-L84.
ALso, we can get rid of `p`  by moving it to jsx.

app/config/externalRoutes.ts
```diff
@@ -10,4 +10,5 @@ export const externalRoutes = {
     "https://blog.neufund.org/neufund-101-how-to-accept-a-payout-from-neu-4de6eb4cc4a2",
   quintessenceLanding: "https://quintessence.global/",
   quintessenseTermsOfUse: "https://quintessence.global/terms-of-use.pdf",
+  ledgerSupport: "https://ledger.com",
```
desfero: btw, can we navigate the user to the exact page related to firmware update?

andywalton:
Guys, I'm sorry but my updates to #2760 didn't save last week. I've updated the ticket with the exact copy and a screengrab of the design. What you have done isn't far off so it shouldn't take long to tweak this. Please will you review 2760 again with my latest changes. Thanks!

app/components/wallet-selector/ledger/WalletLedgerInitComponent.tsx
Moejoe90: done

app/components/wallet-selector/ledger/WalletLedgerInitComponent.tsx
Moejoe90: even better we can use `<FormattedMessage>` directly 
done

app/lib/web3/ledger-wallet/ledgerUtils.ts
Moejoe90: hmmm and it will automatically detect all version changes really?

intl/locales/en-en.json
Moejoe90: good catch i missed that point
I'll add this to documentation

docs/component-development-guidelines.md
```diff
@@ -27,3 +27,8 @@ enum EMyEnum {
   UPPERCASE_CONSTANT = "uppercase_constant"
 }
 ```
+
+## External Links
```
desfero: 🎖

docs/component-development-guidelines.md
```diff
@@ -27,3 +27,8 @@ enum EMyEnum {
...
+
+When using an external link do not use `<a>` due to security concerns but use directly but
```
desfero: Just typo, last `but` is not needed

intl/locales/en-en.json
```diff
@@ -1114,6 +1116,8 @@
   "wallet-selector.ledger.start.step5.header": "5. Enable \"Contract support\"",
   "wallet-selector.ledger.start.step6.description": "Choose Browser Support and select Yes by pressing both left and right buttons simultaneously.",
   "wallet-selector.ledger.start.step6.header": "6. Enable \"Browser support\"",
+  "wallet-selector.ledger.update-instructions": "<p>For update instructions visit {ledgerUpgradeLink}</p>",
```
desfero: Please move `p` to jsx

app/lib/web3/ledger-wallet/ledgerUtils.ts
desfero: yeah, semver is really small and using simple string for such comparison can be painful for other devs

Moejoe90:
@glupiawymowka 
After investigation, In order to solve the problem fully with the ledger, I will need to make some changes that touch the full app. Thus it won't be part of this PR for now

Issue #2809, opened 2019-04-27
new eto card layout broken after panel switched to grid


AlexeyTuboltsev:
**Description**
after merging grid panel layout new eto cards layout is broken
**Environment**

**Steps To Reproduce**
set `NF_ETO_LIST_GRID=1`, login as wallet_4, go to dashboard
 
**Actual behaviour**
eto card layout has a large white margin on the right

**Expected behavior**
eto cards should look as designed (https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5ca235f479733c07fcbb4341)

**Comments, context, logs**

**Screenshots**
![image](https://user-images.githubusercontent.com/14041336/56855798-5e32c400-6956-11e9-9f35-c94804358e8d.png)

glupiawymowka:
*tested on io*

glupiawymowka:
*tested on .net*

PR #2811, opened 2019-04-27
#2809 fix new etoCard layout

AlexeyTuboltsev:
quick fix for broken css of new eto cards
the eto cards layout in general should be changed to grid as well, i'll do it after vacation


app/components/dashboard/eto-list/EtoList.tsx
```diff
@@ -46,6 +46,7 @@ const EtoListDefault: React.FunctionComponent<TListProps> = ({ etos }) => (
 
 const EtoListComponent: React.FunctionComponent<IStateProps> = ({ etos }) => (
   <>
+    sadfasdf
```
Moejoe90: ?

AlexeyTuboltsev: 😱

AlexeyTuboltsev:
looks like we merged it without Nikola approving it first. Sorry for that...

Issue #2614, opened 2019-03-20
[transactions, platform]  add a button to reset pending transactions


glupiawymowka:
**Background**

On .io, some transactions are pending forever, blocking the fixtures for everyone who needs to use them. 

Please add a button to reset them easily each time it happens. 

**Acceptance criteria**
- [x] a pending transaction can be reset
- [x] new transaction can be started 

**Technical notes**

**Testing notes**
This can be tested with many transactions, like Redeem flow

banciur:
In following link you will find some notes regarding this problem - https://neufund.atlassian.net/wiki/spaces/TEC/pages/585728013/Troubleshooting

rudolfix:
@glupiawymowka flag addes - can be tested

glupiawymowka:
works on io <3

PR #2815, opened 2019-04-28
#2614 - Add transaction reset button on testing environments

drakkein:


app/components/modals/tx-sender/shared/TxPending.stories.tsx
```diff
@@ -18,7 +19,7 @@ const txData: React.ComponentProps<typeof TxPendingLayout> = {
     from: "0x0020d330ef4de5c07d4271e0a67e8fd67a21d523",
     gas: "0x7b0c",
     gasPrice: "0xb2d05e00",
-    hash: "0xe0cbf82ceee3d0a84b762fccf7eefbb4744bf68a6c0e9038a7db57ec8f2346f4",
+    txash: "0xe0cbf82ceee3d0a84b762fccf7eefbb4744bf68a6c0e9038a7db57ec8f2346f4",
```
Moejoe90: `txHash` you mean here

app/components/modals/tx-sender/shared/TxPending.tsx
```diff
@@ -50,11 +57,20 @@ const TxPendingLayout: React.FunctionComponent<TTxPendingLayoutProps> = props =>
     <TxDetails className="mb-3" {...props} />
 
     {props.txHash && <TxHashAndBlock txHash={props.txHash} blockId={props.blockId} />}
+
+    {process.env.NF_ENABLE_TRANSACTION_RESET && (
```
Moejoe90: please assert that `process.env.NF_ENABLE_TRANSACTION_RESE==='1'` because like not if someone puts 
NF_ENABLE_TRANSACTION_RESET=0 in the env it will still pass

app/components/modals/tx-sender/shared/TxPending.tsx
```diff
@@ -50,11 +57,20 @@ const TxPendingLayout: React.FunctionComponent<TTxPendingLayoutProps> = props =>
...
+      <Button className="mt-4" onClick={() => props.deletePendingTransaction()}>
+        Delete transaction
```
Moejoe90: Lets add a comment here that this is on purpose not in the translation file due to it being only for testing flow

app/modules/tx/transactions/sagas.ts
```diff
@@ -188,6 +189,23 @@ export function* etoSignInvestmentAgreementSaga(
   }
 }
 
+export function* removePendingTransaction({ logger }: TGlobalDependencies): any {
+  if (!process.env.NF_ENABLE_TRANSACTION_RESET) {
```
Moejoe90: why do we need to have this assertion here? 
its already asserted in the button

app/modules/tx/transactions/sagas.ts
```diff
@@ -188,6 +189,23 @@ export function* etoSignInvestmentAgreementSaga(
...
+  }
+
+  try {
+    // when modal was closed delete transaction for pending list
+    yield neuCall(deletePendingTransaction);
+    yield put(actions.txSender.txSenderHideModal());
+  } catch (e) {
+    logger.info("Unable to delete pending transaction");
+  } finally {
+    logger.info("Pending transaction has been deleted");
```
Moejoe90: `    logger.info("Pending transaction has been deleted");` 
should be in the `try {}` not in finally

app/modules/tx/transactions/sagas.ts
```diff
@@ -188,6 +189,23 @@ export function* etoSignInvestmentAgreementSaga(
...
+  if (!process.env.NF_ENABLE_TRANSACTION_RESET) {
+    return;
+  }
+
+  try {
+    // when modal was closed delete transaction for pending list
+    yield neuCall(deletePendingTransaction);
+    yield put(actions.txSender.txSenderHideModal());
+  } catch (e) {
+    logger.info("Unable to delete pending transaction");
```
Moejoe90: this should be `logger.error` so that we can log it in sentry 
also please wrap it around an `Error` object

app/components/modals/tx-sender/shared/TxPending.stories.tsx
drakkein: should be just `hash`

app/modules/tx/transactions/sagas.ts
drakkein: you are right, there is no need fo additional checks

app/modules/tx/transactions/sagas.ts
```diff
@@ -190,18 +190,15 @@ export function* etoSignInvestmentAgreementSaga(
...
 export function* removePendingTransaction({ logger }: TGlobalDependencies): any {
-  if (!process.env.NF_ENABLE_TRANSACTION_RESET) {
-    return;
-  }
-
   try {
-    // when modal was closed delete transaction for pending list
+    // call delete transaction saga
     yield neuCall(deletePendingTransaction);
     yield put(actions.txSender.txSenderHideModal());
```
Moejoe90: we don't need this action since its already in finally

Issue #2761, opened 2019-04-15
[ISSUER FLOW] error "could not access ETO files data" thrown just after registration 


glupiawymowka:
**DESCRIPTION**

Just after registration, the issuer gets this error saying he should check if his email/KYC is done.
This is wrong of course, as at this point the user did not have a chance to process verification yet. 

**ENVIRONMENTS**
the test happened on stage, but I guess it's everywhere now 

**STEPS TO REPRODUCE**
1. You can try it even on .io, simply use any metamask account that is not a fixture
2. /eto-landing register/login with metamask
3. This error will popup automatically the moment the dashboard loads 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/83d07f0a-64d8-46c2-a6d1-e5e778a8a45b)

**CONSOLE**
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/cf2aa505-a598-4856-bbcf-579374c65fe3)

**ACCEPTANCE CRITERIA**
- [ ] a completely new issuer doesn't have any error popups just after registering 
- [ ] the given error is not showed also when the user logged out & in without adding email or doing KYC before

rudolfix:
you cannot ask backend for ETO data before e-mail and KYC is done
```
[...]

PR #2844, opened 2019-05-03
#2761 stop requesting eto data before kyc

AlexeyTuboltsev:
refactor EtoDashboard to request eto data only if user have kyc
rename selectShouldEtoDataLoad selector to userHasKycAndEmail


app/components/documents/Documents.tsx
```diff
@@ -28,7 +28,7 @@ import {
   selectIssuerEtoLoading,
   selectIssuerEtoState,
   selectIssuerEtoTemplates,
-  selectShouldEtoDataLoad,
+  userHasKycAndEmail,
```
Moejoe90: then let's rename it to 
`userHasKycAndEmailVerified` to make it explicitly clear

app/components/eto/EtoDashboard.tsx
```diff
@@ -208,37 +224,33 @@ const EtoDashboardStateViewComponent: React.FunctionComponent<IEtoStateRender> =
...
-                step={1}
-                title="VERIFICATION"
-                data-test-id="eto-dashboard-verification"
-              />
-            </Container>
-            <SettingsWidgets
-              isDynamic={true}
-              {...this.props}
-              columnSpan={EColumnSpan.ONE_AND_HALF_COL}
+const EtoDashboardComponent: React.FunctionComponent<IComponentProps & IDispatchProps> = props => {
```
Moejoe90: :+1:

app/components/eto/EtoDashboard.tsx
```diff
@@ -278,21 +287,16 @@ const EtoDashboard = compose<React.FunctionComponent>(
...
-      props.verifiedEmail &&
-        props.backupCodesVerified &&
-        props.requestStatus === ERequestStatus.ACCEPTED,
-    ),
+    isVerificationSectionDone: props.userHasKycAndEmail && props.backupCodesVerified,
     shouldViewSubmissionSection: Boolean(
       props.etoFormProgress && props.etoFormProgress >= SUBMIT_PROPOSAL_THRESHOLD,
     ),
   })),
   onEnterAction<IStateProps>({
```
Moejoe90: What will happen if the dashboard is already loaded and `userHasKycAndEmail` is initially false then user KYC gets approved from the back-office app. 
I think the dashboard will not update since it relies on `onEnterAction` we need to change `onEnterAction` to something that gets called on componentStateUpadate

app/components/documents/Documents.tsx
AlexeyTuboltsev: done

app/components/eto/EtoDashboard.tsx
AlexeyTuboltsev: fixed

app/utils/OnPropsChangeAction.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from "react";
```
Moejoe90: Please remove this redundant component and use lifecycle methods from recompose

AlexeyTuboltsev: done

Issue #2602, opened 2019-03-19
Remove Neufund-Green Line Marker from Off-Card Titles


Louisswryan:
This is a platform-wide change.

If this cannot be implemented in one go, that is ok. I will identify instances for it to be changed.

In some cases, the Neufund-Green Line appears next to titles that are not on cards such as the below example.

![OPPORTUNITIES PAGE.png](https://images.zenhubusercontent.com/5c5184c7df0733a30553c211/6f7f6f77-292b-48aa-a406-dce5e5438417)

 This indicator should just be used within cards and not on the light blue background.

The goal is to make all of the off-card titles look basically the same but without the line and the text justified left within the same content margin. Something like the below.

![OPPORTUNITIES NO LINE.png](https://images.zenhubusercontent.com/5c5184c7df0733a30553c211/6031350b-d4fb-46fd-8e4c-d78cfc26ed0e)

Thanks :)

glupiawymowka:
*tested on .net*

PR #2846, opened 2019-05-04
#2602 dashboard heading

AlexeyTuboltsev:
change all dashboard headings to DashboardHeading


app/components/eto/shared/DashboardHeading.tsx
```diff
@@ -1,33 +1,35 @@
...
-  className = "my-4",
+  className,
   size,
+  description,
 }) => (
   <Heading
     level={3}
     className={className}
-    decorator={hasDecorator}
+    decorator={false}
```
Moejoe90: Why not default false. Maybe the request changes in the future but in all cases its non blocking

app/components/eto/shared/DashboardHeading.tsx
```diff
@@ -1,33 +1,35 @@
...
 
+import { TTranslatedString } from "../../../types";
 import { EHeadingSize, Heading } from "../../shared/Heading";
 
 interface IProps {
   title: string | React.ReactNode;
   step?: number;
+  description?: TTranslatedString;
   "data-test-id"?: string;
-  hasDecorator?: boolean;
```
Moejoe90: Just keep it false if you don't need it. 
Why remove this?

PR #2858, opened 2019-05-06
Add `noFallthroughCasesInSwitch` and `noImplicitReturns`

desfero:
Should generally make the code more verbose and reduce the amount of unhandled edge cases


app/components/eto/registration/pages/etoTermsUtils.tsx
```diff
@@ -29,5 +31,9 @@ export const getProductMeaningfulName = (productName: EProductName) => {
       );
     case EProductName.RETAIL_ETO_LI_VMA:
       return <FormattedMessage id="eto.form.section.eto-terms.product.name.retail-eto-li-vma" />;
+    case EProductName.FIFTH_FORCE_ETO:
+      return invariant(false, "Should not be displayed on UI");
```
AlexeyTuboltsev: maybe make the error text more descriptive? Imagine we have many of those in sentry, we need to distinguish

Issue #2866, opened 2019-05-06
forms justified left after switch to grid


AlexeyTuboltsev:
**Description**
forms are justified to left after new Panel was switched to grid layout

**Environment**
any issuer

**Steps To Reproduce**
log in as issuer and go to any of the registration forms

**Actual behaviour**
forms are justified to the left inside Panel

**Expected behavior**
central justification, width 100% 

**Comments, context, logs**

**Screenshots**
![image copy](https://user-images.githubusercontent.com/14041336/57239826-4aa9ec00-702d-11e9-84bc-d7d6dc1cd7b5.png)

kelexa:
tested on staging - OK

PR #2867, opened 2019-05-06
[hotfix] #2866 fix form justification 

AlexeyTuboltsev:
fix panel justification issue in forms
adjust storybook tests to show the justification of form in panel


AlexeyTuboltsev:
visual regression is ok

PR #2886, opened 2019-05-08
Remove `allowRetailInvestors` and `notUnderCrowdfundingRegulations`

desfero:
@glupiawymowka this one should fix bug with issuer `Documents` not being displayed. Should also allow saving `Eto Terms` section.


app/test-e2e/1.long-e2e-tests/2.ChangeEmail.spec.e2e.ts
```diff
@@ -17,7 +17,7 @@ import { DEFAULT_PASSWORD, generateRandomEmailAddress } from "../utils/userHelpe
 describe("Change Email", () => {
   let email: string;
 
-  describe("Has verified email", () => {
+  describe.skip("Has verified email", () => {
```
desfero:
```suggestion
  describe("Has verified email", () => {
```

Issue #2757, opened 2019-04-14
Remove Mist / Parity logos  from MetaMask registration/log in views


andywalton:
**Background**
When you browse to our Registration or Login pages and click on the MetaMask tab you'll see a page that looks like the Login example below:

<img width="1392" alt="Screenshot 2019-03-28 at 13 02 40" src="https://user-images.githubusercontent.com/609657/55156561-95953180-515a-11e9-8589-6d2449c19455.png">

Remove the section 'MetaMask also supports'. Remove that text, the Mist and Parity logos and the horizontal line that divides this section so that the screens will look as follows:
<img width="1392" alt="Screenshot 2019-03-28 at 13 02 36" src="https://user-images.githubusercontent.com/609657/55156588-a47be400-515a-11e9-8435-0a0c4075e574.png">

**Designs**
As per the screenshot and description above.

**Acceptance criteria**
- [ ] Both the Registration or Login pages no longer have the "MetaMask also supports" section.

**Technical notes**
None

glupiawymowka:
@andywalton what about this? You can see it when you click "Are you an ICBM investor? Click here"
![image](https://user-images.githubusercontent.com/18315589/57460216-fe50ed00-7274-11e9-828e-19f3092883a2.png)

glupiawymowka:
this is on stage already

PR #2891, opened 2019-05-08
#2757 - Remove Mist and Parity logos from login page

drakkein:


Moejoe90:
To be tested after merge

PR #2907, opened 2019-05-10
Remove unused bootstrap css

desfero:
This will reduce the css bundle for 8%. There is still a lot to remove as if you look at generated CSS bundle half of it relates to bootstrap. Unfortunately, it requires more time to find and remove other unused styles and at that moment it's not worth it (like better to invest time by splitting from main bundle all tx related stuff).


AlexeyTuboltsev:
❤️

PR #2909, opened 2019-05-10
#2910 - Filters only offers for multi-jurisdiction logic

Moejoe90:


app/modules/eto/sagas.ts
```diff
@@ -283,11 +280,19 @@ function* loadEtos({ apiEtoService, logger, notificationCenter }: TGlobalDepende
...
 
     const jurisdiction: string | undefined = yield select(selectClientJurisdiction);
-    const filteredEtosByJurisdictionRestrictions = filterEtosByRestrictedJurisdictions(
-      etos,
-      jurisdiction,
+
+    yield all(
+      etos
+        .filter(eto => eto.state === EEtoState.ON_CHAIN)
+        .map(eto => neuCall(loadEtoContract, eto)),
```
AlexeyTuboltsev: @desfero always comments on a map without using the result of it. Let's change it to forEach?

app/modules/eto/selectors.ts
```diff
@@ -120,6 +122,37 @@ export const selectTokenData = (
...
+) =>
+  jurisdiction
+    ? etos.filter(eto => {
+        // @See https://github.com/Neufund/platform-frontend/issues/2789#issuecomment-489084892
+        const isEtoAnOffer = selectIsEtoAnOffer(state, eto.previewCode, eto.state);
+        return (
+          !isEtoAnOffer ||
+          !hiddenJurisdictions[jurisdiction].some(
+            (hiddenJurisdiction: string) => hiddenJurisdiction === eto.product.jurisdiction,
+          )
```
AlexeyTuboltsev: maybe !(isEtoAnOffer && hiddenJusridictions...) ?
but this is non-blocking, just readability

Moejoe90: But it's an or condition not an and
If one of them passes then it's true

app/modules/eto/sagas.ts
Moejoe90: Hmmm it's not a good idea because the call is async and afaik forEach can't handle that

app/modules/eto/selectors.ts
AlexeyTuboltsev: @Moejoe90 interesting enough both expressions are equal

kelexa:
tested on .io - OK

no way to test properly on staging since fixtures for ongoing DE ETOs and LI ETOs missing.

Issue #2855, opened 2019-05-05
Hide empty cap table "donut" circle graphic


andywalton:
**Description**
If an Issuer doesn’t enter cap table information we currently show an empty gray circle (donut) graphic on the company ETO page. We need to hide this if no information exists and only show it if the Issuer has entered information.

**Actual behaviour**
This is how it currently looks without cap table information (incorrect):

<img width="976" alt="issuer-bugs-002b" src="https://user-images.githubusercontent.com/609657/57199931-4b874300-6f85-11e9-9594-980b65ffa891.png">

**Expected behavior**
It should look like this without cap table information (correct):

<img width="944" alt="issuer-bugs-002c" src="https://user-images.githubusercontent.com/609657/57199933-5346e780-6f85-11e9-9f93-77be5a0ac73f.png">

And like this when cap table information exists (correct):

<img width="944" alt="issuer-bugs-002a" src="https://user-images.githubusercontent.com/609657/57199947-65288a80-6f85-11e9-9bcb-f75e6fda6edd.png">

kelexa:
tested on staging

the donut was dropped when there is no shareholder information entered but there is an issue with the space between texts and alignment with the texts below see screenshot

![image](https://user-images.githubusercontent.com/40295416/57850397-1e852c80-77de-11e9-8ecc-7d432ba3f9ed.png)

glupiawymowka:
fixed

PR #2917, opened 2019-05-12
#2855 Remove shareholders chart when it's empty

drakkein:


app/components/eto/public-view/LegalInformationWidget.tsx
```diff
@@ -144,20 +144,22 @@ export const LegalInformationWidget: React.FunctionComponent<IProps> = ({
...
-        </Col>
+        {companyData.shareholders && companyData.shareholders.length > 0 && (
+          <Col>
+            <ChartDoughnut
+              className="mb-3"
+              data={{
+                datasets: [
+                  {
+                    data: shareholdersData.map(d => d && d.shares),
+                    backgroundColor: shareholdersData.map((_, i: number) => CHART_COLORS[i]),
```
Moejoe90: `CHART_COLORS` has only 6 colours.  
CHART_COLORS[i] will be problematic the moment we pass this amount please let's fix this now by either introducing a max cap where we don't show more than CHART_COLORS elements.

app/components/eto/public-view/LegalInformationWidget.stories.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { testCompany } from "../../../../test/fixtures";
+import { LegalInformationWidget } from "./LegalInformationWidget";
+
+const emptyShareholders = {
+  ...testCompany,
+  shareholders: undefined,
+};
+
+storiesOf("ETO/PublicView/LegalInformationWidget", module)
```
Moejoe90: :+1:

app/components/eto/public-view/LegalInformationWidget.tsx
drakkein: i'm thinking something like this might be good solution. 
```
const color = i % CHART_COLORS.length; 
```

Moejoe90: We can also do that but we need to organize with Product @andywalton if its ok to repeat the same colors

drakkein: It we just add more colors :D

drakkein: #2921

Issue #2905, opened 2019-05-10
critical UX problems when registering as ISSUER


rudolfix:
**Background**
Issuers receive many misleading signals and end registering as investors. That's first platform experience and makes our customers to lose enthusiasm for the product!

**Designs**
Here how it looks now
![image](https://user-images.githubusercontent.com/17202864/57535362-5c97d180-7342-11e9-8b7d-aa933242b327.png)

problems:
1. REGISTER and LOGIN lead to investor registration
2. it nowhere says that it's an issuer login
3. it nowhere says that you should use nano ledger
4. issuer waits 60 seconds if ledger not connected to see instructions!

**Acceptance criteria**

- [x] when registering via secret link drop REGISTER and LOGIN buttons, you can drop full header if that's easier!
- [x] drop wallet selector, we show just ledger anyway. on development you will still be able to select wallet via link. drop it without much hesitation!
- [x] instead of wallet selector please put a text "WELCOME ISSUER!"
- [x] when connecting the ledger show instructions immediately. we had this behaviour before. We should have "Register with your Nano Ledger" then below the waiting indicator (the pulsing) one) - or error message. and then instructions
- [x] on first login there's no KYC and e-mail activated etos/me endpoint will return HTTP 400. that is normal! situation below must be handled gracefully by (1) not showing the red toaster (2) not displaying the forms at all - we have spinner pulsating forever there!
![image](https://user-images.githubusercontent.com/17202864/57542710-b607fc80-7352-11e9-80c2-393c9a769f06.png)
- [x] if in the issuer dashboard please replace investor text below replacing "... invest" with "...fundraise"
![image](https://user-images.githubusercontent.com/17202864/57542821-02ebd300-7353-11e9-95bf-aec20b60efa5.png)

**Technical notes**
issuer registration on
https://platform.neufund.io/eto/register/ledger

force registration via metamask
https://platform.neufund.io/eto/register/browser

glupiawymowka:
- [ ] the lower "login" still leads to the investor login, we should also get rid of this one if this is the case 
![image](https://user-images.githubusercontent.com/18315589/57789519-5ede9f00-7739-11e9-8012-f2fed7f57177.png)

otherwise it's tested on .io and correct

kelexa:
tested on staging - all points are OK

PR #2928, opened 2019-05-14
#2905 Decrease ledger timeout

desfero:
This is a part of #2905 that should just decrease ledger timeout error so user can see instructions without waiting 30 seconds.

### DIFF
--- a/app/lib/web3/ledger-wallet/ledgerUtils.ts
+++ b/app/lib/web3/ledger-wallet/ledgerUtils.ts
@@ -8,13 +8,10 @@ import * as Web3ProviderEngine from "web3-provider-engine";
 import * as RpcSubprovider from "web3-provider-engine/subproviders/rpc";
 
 import { promisify } from "../../../utils/promisify";
-import {
-  LedgerContractsDisabledError,
-  LedgerNotAvailableError,
-  LedgerNotSupportedVersionError,
-} from "./errors";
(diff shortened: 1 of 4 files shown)


app/lib/web3/ledger-wallet/ledgerUtils.ts
```diff
@@ -8,13 +8,10 @@ import * as Web3ProviderEngine from "web3-provider-engine";
...
 import { promisify } from "../../../utils/promisify";
-import {
-  LedgerContractsDisabledError,
-  LedgerNotAvailableError,
-  LedgerNotSupportedVersionError,
-} from "./errors";
+import { LedgerContractsDisabledError, LedgerNotAvailableError, LedgerNotSupportedVersionError } from "./errors";
 import { ILedgerConfig, ILedgerOutput, IPromisifiedHookedWalletSubProvider } from "./types";
 
+const ExchangeTimeout = 3000;
```
rudolfix: does user still have time to sign transaction or message? we should not reduce it to 3 seconds. if we show instructions a 60 seconds timeout was pretty ok - user can connect nano run the app and spinner will still be there

desfero: ahh, right. It seems like the wrong assumption.

desfero:
Will reiterate in another PR

Issue #2887, opened 2019-05-08
Issuer ETO page preview / published toggle


andywalton:
**Background**
On the Issue Dashboard:

When an issuer visits their ETO Page - they see the entire ETO listing preview with All the data they have submitted via all the ETO forms they've completed. We need to implement a switch on that listing view to allow them to toggle between All their fields versus the fields they have currently published (and what investors can see publicly) based on the current state of the Issuers ETO (e.g. they're in Coming soon state so their company listing page will have some disabled documents, fields with TBA data etc.)

We need to design and develop a toggle that will allow this view change between:

1. Published (preview state - ie. what investors can currently see online based on the Issuers ETO state); and
2. All data fields (visible based on what the Issuer has currently entering into their ETO forms)

Here's the brief from Aga.

![image](https://user-images.githubusercontent.com/609657/57381969-4d7f1b00-71ac-11e9-97c0-295f7480eca9.png)

**Designs**
Needed. Please add here.

**Acceptance criteria**
- [x] We have designs for the toggle
- [x] We have agreed to what to call these two states (Published versus All etc.)
- [x] As an issuer I can toggle between these two states
- [x] Toggle is visible when Issuer looks at his own profile no matter if it loads it with `me` or public methods, see below
- [x] Toggle is not visible when outside of logged user context, in investor context or when issuer looks at not his own ETO

**Technical notes**
On toggle implementation:
Toggle is overlay is visible no matter what is the ETO state
If you load eto data via `me` endpoints like in this route https://platform.neufund.io/eto/view/ - will return all the data as you set them up in the forms and this is ISSUER view

If you load same data via public method like in this route https://platform.neufund.io/eto/view/3862e0eb-931e-416d-854f-5c5da7dd2731 - will remove investmnt terms in preview and pending states and this is PUBLIC view

So toggle basically changes the enpoint used to load eto data: backend will clean up data for you and you just display it as usual

**Testing notes**
None

Louisswryan:
Hello @Moejoe90 

So the resolution for this is to use New Tabs (layout: position center) but can we make the size large?

Tabs are preferable to toggle because toggle should only indicate a yes/no value.

We will drop the content of the ETO page down to make space for the tab row so a space above the cover image leaving room for the 44px high tabs plus 6px padding on bottom and top.

There are currently 2 Tab values. Tab 1 = In Draft and Tab 2 = Published

Something like the below:
https://zpl.io/beA6mdY

Any questions, just ask

Thank you sir

PR #2930, opened 2019-05-14
#2887 - Adds Eto view tab

Moejoe90:
https://github.com/Neufund/platform-frontend/issues/2887

Things to settle

- [ ]  When should the ETO Preview be loaded

- [ ] Is there an Inactive state when ETO Preview cannot be loaded


app/components/eto/public-view/EtoInvestmentTermsWidget.tsx
```diff
@@ -124,7 +124,7 @@ const EtoInvestmentTermsWidgetLayout: React.FunctionComponent<TExternalProps & T
               value={<InvestmentAmount etoData={etoData} />}
               data-test-id="eto-public-view-investment-amount"
             />
-            {etoData.templates.investmentAndShareholderAgreementTemplate && (
+            {etoData.templates && etoData.templates.investmentAndShareholderAgreementTemplate && (
```
AlexeyTuboltsev: looks like you forgot `!!`

app/components/eto/shared/EtoView.tsx
```diff
@@ -511,4 +515,15 @@ const EtoView = withMetaTags<IProps>(({ eto }, intl) => {
   };
 })(EtoViewLayout);
 
-export { EtoView, EtoViewLayout };
+const EtoView: React.FunctionComponent<IProps> = ({ eto, etoPreview }) => (
+  <Tabs layoutSize="large" layoutOrnament={false} layoutPosition="center">
+    <TabContent tab="In Draft">
+      <EtoViewInternal eto={etoPreview!} />
+    </TabContent>
```
AlexeyTuboltsev: you assert that etoPreview is there, but in the EtoIssuerView nothing prevents it from being undefined. See my comment to EtoIssuerView

app/components/eto/EtoIssuerView.tsx
```diff
@@ -26,10 +28,15 @@ export const EtoIssuerView = compose<TProps, {}>(
...
-    stateToProps: state => ({
-      eto: selectIssuerEtoWithCompanyAndContract(state),
-    }),
+    stateToProps: state => {
+      const eto = selectIssuerEtoWithCompanyAndContract(state);
+
+      return {
+        eto,
+        etoPreview: eto && selectEtoWithCompanyAndContract(state, eto.previewCode),
+      };
```
AlexeyTuboltsev: see my comment to `EtoView.tsx`
both `eto` and `etoPreview` can be undefined, for instance because they're not loaded yet
i suggest to check for it and `branch` to some loading state.
I mean smth like this: `EtoPublicViewByContractId.tsx`

app/components/eto/public-view/EtoInvestmentTermsWidget.tsx
Moejoe90: why would I need to add `!!`

PR #2948, opened 2019-05-16
Unskip eto terms e2e tests

desfero:


desfero:
These tests seem to be broken right now as new logic is there.

Issue #2759, opened 2019-04-15
ETO company page fixes


andywalton:
New ticket to address things missed from #2654

- [x] Banner at top of page detailing ETO Jurisdiction 

![Screenshot 2019-04-18 at 17 17 23](https://user-images.githubusercontent.com/17235402/56371691-db936180-61fd-11e9-81e1-d6c61a623a29.png)

- [x] Reorder, reformat numbers and put in additional information from ETO forms.  

![Screenshot 2019-04-18 at 17 11 17](https://user-images.githubusercontent.com/17235402/56371517-8bb49a80-61fd-11e9-9a7f-20e2e31fa921.png)

- [x] if a value for any of the points is 0.00, leave an empty space 

![image](https://user-images.githubusercontent.com/18315589/56808875-d86a2800-6832-11e9-909d-8fe1ef58ec3f.png)

- [x] ETO Overview widget - there is a comma in ETO token price, that is not taken from the ETO investment terms table, please change to a decimal point

![image](https://user-images.githubusercontent.com/18315589/56809418-1156cc80-6834-11e9-9700-b9ac5b136b0a.png)

Moejoe90:
@AlexeyTuboltsev 
The new requirement is like this 
If the user is an investor then we will keep the banner like in the top 
If the user is an issue we will display this information instead 
https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5cdd6dc414c8f21d4a706b1f 

It should contain only a simple copy text and a link, in the end, the link should point to 
`http://platform.neufund.org.io/eto/view/{$preview-code}`
the link should open in a new tab completely 
where the `preview-code` is the preview code of the eto

glupiawymowka:
tested on stage

PR #2954, opened 2019-05-16
#2759 add jurisdiction banner to eto view

AlexeyTuboltsev:
this fixes #1 of the attached ticket


app/components/eto/public-view/Cover.tsx
```diff
@@ -1,29 +1,63 @@
...
   companyOneliner: TTranslatedString;
   companyLogo: IResponsiveImage;
+  companyJurisdiction: EJurisdiction;
   tags: ReadonlyArray<TTranslatedString> | undefined;
 }
 
+export const JurisdictionBanner: React.FunctionComponent<{ jurisdiction: EJurisdiction }> = ({
+  jurisdiction,
+}) => {
+  const BannerBase: React.FunctionComponent = ({ children }) => (
```
desfero: Can we reuse https://neufund.github.io/platform-frontend/?selectedKind=Molecules%2FNotification&selectedStory=info&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel? The styling is almost the same

app/components/eto/public-view/Cover.tsx
```diff
@@ -1,29 +1,63 @@
...
 interface IProps {
   companyBanner: IResponsiveImage;
   companyName: TTranslatedString;
   companyOneliner: TTranslatedString;
   companyLogo: IResponsiveImage;
+  companyJurisdiction: EJurisdiction;
   tags: ReadonlyArray<TTranslatedString> | undefined;
 }
 
+export const JurisdictionBanner: React.FunctionComponent<{ jurisdiction: EJurisdiction }> = ({
```
desfero: Do we need to nest `JurisdicitonBanner` inside `Cover`? Seems like a completly different responsibility

app/test-e2e/eto-view/EtoViewUtils.ts
```diff
@@ -5,6 +5,8 @@ export const assertEtoView = (title: string, jurisdiction?: EJurisdiction) => {
   cy.get(tid("eto.public-view")).should("exist");
   cy.title().should("eq", title + " - Neufund Platform");
   if (jurisdiction) {
+    // todo commented out until the problem with fixtures is solved
+    // cy.get(tid("eto.public-view.jurisdiction-banner")).should("exist");
```
desfero: i don't see `data-test-id` anywhere inside `JurisdicitonBanner`. And can we also add exact jurisdiction here? `eto.public-view.jurisdiction-banner.${jurisdiction}` for e.g.

app/components/eto/public-view/Cover.tsx
```diff
@@ -1,29 +1,63 @@
...
+  jurisdiction,
+}) => {
+  const BannerBase: React.FunctionComponent = ({ children }) => (
+    <div className={styles.jurisdictionBanner}>
+      <img src={infoIcon} className={styles.icon} alt="" />
+      {children}
+    </div>
+  );
+
+  switch (jurisdiction) {
```
desfero: Can we reuse the same logic for `jurisdiction` as we have here https://github.com/Neufund/platform-frontend/blob/7a91812bd3c40c1e6f21bd83af5d08b119d4a11d/app/components/eto/public-view/JurisdictionDisclaimerModal.tsx#L28-L32

So then there is even no need for any `switch` here.

app/test-e2e/eto-view/EtoViewUtils.ts
AlexeyTuboltsev: done

app/components/eto/public-view/Cover.tsx
AlexeyTuboltsev: this banner suddenly has got more functionality with different contents layouts, so I'd leave it like this for now.

app/components/eto/public-view/Cover.tsx
AlexeyTuboltsev: this would be cool, but no 😞 - we need an adjective here, not a substantive, so we would need another enum like this for adjectives. 
And then, this is not correct too - this should return a TTranslatedString, not a string. There's more work needed at this place too

app/components/eto/public-view/Cover.tsx
AlexeyTuboltsev: looks like you're right. done

app/components/eto/shared/EtoView.stories.tsx
```diff
@@ -22,8 +22,13 @@ const testStore = {
 
 storiesOf("ETO/EtoView", module)
   .addDecorator(withStore(testStore))
-  .add("default", () => (
+  .add("investor view", () => (
```
Moejoe90: :+1:

app/test-e2e/eto-view/EtoViewUtils.ts
```diff
@@ -5,6 +5,7 @@ export const assertEtoView = (title: string, jurisdiction?: EJurisdiction) => {
   cy.get(tid("eto.public-view")).should("exist");
   cy.title().should("eq", title + " - Neufund Platform");
   if (jurisdiction) {
+    cy.get(tid(`eto.public-view.jurisdiction-banner.${jurisdiction}`)).should("exist");
```
Moejoe90: another nice test is in eto issuer view and assert the view public view link should `exist`.

but non blocking

app/components/eto/overview/EtoTimeline/index.ts
```diff
@@ -1 +0,0 @@
-export * from "./EtoTimeline";
```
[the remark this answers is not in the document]
AlexeyTuboltsev: yes, I removed intermediary import/export file here

Issue #2951, opened 2019-05-16
Company listing page changes need for "coming soon" ETO state


andywalton:
Please make the following changes to the Company profile page an Issuer's ETO is in the  `Coming soon` state:

- [x] **Pitch deck button**

At the top of the company listing page, please add a `Pitch deck` button link above the Term sheet button link as per the image below. **It's visible in all states**

![Screenshot 2019-05-16 at 10.36.55.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/69ed83df-4dd8-45f8-9938-1cc97d18e010)

- [x] **Hide the whole section “ETO investment terms” in Coming soon state**

![Screenshot 2019-05-16 at 10.43.32.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/de3ad150-a2df-4e12-bbd7-d4e03c41ea6e)

- [x] **Default display first Pitch accordion**

In the **Pitch** section, the accordions are all by default set to be closed. Please change this so that the **first accordion** displayed in this section is always set to open (ie. not collapsed).

![Screenshot 2019-05-16 at 10.27.50.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/2af397e9-f53c-4c4c-89b5-093c0f56cbe9)

![Screenshot 2019-05-16 at 10.28.00.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/625f683b-3e82-4475-bcff-1fc70f70e64d)

- [x] **Move the module “Partners” up under the Team module...**

... And above the module with the rest of key individuals. All these modules should not be in Tabs, they should be individual modules.

This is related to ticket: [#2888](https://github.com/Neufund/platform-frontend/issues/2888)

![Screenshot 2019-05-16 at 10.45.39.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/07e7ff21-30fc-4ccc-a386-0cda9f5a5c02)

rudolfix:
technical details:
for pitchdeck there's a new field in the `company_data`:
```
company_pitchdeck_url:
        $ref: "#/definitions/Link"
```
it should be a link and properly validated

see https://github.com/Neufund/platform-backend/pull/1699 to be merged

andywalton:
Hi @drakkein --> See the new `company_pitchdeck_url`. I'm going to provide you with a screen shot of where to put this field shortly...

rudolfix:
[...]

Issue #2888, opened 2019-05-08
ETO page - make Advisors, Investors and Partners visible components


andywalton:
**Task**
Hey Wiktor, Andy asked me to leave a comment here - we misunderstood each other.

We should have three standalone modules:

Team
Partners
Investors, Advisors, rest of the key individual groups
What is new?
Partners as standalone module
&
Order of tabs in the third module: we should always show investors first.

**Acceptance criteria**
- [ ] Team and Partners  are standalone components and are stacked one one below the other.
- [ ] Investors, Advisors and the rest -> still in single component with tabs in this order

andywalton:
@drakkein -> I'm sorry but there has been a change in the order mentioned above.

The correct order above should be:
1. Investors
2. Advisors
3. Partners

Vanilizer:
Hey Wiktor, Andy asked me to leave a comment here - we misunderstood each other. 

We should have three standalone modules:
1. Team
2. Partners
3. Investors, Advisors, rest of the key individual groups 

What is new? 
Partners as standalone module
&
Order of tabs in the third module: we should always show investors first.

Hope this helps.

PR #2966, opened 2019-05-16
#2951 #2888 Company listing page changes for coming soon state

drakkein:


app/components/eto/overview/EtoOverviewStatus/TagsWidget.tsx
```diff
@@ -22,6 +23,8 @@ export interface ITagsWidget {
   offeringDocumentType: EOfferingDocumentType;
   innerClass?: string;
   jurisdiction: EJurisdiction;
+  companyPitchdeckUrl?: { title?: string; url?: string };
```
Moejoe90: This could be type `pitchDeck`-> `{ title?: string; url?: string };`
we don't need to do it now

app/components/eto/shared/EtoView.tsx
```diff
@@ -248,6 +253,18 @@ const EtoViewLayout: React.FunctionComponent<IProps & IStateProps> = ({ eto, eto
...
           </Container>
         )}
+
+        {areThereIndividuals(partners) && (
+          <Container columnSpan={EColumnSpan.THREE_COL}>
+            <DashboardHeading
+              title={<FormattedMessage id="eto.public-view.carousel.tab.partners" />}
+            />
+            <Panel>
+              <PeopleSwiperWidget people={partners.members as IPerson[]} key="partners" />
```
Moejoe90: why the force cast here?

app/components/shared/Accordion.tsx
```diff
@@ -43,6 +53,20 @@ export class AccordionElement extends React.Component<
...
-  <div className={styles.accordion}>{children}</div>
+export const Accordion: React.FunctionComponent<IAccordionProps> = ({ children, openFirst }) => (
+  <div className={styles.accordion}>
+    {React.Children.toArray(children)
+      // filter all empty children
+      .filter((child: React.ReactChild) => child)
+      .map((child: React.ReactChild, index: number) =>
+        openFirst && index === 0
+          ? // clone element and add isOpened to it if it's first one
+            React.cloneElement<IAccordionElementProps>(
```
Moejoe90: why do you clone the element?

app/components/eto/overview/EtoOverviewStatus/TagsWidget.tsx
drakkein: hmmm... As I can see there are some places more which uses this object, so would be nice to crate some kind of `interface linkObject { title?: string; url?: string}`

app/components/eto/shared/EtoView.tsx
drakkein: This component was just moved by me, but as I can se during destruction of `eto.company` at beginning of `EtoView` component we are loosing types definition and inferred type here is `any[]`.

app/components/shared/Accordion.tsx
drakkein: We need to append `isOpened` to first child `AccordionElement`, in my opinion this is the cleaner solution.

app/components/eto/overview/EtoOverviewStatus/TagsWidget.tsx
Moejoe90: no worries though not blocking

rudolfix:
Please fix the following:
1. There's no way to edit new pitchdeck url, please check #2951 - new task was just added
2. Please show pitchdeck green document link (above termsheet) in **every state**
3. problematic edge case: regardin #2888 in the component that has multiple list with tabs. if first item (Investors) is missing, you do not show "Advisors" by default and we have click on the tab. please show first available TAB! see below
image

Issue #2980, opened 2019-05-19
rearrange issuer journey on the dashboard - new widget placements and state transitions


rudolfix:
**Background**
By rearranging issuer flow and adding new semi-state (marketing visible) we broke our flow and now steps are disarranged and conflicting widgets are visible at once. This rather large ticket intends to fix the situation up until LISTED state (where terms are visible to investors)
We also improve texting.
See https://docs.google.com/presentation/d/1v2gQFfd_gCI1XDgoDPCVFQpjEpqXbzYg-ehOHILHUlQ/edit#slide=id.g51830a132c_4_4 pages 46 - 55

**Designs**
No new designs are required however it would be good to have reference design with new layout and some widgets and forms

**Acceptance criteria**
* [x] Unify layout of element on the dashboard
See page 46 of attached document. We have certain elements from the top to the bottom of the page places in order that never changes:
- eto status - always there !
- introduction text - html formatted - depends on the step (see below). currently hardcoded and shown for steps 2 to 8, otherwise hidden
- step number and step name with optional explanation text underneath
- widgets (like submit eto, upload termsheet)
- eto forms

please implement those placements, I've notices that sometimes Steps are attached to widgets and we can see two steps at once

* [ ] implement steps 2 - 8
There's step 1, please do not touch it - it's ok. Makes sure issuer has e-mail verified and KYC done.
Please also note that (with some exceptions) we do not have any changes in when we show widgets and how the states change inside widgets. We are almost good with switching logic.

**Step 2: Fill in information about your company**
Issuer is in step 2 when it didn't yet fill all necessary company data. Issuer goes to step 3 when all this data is filled

**Step 3: Publish your listing page**
In this step our new "submit marketing data" widget becomes visible and we unlock ETO forms (as we do now!). Issuer is in this step as long as he submits marketing data in the widget (->4) or he fills all eto forms and submits terms (->6)
Please show *submit marketing data* and *upload terms* widgets side by side if that case happens!

**Step 4: Listing page in review**
Issuer submitted marketing data (`visibility_pending` on `is_marketing_data_visible`.). All other things as above, except we go to state 5 when backoffice accepts listing

**Step 5: Set up your ETO**
Issuer is in this state when marketing data is visible and he didn't yet fill all required eto forms. During that state 'upload terms' may pop up when enough data is filled (do not change existing logic, it's good). Issuer goes to step 6 once terms are uploaded.

**Step 6: Publish your investment offer**
[...]

PR #2998, opened 2019-05-22
#2980 - Rearrange ETO Dashboard

drakkein:


app/components/eto/EtoDashboard.tsx
```diff
@@ -95,34 +99,135 @@ interface IDispatchProps {
   initEtoView: () => void;
 }
 
-const SubmitDashBoardSection: React.FunctionComponent<{
-  isTermSheetSubmitted?: boolean;
-  columnSpan?: EColumnSpan;
-}> = ({ isTermSheetSubmitted, columnSpan }) => (
-  <>
-    <Container columnSpan={EColumnSpan.THREE_COL}>
+const selectStep = (
```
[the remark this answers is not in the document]
rudolfix: I agree, please see **technical details** for this ticket. easiest way to implement it is to have function that translates eto state, seto state on chain, etoData etc. into a step enum.
then you translate step enum into matrix of which dahsboard components should be displayed. I think this is more or less what Lexey says.
btw. you have it almost done in stories, why not implement dashboard similarly

app/components/eto/EtoDashboard.tsx
```diff
@@ -256,19 +334,14 @@ class EtoDashboardComponent extends React.Component<IComponentProps> {
       userHasKycAndEmailVerified,
       isMarketingDataVisibleInPreview,
       shouldViewSubmissionSection,
+      etoStep,
     } = this.props;
 
     return (
       <WidgetGridLayout data-test-id="eto-dashboard-application">
         {!isVerificationSectionDone && (
```
AlexeyTuboltsev: hey, I know this is not your code, but `isVerificationSectionDone` looks completely redundant. 
`isVerificationSectionDone`  === `userHasKycAndEmailVerified` & `backupCodesAreOk`
that means if `!userHasKycAndEmailVerified` we shouldn't show this section at all. These two vars are overlapping and this is not good.
I suggest to ditch `isVerificationSectionDone`, leave only `userHasKycAndEmailVerified` and `backupCodesAreOk`,
branch in connect or somewhere else if `!userHasKycAndEmailVerified` and only check for backupCodes in this component


this is actually my code 🤦‍♂  I deeply regret

app/components/eto/EtoDashboard.tsx
```diff
@@ -166,47 +248,49 @@ const EtoDashboardStateViewComponent: React.FunctionComponent<IEtoStateRender> =
     case EEtoState.PREVIEW:
       return (
         <>
+          {/*Show actions header only if actions are available*/}
```
AlexeyTuboltsev: again, i'm sorry as it's not your code, but could you pls go to line 231 and remove this `if !eto`? If there's no eto, this whole thing shouldn't exist at all. Components shouldn't be getting undefineds and other junk, let's use branch() for this

app/components/shared/Heading.module.scss
```diff
@@ -61,6 +61,16 @@
 
   }
 
+  .disable-transform {
```
AlexeyTuboltsev: please rename it to smth like `noUppercase` or smth so that it's clear what are you disabling 😀 

or see my comment below

app/components/shared/Heading.module.scss
```diff
@@ -61,6 +61,16 @@
...
+    text-transform: inherit;
+  }
+
+  .inherit-font {
```
AlexeyTuboltsev: looks like this component should look completely different from the normal Heading.  If the use case for it is clear, let's introduce a new SomeOtherHeading component with the right font and stuff and no uppercase. 

or maybe DashboardHeading is what you're looking for?

in general I think it's good practice to introduce new specialized components in the moment the usage is clear. Adding dozens of additional params to existing components maybe not the best approach. It all ends up with contradicting params for which we need branching logic and checks for undefined, confusion of devs and design inconsistency. It's bad, bad, bad 😬

glupiawymowka:
The things mentioned before are fixed. Three more from Aga: 
1. There is a typo, it should say  "start referring investorS " 
image

2. There is the same text represented twice. Please delete the one below Step 6 
image

3. Please change the header of this section to : Company information and promotional material
image

PR #3014, opened 2019-05-22
Fix possible zeros on ETO investment terms

desfero:


app/components/eto/public-view/EtoInvestmentTermsWidget.tsx
```diff
@@ -121,7 +121,7 @@ const EtoInvestmentTermsWidgetLayout: React.FunctionComponent<TExternalProps & T
               }
               data-test-id="eto-public-view-new-shares-to-issue"
             />
-            {etoData.newSharesToIssueInWhitelist && (
+            {!!etoData.newSharesToIssueInWhitelist && (
```
desfero: I just found that we forgot to cast the nonboolean value to boolean here. It can cause some strange `"0"` on UI

Issue #2983, opened 2019-05-20
Change juridiction fields to uppercase


huesforalice:
**Background**
The jurisdiction field in the products will be uppercase after https://github.com/Neufund/platform-backend/pull/1691 is merged.

**Designs**
Not needed

**Acceptance criteria**
Frontend must accept uppercase jurisdiction values on the product.

**Technical notes**
Not needed

**Testing notes**
Product selection should work

PR #3015, opened 2019-05-23
#2983 Change jurisdiction fields to uppercase

desfero:


app/lib/api/eto/EtoProductsApi.interfaces.ts
```diff
@@ -16,8 +16,8 @@ export enum EOfferingDocumentType {
 }
 
 export enum EJurisdiction {
-  GERMANY = "de",
-  LIECHTENSTEIN = "li",
+  GERMANY = "DE",
```
Moejoe90: i was thinking why we don't just use the country enum now since its the same.

this was done before because of the whole lowercase country

PR #3020, opened 2019-05-24
upper cases jurisdiction from backend

Moejoe90:
This should solve the issue where the backend might suddenly change the casing without coordination with us


app/modules/eto/selectors.ts
```diff
@@ -51,8 +52,16 @@ export const selectEtoWithCompanyAndContract = (
   const eto = etoState.etos[previewCode];
 
   if (eto) {
-    return {
+    const amendedEto = {
       ...eto,
+      product: {
```
desfero: As discussed, can we move this to a reducer?

Moejoe90: After checking the code in the reducers IMO it's not a good idea to change the code there as this will require us to do a mapping before putting the data and there are a couple of places where this change needs to be done and managed. 

So I will keep as is for now

Issue #3016, opened 2019-05-23
Find memory leak in a dev build with HOC turned on


desfero:
**Description**
There is some memory leak in a dev build where after hot reloading frontend is frozen.

drakkein:
We still have this issue. Some updates of text nodes are working fine for me and not freezing app, but overall we still struggles with it.

PR #3028, opened 2019-05-26
#3016 fix memory leak in onEnterAction()

AlexeyTuboltsev:
move class declaration out of onEnterAction() function

### DIFF
--- a/app/utils/OnEnterAction.tsx
+++ b/app/utils/OnEnterAction.tsx
@@ -1,5 +1,7 @@
 import * as React from "react";
 import { connect } from "react-redux";
+import { withProps } from "recompose";
+import { compose } from "redux";
 
 import { AppDispatch } from "../store";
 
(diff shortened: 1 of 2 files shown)


app/utils/OnEnterAction.tsx
```diff
@@ -7,32 +7,40 @@ interface IOnEnterActionDispatchProps {
...
+}
+
 export const onEnterAction = <P extends object = {}>(options: IOnEnterActionOptions<P>) => (
-  WrappedComponent: React.ComponentType,
+  wrappedComponent: React.ComponentType,
 ) =>
-  connect<{}, IOnEnterActionDispatchProps, P>(
-    undefined,
+  connect<IOnEnterActionStateProps, IOnEnterActionDispatchProps, P>(
+    () => ({
```
desfero: Do you agree that this look's strange to use `mapStateToProps` as a props passing? Maybe we should use `withProps({ wrappedComponent })` here?

drakkein: yeah, It looks kind weird, but non-blocking

drakkein:
Works great 😍

Issue #2921, opened 2019-05-13
ETO details - shareholders chart 


drakkein:
**Description**
Right now shareholders chart allows to use only 6 colors due definition of `CHART_COLORS`. If user creates ETO with more than 6 shareholders, all shareholders above will have undefined color.

We should implement some deterministic colour generator for this feature. 
I case of randomly generated colors each visit will generate differently colored chart which is unacceptable. 

**Environment**
All

**Steps To Reproduce**

**Actual behaviour**

**Expected behavior**

**Comments, context, logs**

**Screenshots**

PR #3029, opened 2019-05-26
#2921 Add color generator to shareholders chart

drakkein:


app/components/shared/charts/utils.ts
```diff
@@ -0,0 +1,9 @@
+const hashCode = (str: string) => [...str].reduce((p, c) => c.charCodeAt(0) + ((p << 5) - p), 0);
```
AlexeyTuboltsev: This looks cool!
The problem with this is that it generates **random** colors, but our aim is to generate **contrasting** colors based on some initial value (like if we want 2 colors, we shouldn't get a white and black, but take some "neu_blue" as base color and get a contrasting yellowish color).  
As I see it, the algorithm would be to take base color (e.g. _#ffee25_), dissect it in _R_, _G_, _B_ channels (r: ff, g: ee, b: 25), divide the color space (0 to 255 or #00 to #ff) into number of segments (if we need 2 colors - add 255 to each channel, if we need 3 colors - divide span by 2 and add 127 to each channel for 2nd color and 255 for 3d color).  
By this you'll achieve that the colors are a) contrasting, b) are in a contrasting order - each color stands between contrasting ones. Also we wouldn't need predefined colors. They are anyway generated with some online color scheme generator using this algorithm :)).

But maybe your solution is sufficient, let's see. So this is non-blocking

app/lib/api/eto/EtoApi.interfaces.unsafe.tsx
```diff
@@ -121,6 +121,15 @@ export const EtoKeyIndividualsType = YupTS.object({
 
 type TEtoKeyIndividualsType = YupTS.TypeOf<typeof EtoKeyIndividualsType>;
 
+const EtoLegalShareholderType = YupTS.object({
```
AlexeyTuboltsev: 👍

app/components/eto/public-view/LegalInformationWidget.tsx
```diff
@@ -159,11 +163,15 @@ export const LegalInformationWidget: React.FunctionComponent<IProps> = ({
               data={{
                 datasets: [
                   {
-                    data: shareholdersData.map(d => d && d.shares),
-                    backgroundColor: shareholdersData.map((_, i: number) => CHART_COLORS[i]),
+                    data: shareholdersData!.map(d => d && d.shares),
```
AlexeyTuboltsev: here and below there shouldn't be non-undefined assertions. `shareholdersData` is an array, empty or full, it never returns undefined. Please change the 'generateShareholders' signature to convince ts, don't use assertions

app/components/shared/charts/utils.ts
drakkein: Let's summon @Moejoe90 to discuss this solution

Moejoe90: Let's test with this and see you're idea is good @AlexeyTuboltsev but I don't think we need to over-engineer this solution

AlexeyTuboltsev: hm. why over-engineering? I'd call it simply  engineering 😃

AlexeyTuboltsev: actually I forgot one step to make them in contrasting order

Issue #2976, opened 2019-05-17
Design changes to "Further investments coming soon banner"


andywalton:
When an investor logs into the Dashboard below the Investment Opportunities we have a banner that says `Further investment opportunities are coming soon!` as follows:

![Coming soon current.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/df217bc2-4a06-422d-9990-7aa115b3b5f8)

We need to change this banner to look like the design below:

![Coming soon example.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/50da93f3-fda8-42ba-aaff-1b93bedcc660)

This card should sit at the end of the Investment Opportunities list as and only show if we have less than 4 cards (our current business rule):

![Coming soon example in situ.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/fc6ff490-f1d7-4f17-a45f-e52c5bb97ca9)

**Design notes:**

I've exported the image below for use in your component - it's a 2x sized file export. The max image width should be `400 pixels` wide.
![further_investments_coming_soon](https://user-images.githubusercontent.com/609657/57931018-bc98f580-78b7-11e9-87e8-06e2184ff4ad.png)

The text should be:
`Further investment opportunities  coming soon!`

The css for this text should be:

```
font-size: 20px;
line-height: 28px;
color: #4C4C4C;
text-align: center;
```

Preferably the text should have a line break so that the words `coming soon!` are on their own line.

Lastly, the box shadow should be the same as the ETO preview thumbnail cards. There should be no hover effect on this card.

PR #3030, opened 2019-05-27
#2976 Add coming soon thumbnail

drakkein:


app/components/App.tsx
```diff
@@ -96,6 +95,6 @@ const App = compose<React.ComponentClass>(
   }),
 )(AppComponent);
 
-const AppHot = hot(App);
+const AppHot = App;
```
desfero: please revert back :)

app/components/eto/overview/EtoOverviewStatus/EtosComingSoonThumbnail.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { Col } from "reactstrap";
+
+import { Panel } from "../../../shared/Panel";
+import { ResponsiveImage } from "../../../shared/ResponsiveImage";
+
+import * as comingSoonImg from "../../../../assets/img/etos-comings-soon/comming_soon.png";
+import * as styles from "./EtosComingSoonThumbnail.module.scss";
+
+export const EtosComingSoonThumbnail: React.FunctionComponent = () => (
+  <Col className="mb-4" xs={12} md={6} lg={4}>
```
desfero: For consistency `Col` should be a part of `EtoList`

app/components/eto/overview/EtoOverviewStatus/EtosComingSoonThumbnail.module.scss
```diff
@@ -0,0 +1,6 @@
+.text {
+  font-size: 20px;
```
desfero: Please always use font-size from `typography.scss`. It's then easier to track our global styles

app/components/eto/overview/EtoOverviewStatus/EtosComingSoonThumbnail.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <Panel data-test-id={`eto-overview-coming-soon`}>
```
desfero: `{` and `}` are not needed

app/components/eto/overview/EtoOverviewStatus/EtosComingSoonThumbnail.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <ResponsiveImage
+        theme={"light"}
+        width={400}
+        height={400}
+        srcSet={{
+          "1x": comingSoonImg,
+        }}
+        alt="Coming soon"
```
desfero: It's strictly decorative image, therefore`alt` not needed here, please use `alt=""`

app/components/eto/overview/EtoOverviewStatus/EtosComingSoonThumbnail.tsx
```diff
@@ -0,0 +1,28 @@
...
+      />
+      <span className={styles.text}>
```
desfero: `p` please

drakkein:
> Please
> 
> 1. Try to optimize size of new asset
> 2. Rename `coming_soon` to `coming-soon` for consistency (there is not consitency in naming of assets right now but let's always use `-` as a separators)
> 3. `Preferably the text should have a line break so that the words coming soon! are on their own line.` Do we need to put `{breakingLine}` in translation?

I would go for word break opportunity tag instead of simply breaking line. For bigger resolutions there is no need to break line, because text can fit into one line. 

For the image, the only thing left for optimization is its resolution, but I would rather not change it to not lose quality.

app/components/dashboard/eto-list/EtoList.tsx
```diff
@@ -28,6 +29,8 @@ const EtoListThumbnails: React.FunctionComponent<TListProps> = ({ etos }) => (
     {etos.map(eto => (
       <EtoOverviewThumbnail eto={eto} key={eto.previewCode} />
     ))}
+
+    {etos.length < 4 && <EtosComingSoonThumbnail />}
```
Moejoe90: so you will one coming soon for every eto that's there?

Issue #2970, opened 2019-05-17
JWT are incorrectly escalated and refreshed


rudolfix:
**Background**
JWT actions done incorrectly by the frontend, and is used differently than intended. This prevents implementation of sensible timeouts and is potential security issue

**Acceptance criteria**
1. [x] when you escalate permissions, you do not send previous JWT, that means that you obtain new JWT and all previously obtained permissions are erased
https://github.com/Neufund/platform-backend/tree/master/signature_auth_api#escalate-jwt
2.  [x] when JWT expires (main timestamp), please use `refresh` backend method to obtain extended JWT
https://github.com/Neufund/platform-backend/tree/master/signature_auth_api#escalate-jwt (now there's a timer that logs you out)
3. [x] `setTimeout` is used to execute actions when JWT is expired. this will not work - does not take into account sleep and hibernation (it just counts ticks of the clock, not absolute time)
4. [ ] ~~when JWT is already expired you must re-login user to obtain fresh JWT. that could be global reaction for `401` HTTP Status on any backend endpoint call as specified here
https://github.com/Neufund/platform-backend/blob/master/signature_auth_api/README.md#backend-request-to-escalate-permissions~~ (discussed with @rudolfix, won't do for now)

**Technical notes**
Please read the whole JWT spec as linked above.

**Testing notes**
1. Should refresh jwt token closer to expiration time (10 minutes on `.io`). From user point, if he's still using the app it should not log out after expiration time with session timeout but should rather still allow using platform.
2. If you hibernation laptop and keep it in this state until jwt token expires (10 minutes on `.io`) then if you open it should logout you with session timeout
3. If you login and then close the tab and try to open in jwt expire it should show login page but without session timeout message

PR #3031, opened 2019-05-27
#2970 jwt security fixes

desfero:


app/components/wallet-selector/WalletSelector.spec.tsx
```diff
@@ -14,7 +15,12 @@ describe("<WalletSelector />", () => {
         oppositeRoute="oppositeRoute"
         userType="investor"
         isLoginRoute
-        openICBMModal={() => {}}
+        openICBMModal={spy()}
+        history={{} as any}
+        location={{} as any}
+        match={{} as any}
+        logoutReason={undefined}
+        hideLogoutReason={spy()}
```
Moejoe90: Do you use this spy in any way?

app/lib/api/SignatureAuthApi.ts
```diff
@@ -39,22 +44,57 @@ export class SignatureAuthApi {
...
-        challenge,
-        response: signedChallenge,
-        signer_type: signerType,
-      },
-    });
+  ): Promise<ICreateJwtEndpointResponse> {
+    return this.httpClient
+      .post<ICreateJwtEndpointResponse>({
+        baseUrl: SIGNATURE_BASE_URL,
+        url: "/jwt/create",
```
Moejoe90: `/jwt/create` why we don't make this a constant since it's used in more than one place

app/lib/api/SignatureAuthApi.ts
```diff
@@ -39,22 +44,57 @@ export class SignatureAuthApi {
...
+  public refreshJwt(): Promise<ICreateJwtEndpointResponse> {
+    return this.authorizedHttpClient
+      .post<ICreateJwtEndpointResponse>({
+        baseUrl: SIGNATURE_BASE_URL,
+        url: "/jwt/refresh",
+        responseSchema: Yup.object().shape({
+          jwt: Yup.string().required(),
+        }),
+      })
+      .then(r => r.body);
```
Moejoe90: IMO we should make all urls into constants

app/modules/auth/email/sagas.ts
```diff
@@ -1,8 +1,47 @@
...
+  selectEmailFromQueryString,
+} from "../../web3/selectors";
+import { selectUserEmail, selectVerifiedUserEmail } from "../selectors";
+import { ELogoutReason } from "../types";
+import { loadUser } from "../user/sagas";
+
+/**
+ * Email Verification
+ */
+export function* verifyUserEmail({ notificationCenter }: TGlobalDependencies): Iterator<any> {
```
Moejoe90: I think the issue with moving 
`verifyUserEmail` is that its a high-level saga as in gets activated with an action directly. that action is activated from `/auth/sagas.ts` 

I wanted to do this for a while but i think if we want to do it correctly we would add a watcher on this level that gets called on the higher level that gets called on the higher level.

cypress/support/fetch-support.js
```diff
@@ -0,0 +1,17 @@
+/*
+ * Cypress doesn't support `fetch` so we need to replace it by XHR during e2e tests.
+ * Remove after https://github.com/cypress-io/cypress/issues/95 is done.
```
Moejoe90: windows 95

app/modules/auth/email/sagas.ts
desfero: i really wanted to do the same (like to add watcher here) but don't want to make pr bigger. Do you think we should do it here or move as a separate ticket?

app/modules/auth/constants.ts
```diff
@@ -0,0 +1,3 @@
+import { minutesToMs } from "../../utils/Date.utils";
+
+export const AUTH_TOKEN_REFRESH_THRESHOLD = minutesToMs(1);
```
rudolfix: that's too frequest. or export to envs so we can set differently on dev/production or at least change to 5 mins

app/modules/auth/jwt/sagas.ts
```diff
@@ -112,46 +176,37 @@ export function* ensurePermissionsArePresentAndRunEffect(
...
-export function* watchRedirectChannel(): any {
-  yield startRedirectChannel();
-  while (true) {
-    const userAction = yield take(redirectChannel);
-    switch (userAction.type) {
-      case EUserAuthType.LOGOUT:
-        yield put(actions.auth.logout());
+    const timeLeft = calculateTimeLeft(expiryDate, true, "milliseconds");
+
+    const timeLeftWithThreshold =
```
rudolfix: I backtrack my previous comment on too frequent checks. this is good!

app/modules/auth/jwt/sagas.ts
```diff
@@ -112,46 +176,37 @@ export function* ensurePermissionsArePresentAndRunEffect(
...
+    // If timing matches exact refresh jwt
+    // in case timeout was delayed (for e.g. hibernation), logout with session timeout message
+    switch (timing) {
+      case EDelayTiming.EXACT:
+        yield neuCall(refreshJWT);
         break;
-      case EUserAuthType.LOGIN:
-        yield put(actions.init.start(EInitType.APP_INIT));
+      case EDelayTiming.DELAYED:
+        yield put(actions.auth.logout({ userType, logoutType: ELogoutReason.SESSION_TIMEOUT }));
```
rudolfix: this is what we want :)

app/modules/auth/sagas.ts
```diff
@@ -1,54 +1,68 @@
...
-      yield put(actions.routing.goHome());
-    } else {
-      yield put(actions.routing.goEtoHome());
-    }
+
+  switch (logoutType) {
+    case ELogoutReason.USER_REQUESTED:
+      {
+        if (userType === EUserType.ISSUER) {
+          yield put(actions.routing.goEtoHome());
```
rudolfix: when we are at that: you can remove `goEtoHome` fully and log out to normal home. we should not show this old page anymore

app/modules/auth/sagas.ts
```diff
@@ -1,54 +1,68 @@
...
+        } else {
+          yield put(actions.routing.goHome());
+        }
+      }
+      break;
+    case ELogoutReason.SESSION_TIMEOUT:
+      if (userType === EUserType.ISSUER) {
+        yield put(actions.routing.goToEtoLogin({ logoutReason: ELogoutReason.SESSION_TIMEOUT }));
```
rudolfix: same

app/utils/JWTUtils.ts
```diff
@@ -38,28 +38,40 @@ export function isJwtExpiringLateEnough(token: string): boolean {
...
-    permissions.forEach(p => {
-      if (!jwtPermissions[p] || jwtPermissions[p] < now) {
-        valid = false;
-      }
-    });
-    return valid;
+
+    if (!currentPermissions) return false;
+
+    return permissions.every(p => !!currentPermissions[p] && currentPermissions[p] > now);
```
Moejoe90: :+1: 
We should have some util tests for `hasValidPermissions` btw not in this pr

app/utils/JWTUtils.ts
```diff
@@ -38,28 +38,40 @@ export function isJwtExpiringLateEnough(token: string): boolean {
...
+  const base64Url = token.split(".")[1];
+  const base64 = base64Url.replace("-", "+").replace("_", "/");
   return JSON.parse(window.atob(base64));
 }
 
 /**
- * Check wether permissions are valid
+ * Check whether token is still valid
+ */
+export function isValid(token: string): boolean {
```
Moejoe90: we should have some `util` tests for this one since its added in this pr

app/modules/auth/sagas.ts
```diff
@@ -81,53 +95,55 @@ function* handleSignInUser({ logger }: TGlobalDependencies): Iterator<any> {
...
+      case EUserAuthType.LOGIN:
+        yield put(actions.init.start(EInitType.APP_INIT));
+        break;
+    }
+    yield delay(REDIRECT_CHANNEL_WATCH_DELAY);
   }
 }
 
-export const authSagas = function*(): Iterator<Effect> {
+export function* authSagas(): Iterator<Effect> {
```
Moejoe90: why change it to the function context?

app/modules/auth/sagas.ts
```diff
@@ -1,54 +1,68 @@
...
+      break;
+    case ELogoutReason.SESSION_TIMEOUT:
+      if (userType === EUserType.ISSUER) {
+        yield put(actions.routing.goToEtoLogin({ logoutReason: ELogoutReason.SESSION_TIMEOUT }));
+      } else {
+        yield put(actions.routing.goToLogin({ logoutReason: ELogoutReason.SESSION_TIMEOUT }));
+      }
+      break;
+    case ELogoutReason.ALREADY_LOGGED_IN:
+      // no action is required
```
Moejoe90: maybe we log it anyway? 
It seems to me like this is an edge case worth noting as how can a user already logged out do it. if they found a way we should know about it

app/utils/safeTimers.ts
```diff
@@ -0,0 +1,69 @@
...
+ * @returns timerId Use `clearSafeTimeout(timerId)` to cancel timeout
+ */
+export const safeSetTimeout = (fn: (d: EDelayTiming) => void, ms: number): any => {
+  const timerId = ++lastActiveTimerId;
+
+  activeTimers.push(timerId);
+
+  safeSetTimeoutRec(fn, Date.now() + ms, timerId);
+
+  return timerId;
```
Moejoe90: hmmm maybe I got it wrong but `safeSetTimeout` returns `timerId` correct and not actually `EDelayTiming` which is returned by safeDelay. 

I think this wasn't caught because of type `any`

app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      cy.clock(Date.now());
+
+      const jwtExpiryDate = getJwtExpiryDate(getJwtToken());
+
+      const diff = jwtExpiryDate.diff(moment(), "milliseconds");
+      const expectedTokenRefreshTimeFromNow = diff * AUTH_TOKEN_REFRESH_THRESHOLD;
+
+      // add ten seconds to simulate inactivity (for e.g. hibernation)
+      cy.tick(expectedTokenRefreshTimeFromNow + 10000);
```
Moejoe90: Does `cy.tick` work? 
I think last time I tried it had some problems

app/modules/auth/jwt/sagas.ts
```diff
@@ -1,105 +1,169 @@
...
   }
 }
 
 /**
- * Saga & Promise to fetch a new jwt from the authentication server
+ * Save jwt to the browser storage and update the store
  */
-export async function obtainJwtPromise(
+function* setJwt({ jwtStorage }: TGlobalDependencies, jwt: string): Iterator<Effect> {
+  jwtStorage.set(jwt);
```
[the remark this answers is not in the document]
desfero: localStorage and sessionStorage are both sync operation (unfortunately as they are quite slow). There is already new build in module called KV storage but the support is very bad right now (https://developers.google.com/web/updates/2019/03/kv-storage)

app/components/eto/overview/EtoOverviewStatus/CampaigningWidget/CampaigningActivatedInvestorApprovedWidget.tsx
```diff
@@ -39,15 +39,15 @@ interface IWithProps {
   pledgedAmount: number | "";
 }
 
-interface ILocalStateProps {
+type TLocalStateProps = {
```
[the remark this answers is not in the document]
desfero: For consistency with `TLocalStateProps` of another component

app/components/shared/WarningAlert.stories.tsx
```diff
@@ -3,4 +3,6 @@ import * as React from "react";
 
 import { WarningAlert } from "./WarningAlert";
 
-storiesOf("WarningAlert", module).add("default", () => <WarningAlert>Alert message</WarningAlert>);
+storiesOf("Core|Atoms/WarningAlert", module).add("default", () => (
```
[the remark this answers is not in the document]
desfero: yes, i will do that closer to the storybook migration date (updating storybook from 4 to 5)

app/components/wallet-selector/WalletRouter.tsx
```diff
@@ -9,13 +10,16 @@ import { getRedirectionUrl } from "./walletRouterHelpers";
...
 }
 
-export const WalletRouter: React.FunctionComponent<IProps> = ({ rootPath }) => (
+export const WalletRouter: React.FunctionComponent<IProps> = ({ rootPath, state }) => (
   <SwitchConnected>
     <Route path={`${rootPath}/light`} component={WalletLight} exact />
     <Route path={`${rootPath}/browser`} component={WalletBrowser} exact />
     <Route path={`${rootPath}/ledger`} component={WalletLedger} exact />
-    <Redirect to={getRedirectionUrl(rootPath)} />
+    {/* Preserve location state after redirect, otherwise session timeout message won't work */}
```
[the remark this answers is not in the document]
desfero: Like we store the logout reason in the history api state. During logout i'm redirecting to `login` route. If we didn't pass route state here then immediately after moving to `login` wi will lose the logout reason

app/components/wallet-selector/WalletRouter.tsx
```diff
@@ -9,13 +10,16 @@ import { getRedirectionUrl } from "./walletRouterHelpers";
 
 interface IProps {
   rootPath: string;
+  // we don't care here about exact type
+  state: unknown;
```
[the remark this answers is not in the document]
desfero: agree, maybe even let's call it historyApiState?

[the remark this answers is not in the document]
desfero: done

app/components/wallet-selector/WalletSelector.spec.tsx
desfero: hm, let's replace it with noop function then

app/lib/api/SignatureAuthApi.ts
desfero: agree, done

app/modules/auth/sagas.ts
desfero: moved to #3109

app/modules/auth/jwt/sagas.ts
```diff
@@ -1,105 +1,169 @@
...
+  { logger }: TGlobalDependencies,
   effect: Iterator<any>,
   permissions: Array<string> = [],
   title: TMessage,
   message: TMessage,
   inputLabel?: TMessage,
 ): Iterator<any> {
-  const jwt = jwtStorage.get();
-  // check wether all permissions are present and still valid
+  const jwt: string = yield select(selectJwt);
```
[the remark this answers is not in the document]
desfero: hm, i don't know. My general idea is to minimize the use of `localStorage.set/get` in the app as this is harder to track. Ideally, I would prefer to have only a single set and get per item. Like to feed data into the redux store and then always use redux store. What do you think? If you are against this approach we can rever this

app/utils/JWTUtils.ts
desfero: done, whole utils are fully tested

app/modules/auth/sagas.ts
desfero: consistency, i see almost all sagas use function declaration

app/utils/safeTimers.ts
desfero: hm, i don't understand. What do you mean here? You are right here we should replace `any` by `number` as this is a timer id that is used to cancel timer.

app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
desfero: hm, i checked with wrong values and test is failing so seems like it works here. But it works here because `safeSetTimeout` call recursively `setTimeout`. Because of that `cy.clock` can mock `setTimeout`

PR #3068, opened 2019-06-04
Enables Skipped tests 

Moejoe90:


app/test-e2e/eto/Campaigning.spec.e2e.ts
```diff
@@ -30,20 +30,6 @@ describe.skip("Eto campaigning state", () => {
     assertRegister();
   });
 
-  it("should show founders quote when logged in and campaigning date is not set", () => {
```
desfero: We still need e2e tests for `eto-overview-status-founders-quote` as now it's not used anywhere.

Issue #2877, opened 2019-05-08
Use formatter components for number rendering


AlexeyTuboltsev:
**Background**
all numbers must be rendered with new number formatting components:
MoneyNew :)
MoneyRange
FormatNumber
FormatNumberRange

* [x] eto public view / LEGAL INFORMATION
* [x] eto public view / ETO INVESTMENT TERMS
* [x] eto overview panel
* [x] incomingPayoutWidget
* [x] myWallet
* [x] bookbuildingWidget
* [x] claimWidget
* [x] new EtoCard
* [x] etoRaisedAmount
* [x] bankTransferFlow
* [x] icbmWalletBalanceModal
* [x] txSender
* [x] wallet
* [ ] EtoPledgeApi.interfaces.unsafe.tsx (!)
* [x] moneySuite

**Designs**

**Acceptance criteria**

**Technical notes**

**Testing notes**

Issue #2320, opened 2019-01-16
move FormattedMessage out of interfaces


AlexeyTuboltsev:
after translations PR has been merged we can continue to moving all translations to TMessage and resolve them at rendering place only

PR #3069, opened 2019-06-04
#2877 money formatters in bank flow

AlexeyTuboltsev:
this pr removes old Money component from 
- bank transfer flow components
- portfolio components

it changes some typings and util functions to stop putting BigNumbers to state


app/modules/investor-portfolio/selectors.ts
```diff
@@ -123,10 +123,12 @@ export const selectCalculatedEtoTicketSizesUlpsById = (state: IAppState, etoId:
   const contrib = selectCalculatedContribution(state, etoId);
   const investorTicket = selectInvestorTicket(state, etoId);
 
-  let min = (contrib && contrib.minTicketEurUlps) || (eto && Q18.mul(eto.minTicketEur || 0));
+  let min =
+    (contrib && new BigNumber(contrib.minTicketEurUlps)) ||
```
Moejoe90: I would prefer that `new BigNumber` declarations are up instead of having them in the operation its self.

app/modules/investor-portfolio/selectors.ts
```diff
@@ -123,10 +123,12 @@ export const selectCalculatedEtoTicketSizesUlpsById = (state: IAppState, etoId:
...
+    (eto && Q18.mul(eto.minTicketEur || new BigNumber(0)));
```
Moejoe90: why do we need to create a new `bigNumber` instance every time just have one constant in a higher scope
` const  = new BigNumber(0)`
and use it

AlexeyTuboltsev: 🙀

AlexeyTuboltsev:
blocked by #2320

app/components/modals/tx-sender/redeem/BankTransferRedeemInit.tsx
```diff
@@ -84,7 +84,7 @@ const getValidators = (minAmount: string, neuroAmount: string) =>
               values={{
                 minAmount: formatNumber({
                   value: minAmount,
-                  roundingMode: ERoundingMode.DOWN,
+                  roundingMode: ERoundingMode.DOWN, //fixme round up
```
Moejoe90: In the future can you use `todo` instead of `fixme`. `todo` is our agreed on convention

AlexeyTuboltsev:
waiting for qa, don't merge

AlexeyTuboltsev:
since this is not a part of bank flow and there's pretty much work needed (this summary uses many custom formatting functions), this will be fixed in a separate pr, expected today

Issue #3075, opened 2019-06-05
Bad formatting of investment cards


pstroinski:
**Description**
Cards are not scaling properly. Background not fully filling the whole card. Also problem with scaling 

**Environment**
PRD

**Steps To Reproduce**
Go to the investor dashboard and see how the cards behave on different resolutions

**Actual behaviour**
Wrong scaling of the cards

**Expected behavior**
The background should fill the whole card and not scale like on the screens

**Comments, context, logs**

**Screenshots**
![image (2)](https://user-images.githubusercontent.com/27957897/58945909-0d985d00-8785-11e9-81bd-a8b7b4a42902.png)
![image (3)](https://user-images.githubusercontent.com/27957897/58945953-23a61d80-8785-11e9-9815-320265a2286f.png)

PR #3077, opened 2019-06-05
#3075 Fix bad formatting of investment cards

### LINKED ISSUE #3057: [ETO cards] - center the counter
**Description**
After mobile fixes, the countdown on the ETO cards got left-aligned 

**Environment**
dev, stage 

**Steps To Reproduce**
1. login as an investor on dev
2. see the investment opportunities (there should be at least one with the countdown to public sale) 
**Actual behaviour**

**Expected behavior**

please make sure that: 
- [ ] Heading-module is left aligned
- [ ] ``EtoOverviewThumbnail-module__group-wrapper___2DoWC`` is left aligned
- [ ] ``CounterWidget-module__counter-widget___2HGx5`` is centered

desfero:
Also fixes #3057.


app/components/eto/overview/EtoOverviewThumbnail/EtoCardPanel.module.scss
```diff
@@ -1,14 +1,19 @@
-.button {
-  transition: transform .1s ease-in-out;
+.eto-card-panel {
   grid-column: span 2;
```
desfero: @AlexeyTuboltsev I'm generally not sure about using `grid-column` here but was trying to make as little changes as possible. Like eto cards is not the case where we should use `grid-column: span 2` as they are always equal in size so in this case, it's safer to just use `3/2/1` column grid over `6/4/2` as it's easier to manage IMO.

app/components/eto/overview/EtoOverviewStatus/CounterWidget.module.scss
```diff
@@ -1,4 +1,5 @@
 .counter-widget {
+  margin: 0 auto;
```
AlexeyTuboltsev: 👍

app/components/eto/overview/EtoOverviewThumbnail/EtoCardPanel.module.scss
AlexeyTuboltsev: yes, I agree. The problem is that I tried to make as little changes as possible too :)). That's why I reused the grid type we already had

PR #3080, opened 2019-06-05
Fix EtoFileApi typings

desfero:
also, use `all` effect combinator for documents to improve performance


app/modules/eto-documents/sagas.ts
```diff
@@ -131,11 +132,17 @@ export function* loadEtoFileData({
...
+}: TGlobalDependencies): Iterator<any> {
   try {
-    yield neuCall(loadIssuerEto);
-    const stateInfo = yield apiEtoFileService.getEtoFileStateInfo();
-    const allTemplates = yield apiEtoFileService.getAllEtoTemplates();
+    const {
+      stateInfo,
+      allTemplates,
+    }: { stateInfo: TStateInfo; allTemplates: TEtoDocumentTemplates } = yield all({
+      stateInfo: apiEtoFileService.getEtoFileStateInfo(),
```
Moejoe90: I have a suspicion that `yield all` is causing some race condition before merging lets run the test a couple of times to make sure nothing is getting flaky

it was like this and then I changed it to have separate `yields`

desfero: talked with Moe about this, it's not related to the changes i made here

PR #3083, opened 2019-06-05
#2320 move formatted message out of interfaces

AlexeyTuboltsev:
this PR 
- removes the last occurrences of old Money from the code. The #2877 is solved (there's still some cleanup needed though)
- moves all `FormattedMessage`s and `React` dependency out of interfaces.


app/components/shared/formatters/MoneyRange.tsx
```diff
@@ -56,17 +71,20 @@ export const MoneyRange: React.FunctionComponent<
...
-      ) : (
+      );
+    } else {
+      formattedValue = (
         <FormatShortNumberRange
           valueFrom={valueFrom}
           valueUpto={valueUpto}
           inputFormat={inputFormat}
-          outputFormat={outputFormat}
+          outputFormat={outputFormat as EAbbreviatedNumberOutputFormat}
```
Moejoe90: why do we need to do the force casting here?

app/components/shared/formatters/Money.tsx
```diff
@@ -70,39 +71,42 @@ const MoneyNew: React.FunctionComponent<IMoneyProps & IMoneyCommonProps & Common
     //todo: this should pass through 0 as well. Use isValidNumber from the #2687 PR when it's merged
     const decimalPlaces = selectDecimalPlaces(moneyFormat, outputFormat);
     if (
-      Object.values(ENumberOutputFormat).includes(outputFormat) &&
-      (outputFormat === ENumberOutputFormat.FULL_ROUND_UP ||
-        outputFormat === ENumberOutputFormat.ONLY_NONZERO_DECIMALS_ROUND_UP)
+      outputFormat === EAbbreviatedNumberOutputFormat.SHORT ||
```
Moejoe90: Do you have unit tests for all these cases?

app/components/shared/formatters/MoneyRange.tsx
AlexeyTuboltsev: done. The reason was that ts doens't understand `Object.values(ENumberOutputFormat).includes(outputFormat)` on enums, mind you

app/components/shared/formatters/Money.tsx
AlexeyTuboltsev: i added a couple

PR #3093, opened 2019-06-06
Fix firefox profile page layout bugs

desfero:
- [ ] Check visual regression

Should fix 
image


desfero:
This was fixed by firefox

PR #3094, opened 2019-06-07
Refactors platform to use case insensitive assertions for jurisdiction

Moejoe90:


Moejoe90:
@AlexeyTuboltsev 
The root of the problem is bigger than API formats etc its the full issue we have with different formats and links. 

From now on if we are to deal with any Jurisdiction we have to keep in mind that the casing should be insensitive unless mentioned otherwise. We also from now on need to take into account that url links can be indeed outdated or with a wrong casing with that in mind there is no alternative than to let go of case sensitivity when it comes to implementation

AlexeyTuboltsev:
Well then the other question is why are we having case-sensitive urls? They shouldn't be. We can't do much with outdated links, but IMO there shouldn't be such thing as url in wrong casing. Frankly, I never thought that react-router even allows it to be handled it in a case-sensitive manner

Anyway, my problem with this solution is that we shouldn't be leaking all this api/data acquisition/data preparation/etc details into components at all. Ideally our state should only have clean and nice and _complete_ data ready to be visualized _or_ some markers that the data is not complete yet. All preparations should go somewhere on the api/sagas level

Moejoe90:
The case sensitivity didn't come from react router even though it will be introduced in version 5 ;). 

It's from our logic. I would love to have 
```
Ideally, our state should only have clean and nice and complete data ready to be visualized or some markers that the data is not complete yet. 
```
But since we are dealing with unknown unknowns, for example, the situation yesterday. Also, the data source coming without a clearly defined state then this will only complicate things.

Again we can do force lowerCasing or upperCasing for example as you mentioned but we need to take into account future use. Every time you use jurisdiction you will need to remember that the input can come different than expected and the sources for this input are many. 

The solution that makes sense IMO for now is let go if case sensitivity for two reasons 

1 - For future development, it's very clear what to expect and what to assert without any confusion
2 - There can be a source of data we didn't think about yet, for example, a link.

If you have a better solution that will address these to problems then we can implement that solution. The mentioned solution here https://github.com/Neufund/platform-frontend/pull/3094#pullrequestreview-247073361 IMO doesn't cover both :) 

But maybe you have an idea

AlexeyTuboltsev:
My solution would be 1. to lowercase the urls in all cases before handling them any further because they should be case insensitive. 2. Lowercase the jurisdicrion inside api calling sagas before putting data in state. This way we'd ensure that app has it in the right format in all cases and future devs won't need to think about it anymore. I don't see any other way data can appear in the app.

Moejoe90:
Temp close until further refactor

PR #3095, opened 2019-06-07
Fix eto terms tests

desfero:


app/components/eto/registration/pages/EtoTerms.tsx
```diff
@@ -224,7 +224,7 @@ const EtoRegistrationProductType: React.FunctionComponent<IProps> = ({
             name="productId"
             label={
               <Tooltip
-                data-test-id={`eto-terms.product.${product.id}.tooltip`}
+                data-test-id={`eto-terms.product.${product.name.replace(/\s/g, "-")}.tooltip`}
```
AlexeyTuboltsev: i suggest to make an utility function for it.

app/test-e2e/1.long-e2e-tests/1.FullEtoRegistrationAndSubmit.spec.e2e.io.ts
```diff
@@ -81,7 +81,7 @@ describe("Eto Forms", () => {
           {
             productId: {
               // Value is an id of a product taken from EtoProductApi
```
AlexeyTuboltsev: please drop the comment if it's not valid anymore

app/components/eto/registration/pages/EtoTerms.tsx
desfero: hm, maybe we have something already?

PR #3097, opened 2019-06-09
Add easy way to connect to VM backend

drakkein:
This PR adds an easy way to connect to yours VM backend. All you have to do is to add your ip address into `.env` file and run `start:remote:vm`.
Ex: 
```
NF_VM_ADDRESS="123.456.789.012"
```


webpack/proxy-urls.js
```diff
@@ -1,74 +1,88 @@
...
+  : "localhost";
 
 /**
  * Unified local proxy configuration. Add new urls here.
  *
  * Params:
  * backendUrl: string
  *
- * NOTE: if backendUrl === http://localhost we assume it's local setup and links are slightly different (services run on different ports)
+ * NOTE: if backendUrl === http://134.209.250.167 we assume it's local setup and links are slightly different (services run on different ports)
```
desfero: hm, we probably should update the comment

webpack/webpack.config.dev.js
```diff
@@ -6,6 +6,11 @@ const ForkTsCheckerWebpackPlugin = require("fork-ts-checker-webpack-plugin");
 const generateProxyConfig = require("./proxy-urls");
 const configCommon = require("./webpack.config.common");
 const paths = require("./paths");
+const loadAppEnv = require("./loadAppEnv");
+
+const applicationEnv = loadAppEnv(process.env);
+const targetAddress =
```
desfero: this is almost a duplicate of line 13 from proxy-url, maybe we can pass it as a parameter to `generateProxyConfig`? (and at the same time move `generateProxyConfig` options as an object

drakkein: yeah, good idea

PR #3103, opened 2019-06-12
Changes awaitedClick to wait for button to be enabled

Moejoe90:
Asserting that `clickable` button is active is better than waiting for an arbitrary amount of time. I expect to shed of some precious time from tests. 

Maybe not much.


Moejoe90:
will close since it yields to issues in the pipeline

PR #3118, opened 2019-06-17
Update Health Check Script + Auto Fetch Latest Sha Agreement 

Moejoe90:


scripts/getShasFromHealthCheck.js
```diff
@@ -43,7 +43,9 @@ const getApiSha = async () => {
         "platform-contracts-artifacts",
       );
 
-      const CI_COMMAND = `git checkout ${response.artifacts_sha} && git status`;
+      const CI_COMMAND = `git submodule update --init --recursive && git checkout ${
+        response.artifacts_sha
+      } && git status`;
```
AlexeyTuboltsev: doesn't the LOCAL_COMMAND need a git fetch or smth too?

Moejoe90: Ah good catch you are correct

app/test-e2e/utils/ethRpcUtils.ts
```diff
@@ -22,6 +25,33 @@ const requestFromWeb3Node = (methodName: string, params: string[] | object[]) =>
     },
   });
 
+// TODO: Move all helper methods to use Fetch instead of cy.request
+const requestFromWeb3NodeFetch = (methodName: string, params: string[] | object[]) =>
```
Moejoe90: Fetch is better for regular promises than `cy.request`

PR #3133, opened 2019-06-18
#3016 Fix hot reloading

desfero:


package.json
```diff
@@ -235,6 +234,7 @@
     "react-dom": "^16.6.3",
     "react-dropzone": "^4.2.8",
     "react-helmet": "^5.2.0",
+    "react-hot-loader": "^4.6.5",
```
desfero: Seems like the problem is with `react-hot-loader` that makes `react` performing an infinite loop. Moved also to `dependencies` cause we use it in production bundle

Issue #3100, opened 2019-06-11
[ETO forms] - mark "company founding date" as obligatory field


glupiawymowka:
**Description**

When going through the forms, the issuer will see "company founding date" as the voluntary field. However, it is not possible to save without filling this one. Please make it visible to the user that this field is obligatory. 

**Env** 
All

**Steps to reproduce**

1. Register as a new company or use ISSUER_BLANK_QA_HAS_KYC on .io 
2. Go to the "Legal" form and fill in only the fields with asterisks 
3. Try to save 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/9546a4b5-b518-4103-a356-a03c5c6d5908)

**Actual behavior** 
Form won't save as it sees "Company_founding_date" as the obligatory field

**Expected behavior** 
There is an asterisk next to the Company_founding_date so the user knows they need to fill this input to continue

desfero:
I wanted to fix this one for ages... Let me fix it

glupiawymowka:
*tested on stage*

PR #3147, opened 2019-06-21
#3100 Mark date as obligatory field

desfero:


app/components/shared/forms/fields/FormFieldDate.tsx
```diff
@@ -0,0 +1,157 @@
...
+};
+
+enum EFieldDateType {
+  YEAR = "year",
+  MONTH = "month",
+  DAY = "day",
+}
+
+class FormFieldDate extends React.Component<IProps & TDataTestId> {
+  cache = {
```
AlexeyTuboltsev: can you pls make them `undefined`? "" is an invalid value, why introduce it into the code? Formik uses `undefined`s as empty values

I mean I understand that that's how the component works, but maybe there's a not very hard way to make it not call `field.onChange` until the last blur, when all fields have value? Like, if there's a blur and the next focus is not in this component, then save value and let formik check for errors etc. But don't let formik save invalid things like `38-""-55`
this is not blocking anyway, just looks ugly

app/components/shared/forms/fields/FormInput.unsafe.tsx
```diff
@@ -1,11 +1,14 @@
...
 import * as React from "react";
 
-import { CommonHtmlProps } from "../../../../types";
-import { FormInputRaw, IFormInputRawExternalProps } from "./FormInputRaw.unsafe";
+import { InputLayout } from "../layouts/InputLayout";
 import { applyCharactersLimit, isNonValid } from "./utils.unsafe";
 
-export type FormInputProps = IFormInputRawExternalProps & FieldAttributes<any> & CommonHtmlProps;
+type TExternalProps = {
+  customValidation?: (value: any) => string | Function | Promise<void> | undefined;
```
AlexeyTuboltsev: well the `value` is not `any`, it's actually a `string | undefined`? It can also be a number, but I'm not sure if it's used this way anywhere and it should def'ly be removed because we need to handle all numbers a) as strings b) via MaskedNumberInput component

app/components/shared/forms/fields/FormFieldDate.tsx
```diff
@@ -0,0 +1,157 @@
...
+    day: "",
+    month: "",
+    year: "",
+  };
+
+  monthInput = React.createRef<HTMLInputElement>();
```
AlexeyTuboltsev: 👍

app/components/shared/MaskedMoneyInput.tsx
```diff
@@ -14,7 +14,7 @@ import {
   stripNumberFormatting,
   toFixedPrecision,
 } from "./formatters/utils";
-import { FormInputRaw } from "./forms/fields/FormInputRaw.unsafe";
+import { InputLayout } from "./forms";
```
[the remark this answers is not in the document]
desfero: hm, i think it's an intellij but. `index.ts` imports are very handy here because we are not tight to folder structure inside shared.

app/components/shared/forms/fields/FormFieldDate.tsx
desfero: this is a good idea, let's add it to our forms technical dept ticket

desfero: Done, it's now tracked here https://github.com/Neufund/platform-frontend/issues/1484

PR #3149, opened 2019-06-22
[All devs approval required] Update react to the one with hooks

desfero:
This release introduces `hooks`, so please have a look at hooks docs 
https://reactjs.org/docs/hooks-intro.html. 
There is no reason to rewrite existing components to hooks but I think it's worth to use hooks for new components.

As a side note, the amount of changes in `yarn.lock` is because of duplicated dependencies that were removed by `yarn dedupe`


app/components/shared/buttons/Button.tsx
```diff
@@ -48,16 +48,12 @@ export enum ButtonTextPosition {
   RIGHT = "text-right",
 }
 
-export interface IGeneralButton {
-  onClick?: (e: React.MouseEvent<HTMLButtonElement>) => void;
-  disabled?: boolean;
-}
+export type TGeneralButton = React.ButtonHTMLAttributes<HTMLButtonElement>;
```
drakkein: <3

PR #3150, opened 2019-06-22
Refactor auth, bookbuilding-flow, contracts, eto-documents and eto-flow modules to use new action creator

desfero:
This is the first part of the final refactoring of our action creators to be more type-safe and easier to navigate. I don't want to change all of them in one pr as this may be difficult to review properly.

Reviewers, please pay attention to whether the actions are replaced correctly as this is manual refactoring.


app/modules/eto-flow/reducer.ts
```diff
@@ -33,14 +33,14 @@ export const etoFlowReducer: AppReducer<IEtoFlowState> = (
...
+    case actions.etoFlow.loadDataStop.getType():
       return {
         ...state,
         loading: false,
         saving: false,
       };
-    case "ETO_FLOW_SAVE_DATA_START":
-    case "ETO_FLOW_SUBMIT_DATA_START":
+    case actions.etoFlow.saveDataStart.getType():
+    case actions.etoFlow.submitDataStart.getType():
```
AlexeyTuboltsev: I wonder why do we have those two? Is there a difference? Maybe leave one?

Issue #2955, opened 2019-05-16
Pitch deck is not working without youtube video


desfero:
**Description**
Right now if you try to add only slideshare presentation without any youtube video it won't display any presentation on the eto profile.

**Environment**
-

**Steps To Reproduce**
* Log in as an issuer 
* go to Media forms
* add slideshare link and save
* go to eto view
* there is no slideshare (pitch deck) displayed

**Actual behaviour**
there is only `pitch deck` tab title without content

**Expected behavior**
slideshare should be properly displayed below tab title

**Comments, context, logs**
seems like we should filter `falsy` children values here https://github.com/Neufund/platform-frontend/blob/7a91812bd3c40c1e6f21bd83af5d08b119d4a11d/app/components/shared/Tabs.tsx#L93-L104

as we are trying to get `activeIndex` of 0, where `children` array equals to `[false, slideshareTab]` because `isYouTubeVideoAvailable` is set to false.
**Screenshots**

glupiawymowka:
*tested on stage*
it's a bit confusing how should the link look tho

PR #3151, opened 2019-06-23
#2955 - Empty tabs fix

drakkein:


app/components/shared/Tabs.tsx
```diff
@@ -86,20 +86,22 @@ class Tabs extends React.Component<ITabsProps & CommonHtmlProps> {
     const { children, layoutSize, layoutPosition, className } = this.props;
     const { activeIndex } = this.state;
 
-    if (!children) {
+    // During props destruction TS loses correct typing cast to TComponent array is required
+    const tabs =
+      children && (children.filter(child => child && typeof child !== "boolean") as TComponent[]);
+
+    if (!tabs) {
       return;
```
desfero: If it's a component then we should return `null` here IMO

app/components/shared/Tabs.tsx
```diff
@@ -86,20 +86,22 @@ class Tabs extends React.Component<ITabsProps & CommonHtmlProps> {
...
     }
 
     return (
       <>
         <div className={cn(styles.tabsWrapper, layoutSize, className)}>
           <div className={cn(styles.tabsOverflowWrapper, layoutPosition)}>
-            {children.map(
-              (child, index) => child && typeof child !== "boolean" && this.renderTab(index, child),
-            )}
+            {tabs.map((child: TComponent, index: number) => child && this.renderTab(index, child))}
```
desfero: What's the reason of `child &&`? Can we now just call `this.renderTab`?

app/components/shared/Tabs.tsx
```diff
@@ -86,20 +86,22 @@ class Tabs extends React.Component<ITabsProps & CommonHtmlProps> {
     const { children, layoutSize, layoutPosition, className } = this.props;
     const { activeIndex } = this.state;
 
-    if (!children) {
+    // During props destruction TS loses correct typing cast to TComponent array is required
+    const tabs =
+      children && (children.filter(child => child && typeof child !== "boolean") as TComponent[]);
```
desfero: What about?
```
 React.Children.toArray(children).filter(React.isValidElement) as TComponent[]
```
This is something we use to filter in tables

desfero: btw, can you also please 
image
i just caught that we put class constructor over prop types there

app/components/shared/Tabs.stories.tsx
```diff
@@ -40,4 +52,5 @@ storiesOf("Basic UI/NewTabs", module)
     </Tabs>
   ))
   .add("layout: position right", () => <Tabs layoutPosition="right">{tabbedContent}</Tabs>)
-  .add("layout: position center", () => <Tabs layoutPosition="center">{tabbedContent}</Tabs>);
+  .add("layout: position center", () => <Tabs layoutPosition="center">{tabbedContent}</Tabs>)
+  .add("tabs with first tab empty", () => <Tabs>{withEmptyTabs}</Tabs>);
```
desfero: can we have an example if array is empty and if array contains a couple of not valid values `[false, undefined, null]` to see how it filters? It's even better to have a unit test for this but it's up to you

Issue #2195, opened 2018-12-06
REFUNDED flow including ETO timeline, summary, pending transaction and refund complete


rudolfix:
hey! we have smth like this on stage now
![image](https://user-images.githubusercontent.com/17202864/49608815-f7f2d800-f999-11e8-89f7-b4f5ecaa5fb0.png)

Can we have a design for what to do on REFUND?

falinebambi:
Overview of changes:
- Claim Refund State in the last section of widget 
- Status changed to 'Refund' with red dot
- Timeline in Refund state 

Zeplin link: 
https://zpl.io/a7OqW3K

Designs: 
![ETO_refund](https://user-images.githubusercontent.com/17235402/55711840-deb47380-59ed-11e9-9324-a228bdd755f6.png)

rudolfix:
@falinebambi transaction modal design

glupiawymowka:
hey @desfero here I can clearly see, that you've fixed the timeline for REFUND state, and also hid it for mobile. 
However, both of these are happening again

glupiawymowka:
no fixture with refund state on stage.

PR #3158, opened 2019-06-26
#2195 - ETO refund

drakkein:


app/components/eto/overview/EtoOverviewStatus/ClaimRefundWidget/RefundWidget.tsx
```diff
@@ -1,28 +1,84 @@
...
+    <p className={style.message}>
+      <FormattedMessage id="shared-component.eto-overview.refund" />
+    </p>
+    {canClaimRefund && (
+      <Button onClick={() => onRefund(etoId)} data-test-id="eto-overview-claim-your-refund">
+        <FormattedMessage id="shared-component.eto-overview.claim-your-refund" />
       </Button>
     )}
+    {alreadyRefunded && !canClaimRefund && (
+      <small className="text-center" data-test-id="eto-overview-refund-claimed">
```
drakkein: For now we display simple message. 
@andywalton I do not want to blindly waste time, so if it's ok, after we find good way to show additional information I will add them as separate PR

app/modules/tx/sender/sagas.ts
```diff
@@ -203,9 +203,14 @@ function* sendTxSubSaga({ web3Manager }: TGlobalDependencies): any {
...
-
-    yield put(actions.txSender.txSenderSigned(txHash, type));
+    const txTimestamp = yield neuCall(markTransactionAsPending, {
+      txHash,
+      type,
+      txData,
+      txAdditionalData,
+    });
+
+    yield put(actions.txSender.txSenderSigned(txHash, type, txTimestamp));
```
drakkein: Small addition to have timestamp immediately in `txSender`state

intl/locales/en-en.json
```diff
@@ -1121,6 +1122,13 @@
   "user-claim-flow.summary": "CLAIM TOKEN SUMMARY",
   "user-claim-flow.summary.explanation": "When you confirm this transaction, you are taking possession of your Equity Tokens and NEU tokens. This will allow you to claim dividends and platform proceeds. You may access your finalized investment agreements below.",
   "user-claim-flow.token-name": "TOKEN",
+  "user-refund-flow.amount": "Amount refunded",
+  "user-refund-flow.confirm": "Confirm",
+  "user-refund-flow.cost": "Transaction const (gas price x gas limit)",
```
AlexeyTuboltsev: hey could you please replace `x` with the multiplication sign (×)? 
X instead of multiplication sign was used only on the old typewriters because there was no other way. But we're on computers now, we can use proper glyphs for this kind of things.

app/components/eto/overview/EtoOverviewStatus/ClaimRefundWidget/RefundWidget.tsx
```diff
@@ -1,28 +1,84 @@
...
+      doesInvestorInvest: selectHasInvestorTicket(state, props.etoId),
+      investorTicket: selectInvestorTicket(state, props.etoId),
+      userType: selectUserType(state),
+    }),
+    dispatchToProps: dispatch => ({
+      onRefund: (etoId: string) => dispatch(actions.txTransactions.startInvestorRefund(etoId)),
+    }),
+  }),
+  withProps<IWithProps, IExternalProps & IStateProps & IDispatchProps>(props => ({
+    canClaimRefund:
```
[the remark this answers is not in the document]
drakkein: No, it's not possible. Your idea with enums is great, so let's go with it.

app/components/modals/tx-sender/investor-refund/RefundDetails.tsx
```diff
@@ -0,0 +1,84 @@
...
+      caption={<FormattedMessage id="user-refund-flow.amount" />}
+      value={
+        !isZero(additionalData.amountEth) ? (
+          <MoneyNew
+            data-test-id="modals.tx-sender.user-refund-flow.amount.eth"
+            value={additionalData.amountEth}
+            moneyFormat={ECurrency.ETH}
+            inputFormat={ENumberInputFormat.ULPS}
+            outputFormat={ENumberOutputFormat.ONLY_NONZERO_DECIMALS}
+            currencySymbol={ECurrencySymbol.CODE}
```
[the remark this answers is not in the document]
drakkein: Lets have it done as separate issue

app/components/modals/tx-sender/investor-refund/RefundDetails.tsx
```diff
@@ -0,0 +1,84 @@
...
+              moneyFormat={ECurrency.ETH}
+              inputFormat={ENumberInputFormat.ULPS}
+              outputFormat={ENumberOutputFormat.FULL}
+            />
+          </>
+        }
+      />
+    )}
+
+    {children}
```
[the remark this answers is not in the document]
drakkein: This is an leftover, just removed it

app/modules/tx/sender/actions.ts
```diff
@@ -18,8 +18,8 @@ export const txSenderActions = {
   txSenderAccept: () => createSimpleAction("TX_SENDER_ACCEPT"),
   txSenderChange: (type: ETxSenderType) => createAction("TX_SENDER_CHANGE", { type }),
   // Signer actions
-  txSenderSigned: (txHash: string, type: ETxSenderType) =>
-    createAction("TX_SENDER_SIGNED", { txHash, type }),
+  txSenderSigned: (txHash: string, type: ETxSenderType, txTimestamp: number | undefined) =>
```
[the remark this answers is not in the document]
drakkein: yes, you are right

app/components/eto/overview/EtoOverviewStatus/ClaimRefundWidget/RefundWidget.tsx
```diff
@@ -1,28 +1,105 @@
...
+
+interface IDispatchProps {
+  onRefund: (etoId: string) => void;
+}
+
+interface IWithProps {
+  claimStatus: EUserRefundStatus;
+}
+
+enum EUserRefundStatus {
```
AlexeyTuboltsev: 👍

app/components/eto/overview/EtoOverviewStatus/ClaimRefundWidget/RefundWidget.tsx
```diff
@@ -1,28 +1,105 @@
...
+      doesInvestorInvest: selectHasInvestorTicket(state, props.etoId),
+      investorTicket: selectInvestorTicket(state, props.etoId),
+      userType: selectUserType(state),
+    }),
+    dispatchToProps: dispatch => ({
+      onRefund: (etoId: string) => dispatch(actions.txTransactions.startInvestorRefund(etoId)),
+    }),
+  }),
+  withProps<IWithProps, IExternalProps & IStateProps & IDispatchProps>(props => ({
+    claimStatus:
```
AlexeyTuboltsev: oh, could you rewrite this as a normal if/else please? Nested ternaries are super hard to read and, to my knowledge, are considered a bad practice overall. 
Maybe it's a good idea to extract it as a separate utility function

app/components/modals/tx-sender/investor-refund/RefundDetails.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <MoneyNew
+            data-test-id="modals.tx-sender.user-refund-flow.amount.eth"
+            value={additionalData.amountEth}
+            moneyFormat={ECurrency.ETH}
+            inputFormat={ENumberInputFormat.ULPS}
+            outputFormat={ENumberOutputFormat.ONLY_NONZERO_DECIMALS}
+          />
+        ) : (
+          <MoneyNew
+            data-test-id="modals.tx-sender.user-refund-flow.amount.neur"
```
AlexeyTuboltsev: btw moneyFormat is called valueType after my latest PR. Don't get afraid when merging, it's the same here, just renamed.

And I will remove the obsolete stuff from MoneyNew, never mind

app/components/modals/tx-sender/investor-refund/RefundSummary.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type IComponentProps = IDispatchProps & IStateProps;
+
+export const RefundSummaryLayout: React.FunctionComponent<IComponentProps> = ({
+  additionalData,
+  txData,
+  onAccept,
+}) => {
+  if (!additionalData) {
```
AlexeyTuboltsev: I suggest to move this to `connect()` to a `branch()`. I think we really should aim for the components to serve **only** as the visualization functions for the valid and clean data. All invalid, undefined and otherwise dirty data, business logic, calculations, etc, should be handled before, in `compose( branch() )`.

app/components/modals/tx-sender/investor-refund/RefundSummary.tsx
```diff
@@ -0,0 +1,73 @@
...
+          >
+            <FormattedMessage id="user-refund-flow.confirm" />
+          </ButtonArrowRight>
+        </Col>
+      </Row>
+    </Container>
+  );
+};
+
+export const RefundSummary = compose<IComponentProps, {}>(
```
AlexeyTuboltsev: maybe something like this? After this you don't need any conditionals in components:
```
export const RefundSummary = compose<IComponentProps, {}>(
  appConnect<IStateProps, IDispatchProps>({
    stateToProps: state => {
      const txData = selectTxDetails(state);
      const additionalData = selectTxAdditionalData<ETxSenderType.INVESTOR_REFUND>(state);
      if (txData && additionalData) {
        return ({
          txData,
          additionalData
        })
      } else {
        throw new Error("Missing refund data");
      }},
      onAccept: () => dispatch(actions.txSender.txSenderAccept()),
    }),
  }),
)(RefundSummaryLayout);
```

app/modules/tx/sender/sagas.ts
AlexeyTuboltsev: 👍

app/components/eto/overview/EtoOverviewStatus/ClaimRefundWidget/RefundWidget.tsx
```diff
@@ -1,28 +1,105 @@
...
+enum EUserRefundStatus {
+  CANNOT_CLAIM = "cannot_claim",
+  CAN_CLAIM = "can_claim",
+  CLAIMED = "claimed",
 }
 
-const RefundWidgetLayout: React.FunctionComponent<IExternalProps> = ({ canClaimToken }) => (
+type TComponentProps = IExternalProps & IStateProps & IDispatchProps & IWithProps;
+
+const RefundActionWidget: React.FunctionComponent<IWithProps & { onRefund: () => void }> = ({
```
desfero: why not `IWithProps & IDispatchProps`.

app/components/eto/overview/EtoOverviewStatus/ClaimRefundWidget/RefundWidget.tsx
```diff
@@ -1,28 +1,105 @@
...
+}
+
+interface IStateProps {
+  doesInvestorInvest: boolean;
+  userType: EUserType | undefined;
+  investorTicket?: IInvestorTicket;
+}
+
+interface IDispatchProps {
+  onRefund: (etoId: string) => void;
```
desfero: You can update `onRefund` to take `etoId` from props. So then i can just pass props throught and always use the same interface. See comment on line `46`. What do you think?

app/components/eto/overview/EtoOverviewStatus/ClaimRefundWidget/RefundWidget.tsx
desfero: agree, i'm not able to understand what's the logic here. Would be good to have it as a util function or as a redux selector

app/components/modals/tx-sender/investor-refund/RefundDetails.tsx
```diff
@@ -0,0 +1,78 @@
...
+  <InfoList className={className}>
+    <InfoRow
+      caption={<FormattedMessage id="user-refund-flow.token-name" />}
+      value={`${additionalData.tokenName} (${additionalData.tokenSymbol})`}
+    />
+
+    <InfoRow
+      caption={<FormattedMessage id="user-refund-flow.amount" />}
+      value={
+        !isZero(additionalData.amountEth) ? (
```
desfero: can we reverse the logic here? It generally easier to read `non` negated conditions. what do you think?

app/components/modals/tx-sender/investor-refund/RefundSummary.tsx
desfero: yeah, agree, we should move this to a `branch`

app/components/modals/tx-sender/investor-refund/RefundSummary.tsx
desfero: but generally how it's possible that refund data are missing?

app/modules/tx/sender/sagas.ts
desfero: 🥇

app/modules/tx/transactions/refund/saga.ts
```diff
@@ -0,0 +1,69 @@
...
+import { neuCall } from "../../../sagasUtils";
+import { selectEtherPriceEur } from "../../../shared/tokenPrice/selectors";
+import { selectEthereumAddressWithChecksum } from "../../../web3/selectors";
+import { selectTxGasCostEthUlps } from "../../sender/selectors";
+import { ETxSenderType } from "../../types";
+
+export function* generateGetRefundTransaction(
+  { contractsService, web3Manager }: TGlobalDependencies,
+  etoId: string,
+): any {
```
desfero: `: Iterator<any>`

app/modules/tx/transactions/refund/saga.ts
```diff
@@ -0,0 +1,69 @@
...
+  const userAddress = yield select(selectEthereumAddressWithChecksum);
```
desfero: please add types for all yielded objects as they are typed by default as any

app/modules/tx/transactions/sagas.ts
```diff
@@ -203,6 +204,25 @@ export function* removePendingTransaction({ logger }: TGlobalDependencies): any
   }
 }
 
+export function* etoRefundSaga(
+  { logger }: TGlobalDependencies,
+  action: TActionFromCreator<typeof actions.txTransactions.startInvestorRefund>,
+): any {
```
desfero: Let's migrate to `Iterator<any>` as this is the more correct type. Then in future, it's easier to replace `Iterator<any>` with more appropriate type

app/test-e2e/eto/Refund.spec.e2e.ts
```diff
@@ -0,0 +1,50 @@
+import BigNumber from "bignumber.js";
+
+import { etoPublicViewByIdLinkLegacy } from "../../components/appRouteUtils";
+import { etoFixtureAddressByName, tid } from "../utils";
+import { acceptWallet, getLockedWalletEthAmount, parseAmount } from "../utils/index";
```
desfero: this is the same import as the one above, we can combine them

app/test-e2e/eto/Refund.spec.e2e.ts
```diff
@@ -0,0 +1,50 @@
...
+            // balance after refund should be increased by the refund amount
+            // there can be rounding issue so we assert balance with small delta
+            expect(balanceAfter.minus(balanceBefore.plus(amount))).to.be.bignumber.lessThan(0.01);
+          });
+        });
+      });
+
+      cy.visit(etoPublicViewByIdLinkLegacy(ETO_ID));
+
+      cy.get(tid("eto-overview-refund-claimed"));
```
desfero: cy.get().`should("exist")`

.circleci/config.yml
```diff
@@ -74,7 +74,10 @@ jobs:
       image: ubuntu-1604:201903-01
       docker_layer_caching: true
 
-    resource_class: xlarge
+    resource_class: large
```
rudolfix: that's applicabe only to docker instances, here's a VM

drakkein: yeah, I'm gonna remove it 👍

app/components/modals/tx-sender/investor-refund/RefundSummary.tsx
AlexeyTuboltsev: @desfero  the problem here is that our YupTS typings are super loose because they are essentially made for the forms where everything has to be optional for the user to be able to save incomplete data. The solution for this is to introduce a set of types for everything so that there's 1. normal data types with optional things only when they are really optional, everything else mandatory, 2. for forms - however loose they have to be. If we ever do it, we;d be able to get rid of tons of such `if/else` in components.

PR #3180, opened 2019-07-09
Enable KYCFlowIndividualIDnow and RegisterRedirectIfNoVerifiedEmail e2e

desfero:


app/modules/kyc/sagas.ts
```diff
@@ -248,7 +248,11 @@ function* startIndividualInstantId({
 }: TGlobalDependencies): Iterator<any> {
   try {
     const result: IHttpResponse<IKycRequestState> = yield apiKycService.startInstantId();
-    if (result.body.redirectUrl) window.location.replace(result.body.redirectUrl);
+
+    if (result.body.redirectUrl) {
```
AlexeyTuboltsev: and thanks for the {} after `if`!

desfero: i have also used to `{}` after `if` even if inside there is just one expression

Issue #2821, opened 2019-04-29
Trans-History #1. Basic History Implementation


Louisswryan:
The final Transaction History will need to be rolled out in stages.

The first of those stages is the basic History interface shown below:

https://app.zeplin.io/project/5a8a92c89c1a166a6a6e8f37/screen/5cc6f62c6c477a3dc9b340ff

The listed events will be shown in order from most to least recent.

When a value is brought into the platform, it is represented with bold green text
When a value leaves the platform, it is represented with regular black text

For now, the included events are as follows:

In Green 

-nEUR Purchase (deposit)
-Payout/Dividend nEUR/EtherToken
-Any token transfer into platform
-Equity Token Claimed (ETO)
-ETO Refund: nEUR or Ether

In Black

-Investment nEUR/EtherToken
-nEUR Redeem
-nEUR confiscation
-Any token transfer out of platform
-ICBM Participation

The width of the History card is imprecise here buit it should meet the edges of the top cards in wallet.

If you have questions please ask : D

rudolfix:
There's backend fixture available ask @tl-neufund for details

rudolfix:
** tech details**
api root: `/api/analytics-api/`

getting the transactions list:
`/transactions/me`
returns a list of maximum of 100 transactions, less is possible (due to post-filtering)
with the transactions we return following fields
`version`: when present is used to refresh transaction list, it's always present when you request first page of transactions via this method
`before_transaction`: when present it means that more data is available (next page).

refreshing the transaction list:
`/transactions/new/me`
where you pass the most recent `version`. it will return **new** and **updated** transactions so you need to add or update existing transactions (can happen for example if your eurt withdrawal is settled)
returns new `version`, use it
you can poll on this method each 5 seconds, you can do it in the background when in wallet view and when you enter wallet view and some transactions are in state

getting next page:
`/transactions/me`
pass `before_transaction` you've got from previous call. transactions will not overlap with what you've got previously. 
it very rare situations you may get 0 transactions but there's still `before_transaction` present so it means that there's more data. that's because of post-filtering. should not happen really.

use fixtures:
only on `io`/`localhost` pass `use_fixtures` `true` to `/transactions/me` to get a list of all possible transaction types with several interesting states. we can add more on demand. good for e2e test

**unification of transaction types and extra data**
@desfero it seems that we have very similar schema that is used to store transactions in `user/me`. please ping me so we can
* unify transaction types (names that we use)
[...]

PR #3189, opened 2019-07-12
#2821 TX history modals step 1

desfero:
Demo:
Jul-12-2019 15-34-07


app/components/shared/table/NewTable.module.scss
```diff
@@ -1,19 +1,54 @@
-.table-wrapper {
-  overflow: auto;
-
-  &:not(:global(.keep-rhythm)) {
+@mixin hover-and-focus {
```
desfero: This is really ugly but I don't know a better way to add green line border on `hover` and 'focus`(see zeplin zpl.io/bLdEQrd). I will really appreciate if someone can find a better solution for this.

app/utils/enums/keys.enum.ts
```diff
@@ -0,0 +1,8 @@
+/**
```
desfero: Maybe we can find a better place for enums? I didn't find any so create a new folder `app/utils/enums`. Let me know if you think it's okay.

app/components/shared/hooks/useButtonRole.ts
```diff
@@ -0,0 +1,37 @@
...
+) =>
+  React.useMemo(() => {
+    if (onClick) {
+      return {
+        tabIndex: 0,
+        role: "button",
+        onClick,
+        onKeyDown: (event: React.KeyboardEvent<T>) => {
+          // Check to see if space or enter were pressed
+          if (event.key === EKeys.SPACE || event.key === EKeys.ENTER) {
```
Moejoe90: Do you think its a good idea to generate an `onClick` event even for `SPACE` button?

app/components/shared/table/NewTable.tsx
```diff
@@ -86,39 +96,46 @@ const Table: React.FunctionComponent<TProps> = ({
...
 
+const Table: React.FunctionComponent<TProps> = props => (
+  <div className={styles.wrapper}>
+    <TableLayout {...props} />
+  </div>
+);
+
 const NewTable: React.FunctionComponent<TProps> = props => (
-  <Panel narrow={true} className={styles.panel}>
+  <div className={styles.panel}>
```
Moejoe90: Why not use `Panel`? 

Because this means that either `Panel` is not too generic or we might need to just get rid of it

app/components/wallet/transactions-history/TransactionDetailsModal.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+
+import { TTxHistory } from "../../../modules/tx-history/types";
+
+interface IExternalProps {
+  closeModal: () => void;
+  transaction: TTxHistory;
+}
+
+const TransactionDetailsModal: React.FunctionComponent<IExternalProps> = ({ transaction }) => (
```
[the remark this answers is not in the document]
desfero: yes, it's already in progress

app/components/shared/hooks/useButtonRole.ts
desfero: It's probably required functionality to make proper accessible elements. Try to focus some normal button or link and press space, it will then normally call `onClick` handler. Have a look at mdn docs about role="button" in `Required Javascript Features` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Required_JavaScript_Features

app/components/shared/table/NewTable.tsx
desfero: The problem here is that this left border on hover complicates everything now. As we need to have `padding on the left attached to the table container to implement this border-left. So in this example, it's easier to just copy panel background and border to a table than overwriting panel styles.

Issue #3185, opened 2019-07-10
[TOPNAV] - feature improvements - please don't close yet 


glupiawymowka:
This new top navigation bar is awesome! And still needs some improvements

**1. Notifications**
the current notifications are designed only to show a pending transaction. 
We had the logics of multiple statuses in notifications, so we simply want it back + the new icons 

No pending transactions - we already do have a message to show on hover, let's keep it) 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/0a5fb145-65e3-4f4e-9e85-04b8ab5bfa3f)

Pending transaction - we already do have a message to show on hover. 
If the transaction pending is external (for example, done in Metamask), we do have a separate summary. If there is no `on-hover message` for external transaction, the regular one + "external" will do
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/7e1f0394-4f5c-4950-8a62-3cd35e805eb7)

Error + message on hover 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/85471d73-ac2c-4421-a040-b5bf50cc4c6f)

Success +message on hover 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/04363e7c-79dd-4196-9a9b-72adb6961b84)

The logic should be as before the changes, so : 
- [x] pending transaction status changes into error or success, but never disappears completely 
- [x] error or success notification changes into "no pending transactions" when the user: 
- has the transaction summary open and follows the whole flow as it happens
- already clicked on the notification icon and saw the summary of error/success 

**2. Buttons positioned unevenly**

- [x] in the navbar (mind the spacing between  Dashboard & Portfolio, then between Portfolio & Wallet) 

![image](https://user-images.githubusercontent.com/18315589/60971519-a55d0e00-a324-11e9-835a-affc820153be.png)

- [ ] for tablets
Play with the screen resolutions <1000 x 400 to see it clearly  

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/296a5b70-8889-4fb1-b719-9600d121e166)

**3. Wrong step triggered from ETO Page**

When an inactive user takes the following journey: 
Landing Page => company ETO Page => ``Login`` OR ``Get Started``
They're being forwarded to the Landing Page first.
 Login or Register page appears only on the second click (so, if the user already is on the landing page) 

[...]

PR #3195, opened 2019-07-15
#3185 top navbar fixes

AlexeyTuboltsev:
small fixes to top navigation bar and related stuff. Mostly css tweaks
couldn't reproduce the points 2.2 and 3 of the ticket so there are no changes related to them.


app/assets/img/icon_txn_status_error.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 55.2 (78181) - https://sketchapp.com -->
+    <title>icon_txn_status_error</title>
```
desfero: Can we remove all kind of comments, `id` `title` and `desc` from all SVG's? Having `title` is bad for accessibility as it's announced by a screen reader.

app/components/layouts/header/PendingTransactionStatus.module.scss
```diff
@@ -0,0 +1,17 @@
...
+    transform: rotate(180deg);
+  }
+}
+
+.icon {
+  width: px-to-rem(20px);
+}
+
+.pending-transaction {
+  animation: rotation 1s linear 0s infinite normal forwards running;
```
desfero: @AlexeyTuboltsev I'm just curious, is it in some way better to specify default values (`normal forwards running`)?

AlexeyTuboltsev: wow, I just used animation for the first time in my life. So it's a good old copy-paste

Issue #3190, opened 2019-07-12
Platform landing page visual fix


andywalton:
Please fix the landing page to the below layout. It's more inline with the original design - see: https://platform.neufund.org/

![Screenshot 2019-07-12 14.48.23.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/0168dd25-e56c-4da5-a3aa-01f53fdd3d73)

The current page on dev `https://platform.neufund..io` has a few problems:

![Screenshot 2019-07-12 14.48.51.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/c12192d1-26d5-4dcc-b6bd-76ee3380ddef)

1) The herospace section is boxed in. The section bg should be `100%` the width of the page and the content inside should be centered and have a maximum width of `1140px`

2) The Investment opportunities sections background colour needs to change to `#f3f5f7`.

After these changes it should look as follows:
![Screenshot 2019-07-12 14.48.30.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/c20f0e7f-9800-4435-ae93-e9b852a472b0)

3) Adjust the spacing between the heading: `Equity investments made better` and the yellow detail line above it. It should be the same as what we have on .org currently.

I looks to be `2rem. (32px)`.

After this change the heading should look as follows:
<img width="640" alt="Screenshot 2019-07-12 15 13 15" src="https://user-images.githubusercontent.com/609657/61130824-e0437b00-a4b7-11e9-9155-3d517b886071.png">

glupiawymowka:
*tested on stage*

PR #3197, opened 2019-07-15
#3190 Fix platform landing page

desfero:
This pr:
- Adds a new boolean prop to `Layout` `contentFullWidth`.
- Global rename of `LayoutNew` to `Layout`
- Minor css cleanup and fixes in `Landing` component

After build is green:
- [x] Make sure visual regression is okay


app/components/layouts/Content.tsx
```diff
@@ -6,25 +6,16 @@ import { LoadingIndicator } from "../shared/loading-indicator/LoadingIndicator";
...
+  fullWidth,
 }) => (
-  <div className={cn(styles.landingContent, className)}>
-    <React.Suspense fallback={<LoadingIndicator />}>{children}</React.Suspense>
-  </div>
-);
-
-export const Content: React.FunctionComponent<CommonHtmlProps> = ({ children, className }) => (
-  <div className={cn(styles.content, className)}>
+  <div className={cn(styles.content, { [styles.contentFullWidth]: fullWidth }, className)}>
```
AlexeyTuboltsev: this is a nice idea, but can be a bit confusing. It took me some time to understand how it works with `not`-selector etc. I'd do it with normal styles, written out in module.scss and applied, but maybe it's ok so too.

app/components/layouts/Content.tsx
```diff
@@ -6,25 +6,16 @@ import { LoadingIndicator } from "../shared/loading-indicator/LoadingIndicator";
 
 import * as styles from "./Content.module.scss";
 
-export const LandingContent: React.FunctionComponent<CommonHtmlProps> = ({
+type TExternalProps = {
+  fullWidth?: boolean;
```
AlexeyTuboltsev: I would make it an enum, like EContentWidth {Full | Constrained } and set a default param on Content to Full.
Not a big difference now, but it's my personal idiosyncrasy - `foo?:boolean` mostly means/ leads to a nasty mess in the code, especially when the number of parameters starts to grow. What do you think?

app/components/layouts/Layout.tsx
```diff
@@ -22,28 +22,30 @@ interface IStateProps {
 }
 
 interface ILayoutUnauthProps {
-  hideHeaderCtaButtons: boolean;
+  hideHeaderCtaButtons?: boolean;
```
AlexeyTuboltsev: but why optional? Optional boolean is 1) a complete nonsense from data modelling point of view 🤓  2) leads to a nasty mess when people start adding parameters (which is inevitable). It's not hard to supply this param IMO, why skip it? 
If you think it's too much work 😄, a good solution would be to build two components, with and without buttons, since there's a valid use case for both.

desfero: But it was optional before (see line 28), I have just removed duplicates

app/components/layouts/Content.tsx
desfero: done

app/components/layouts/Content.tsx
desfero: done

app/components/layouts/Layout.tsx
AlexeyTuboltsev: 🤦‍♂ My care about backwards compat. never brings any good...

PR #3200, opened 2019-07-15
Fix dead code elimination

desfero:


app/config/constants.ts
```diff
@@ -64,12 +64,6 @@ export const TOAST_COMPONENT_DELAY = 4000;
 
 export const BROWSER_WALLET_RECONNECT_INTERVAL = 1000;
 
-/*
- * General constants•
- */
-export const IS_CYPRESS = !!process.env.IS_CYPRESS;
```
desfero: Webpack is not able to correctly mark this as `true` or `false` when it's imported because of the way javascript `import` works.

patches/bitcore-mnemonic+8.3.3.patch
```diff
@@ -1,13 +1,13 @@
...
-index 513eb36..dba006d 100644
+index 513eb36..94de870 100644
 --- a/node_modules/bitcore-mnemonic/lib/mnemonic.js
 +++ b/node_modules/bitcore-mnemonic/lib/mnemonic.js
 @@ -94,7 +94,7 @@ var Mnemonic = function(data, wordlist) {
    });
  };
  
 -Mnemonic.Words = require('./words');
-+Mnemonic.Words = {"ENGLISH": require('./words').ENGLISH};
```
desfero: If you `require` (commonjs module) `./worlds` then it's not possible to eliminate unused variables because of the way commonjs works. So in the above case, the correct approach is to require specific files

Issue #3126, opened 2019-06-18
ETH Withdrawal UI Review


Louisswryan:
Project Link - https://zpl.io/2GAJWer

Hello,

Designs are ready for the ETH Withdraw feature.

Basic transfer editor

NEW TRANSFER
- https://app.zeplin.io/project/5d10c5d1bb83d271220e0af7/dashboard?seid=5d1240cca41ede711b2f2877

Error states that appear if entered information is problematic

ERROR STATES
- https://app.zeplin.io/project/5d10c5d1bb83d271220e0af7/dashboard?seid=5d1240dddb07c27116939c93

Checked box state for certain errors which activates CTA

BOX CHECKED
- https://app.zeplin.io/project/5d10c5d1bb83d271220e0af7/dashboard?seid=5d1b3dba48d5685ac008a658

Summary page with status updates for Pending and Completion

SUMMARY
- https://app.zeplin.io/project/5d10c5d1bb83d271220e0af7/dashboard?seid=5d124126281c595a1ed71e27

Please let me know if there are any issues.

Thanks!

Moejoe90:
@Louisswryan 
As discussed there are a couple of things we will need to change and for design to figure out 

1 - Having the `Transfer all` as a feature or check box once activated the user shouldn't be able to enter any amount 

2 - Due the big variation in gas limit depending on the given address we need to disable the amount or somehow force the user to enter the address first why 

without that we can't know if the user has enough money etc

Moejoe90:
After discussion and review we have reached a couple of working actions points: 

1 - Gas Price will have the value of < 0.0001 instead of 0.0000 which doesn't make sense
2 - We will read the intent of the user as in if the value entered manually equals the highest rounded amount we will assume that the user wants to activate the full amount transfer 
3 - Once the user clicks transfer all button the highest rounded amount will be entered into the value field

4 - As long as no valid address was entered the user can still enter a value to transfer and also conduct a transfer all operation with the use of a random address as input. The transaction should never follow through until a valid address is entered but never the less. It is agreed that the cases where bad UX will show is very minimal till the point where we can discard. 

5 - Transfer all will have a separate flow. The flow will be activated once the intent to transfer the full amount is found 

6 - Tests that check entered values with the Real node will be written as a blue print for all important transactions. As its important to assert if the transactions generated by UX translate into correct transactions to the node

Louisswryan:
Hello

Louisswryan:
[...]

PR #3219, opened 2019-07-19
#3126 - Withdraw ui redesign

drakkein:


app/modules/tx/transactions/withdraw/sagas.ts
```diff
@@ -48,12 +78,50 @@ export function* generateEthWithdrawTransaction(
...
   }
 }
 
-export function* ethWithdrawFlow(_: TGlobalDependencies): any {
+export function* selectPossibleMaximum(value: string): Iterator<any> {
+  // Use full precision value if full amount was provided as value
+  const additionalData = yield select((state: IAppState) =>
+    selectTxAdditionalData<ETxSenderType.WITHDRAW>(state),
+  );
+  const maxEther =
```
rudolfix: you do the same check in validator/sagas:140, can we just set a flag in modal state that full amount is transferred out?

app/modules/tx/validator/sagas.ts
```diff
@@ -65,6 +95,79 @@ export function* validateGas(txDetails: ITxData): any {
...
+
+      return {
+        warnings,
+        inputValue: Q18.mul(payload.value).toString(), // use WEI formatted value
+        maximumAvailableEther: maximumAvailableEther.toString(),
+      };
+    } catch (error) {
+      return {
+        warnings,
+        inputValue: Q18.mul(payload.value).toString(), // use WEI formatted value
```
rudolfix: if this is generic try/catch you should also set inputValue to 0 I assume?

app/modules/tx/validator/sagas.ts
```diff
@@ -65,6 +95,79 @@ export function* validateGas(txDetails: ITxData): any {
...
+        decimalPlaces: selectDecimalPlaces(ECurrency.ETH),
+        inputFormat: ENumberInputFormat.ULPS,
+        roundingMode: ERoundingMode.DOWN,
+      });
+
+      if (
+        compareBigNumbers(
+          convertToBigInt(payload.value || "0"),
+          convertToBigInt(maxAvailableEtherFixed),
+        ) >= 0
```
rudolfix: let's define a constant MINIMUM_ETH_RESERVE_GAS_UNITS and set it to 1000000. then multiply by gasPrice. this warning should not only be present if we zero the account but also when we almost zero it

app/modules/web3/utils.ts
```diff
@@ -31,8 +32,18 @@ export function ethereumNetworkIdToNetworkName(networkId: EthereumNetworkId): st
   }
 }
 
-export const validateAddress = (value: string): boolean =>
-  !!value && isAddress(value.toUpperCase());
+export const validateAddress = (value: string): boolean => {
+  const v = value.substring(2);
+
+  // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md
```
rudolfix: I do not see where it's used and I do not see where is error message for it?

intl/locales/en-en.json
```diff
@@ -759,11 +759,21 @@
...
-  "modal.sent-eth.button": "Withdraw",
-  "modal.sent-eth.eth-amount-placeholder": "Please enter the ETH amount here",
-  "modal.sent-eth.title": "Withdraw your ETH",
-  "modal.sent-eth.to-address": "Send ETH to address:",
+  "modal.sent-eth.amount": "For amount of:",
+  "modal.sent-eth.available-balance": "Available Ether Balance",
+  "modal.sent-eth.button": "Continue",
+  "modal.sent-eth.button-loading": "Processing data",
+  "modal.sent-eth.change": "EDIT TRANSFER",
+  "modal.sent-eth.new-address": "This is the first time this address is recieving a payment. Would you like to continue?",
```
rudolfix: This is not a correct warning but anyway idea is good. I know that I'm adding stuff but
1. can we check also the balance of the receiving address AND transaction count is 0 we display this message
2. if just transaction count is 0 we display
"This is a new address that was not yet used by the owner. Would you like to continue"?

intl/locales/en-en.json
```diff
@@ -759,11 +759,21 @@
...
+  "modal.sent-eth.not-accepting-ether": "You are sending Ether to Smart Contract that does not accept Ether transfers.",
+  "modal.sent-eth.smart-contract-address": "This address is for a Smart Contract. Would you like to continue with this transfer?",
+  "modal.sent-eth.title": "New Eth Withdraw",
```
rudolfix: Rename to "Withdraw Ether". current text breaks by heart ;>

intl/locales/en-en.json
```diff
@@ -788,7 +798,8 @@
   "modals.tx-sender.user-claim-summary.download-confirmation": "I have downloaded the above agreements",
   "modals.tx-sender.user-claim.title": "Claim",
   "modals.tx-sender.withdraw-flow.withdraw-component.errors.not-ethereum-address": "This Ethereum Address is invalid",
-  "modals.tx-sender.withdraw-flow.withdraw-component.errors.value-higher-than-balance": "The amount exceeds your wallet balance",
+  "modals.tx-sender.withdraw-flow.withdraw-component.errors.value-higher-than-balance": "Insufficient Eth for Transfer",
```
rudolfix: rename to  "You do not have sufficient Ether to pay transaction costs." also why this is not universal text shared by all the transaction modals?

app/modules/web3/utils.ts
drakkein: This is connected to `getWithdrawFormSchema` in _Withdraw.unsafe.tsx_ if condition is not met it will display `modals.tx-sender.withdraw-flow.withdraw-component.errors.not-ethereum-address`

intl/locales/en-en.json
glupiawymowka: who is the owner here? 
Please, if implementing this, stick to "you"-attitude towards the user

app/modules/web3/utils.ts
rudolfix: OK - I think it's good enough

intl/locales/en-en.json
glupiawymowka: ETH, not Ether, always :D (sorry)

app/modules/tx/transactions/withdraw/sagas.ts
drakkein: This is used to take full precision value, the validator saga check gonna be changed to use `MINIMUM_ETH_RESERVE_GAS_UNITS` so this is not relevant anymore

rudolfix:
Information on "ERROR" state
transaction can get into ERROR state for two reasons:
1. transaction is reverted: we get transaction receipt with the status.
2. any other error before transaction was submitted for execution (there's a set of Parity errors).

We need to display detailed error message in every case

In case we have a receipt we should display optional infromation on transaction fee that was unfortunately paid. See comments to designs on ERRROR state

rudolfix:
how to verify if address is Neufund user.
using `IdentityRegistry` - we have a method we use to check if current user is verified on-chain
```
function* loadIdentityClaim({ contractsService }: TGlobalDependencies): Iterator<any> {
  const identityRegistry: IdentityRegistry = contractsService.identityRegistry;

  const loggedInUser: IUser = yield select<IAppState>(state => selectUser(state.auth));

  const claims: string = yield identityRegistry.getClaims(loggedInUser.userId);

  yield put(actions.kyc.kycSetClaims(deserializeClaims(claims)));
}
```
If we see that user `isVerfied` and not `isFrozen` then we:
- do not display any warnings that account is used for a first time or has 0 balance etc.
- we display a nice information that @Louisswryan will provide

app/modules/tx/validator/sagas.ts
drakkein: We want to preserve value set by user

intl/locales/en-en.json
drakkein: Most modals have their own instances of this error message with value `The amount exceeds your wallet balance`. For now I will use this one from your comment (with Ether changed to ETH :))

app/modules/sagasUtils.ts
```diff
@@ -111,3 +113,39 @@ export function* neuRestartIf(cancelAction: TType, saga: TSagaWithDeps, ...args:
...
+      let action = yield take(pattern);
+
+      while (true) {
+        const { debounced, latestAction } = yield race({
+          debounced: delay(ms),
+          latestAction: take(pattern),
+        });
+
+        if (debounced) {
+          yield neuFork(task, ...args, action);
```
Moejoe90: why `neuFork` and not `neuCall` 
I don't think we need to fork a new generator but just wait until its finished

app/modules/tx/transactions/withdraw/sagas.ts
```diff
@@ -48,12 +92,65 @@ export function* generateEthWithdrawTransaction(
...
+  }
+}
+
+export function* ethWithdrawFlow(_: TGlobalDependencies): Iterator<any> {
+  const maxEther = yield select(selectLiquidEtherBalance);
+  const account = yield new Web3Accounts().create();
+
+  const initialTxDetails = yield neuCall(generateEthWithdrawTransaction, {
+    to: account.address,
+    value: maxEther.toString(),
```
Moejoe90: The question is why selectLiquidEtherBalance is not returning strings directly?

app/test-e2e/pending-transactions/utils.ts
```diff
@@ -53,19 +64,18 @@ export const assertPendingWithdrawModal = (address: string, amount: string) => {
   cy.get(tid("modals.shared.tx-pending.modal")).should("exist");
 
   // should propagate correct data to modal
-  cy.get(tid("modals.tx-sender.withdraw-flow.summary.to")).contains(address);
-  cy.get(tid("modals.tx-sender.withdraw-flow.summary.value")).contains(amount);
-  cy.get(tid("modals.tx-sender.withdraw-flow.summary.cost")).contains(/0\.\d{4}/);
-  cy.get(tid("timestamp-row.timestamp")).should("exist");
```
Moejoe90: This is not in the designs anymore?

pstroinski:
Looking at the screens from Nikola:
- once you have Send ETH
- then you have ETH Send
- then you have Confirm Eth Send (small letters)

This should be definitely unified

app/modules/tx/validator/sagas.ts
```diff
@@ -139,9 +141,10 @@ export function* txProcessAdditionalData(
 
     const maxEtherFixed = toFixedPrecision({
       value: allEther,
-      decimalPlaces: selectDecimalPlaces(ECurrency.ETH),
+      decimalPlaces: 8,
```
Moejoe90: why is this hardcoded?

PR #3226, opened 2019-07-24
Refactors ETO Public View 

Moejoe90:
This also refactors `EtoView` component to use smaller components rather being one big component

still, some work needs to be done on the grid system 
image

As Youtube is not aligned with legal information but it's not a simple problem as size can change drastically depending on the sections before so this should have a separate ticket with a full charge in the grid request.

image
Also as an additional change marketing documents were changed to a full view as per @andywalton suggestion. 

Testing Notes: 

Please check how it looks with our ETO. The components, in my opinion, are not stable this will also require UX review


app/components/eto/public-view/CompanyDescription.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <Container columnSpan={EColumnSpan.TWO_COL}>
+        <DashboardHeading title={brandName} />
+      </Container>
+      <Panel columnSpan={EColumnSpan.TWO_COL}>
+        {companyDescription && (
+          <p className="mb-4" data-test-id="eto-view-company-description">
+            <Field name="companyDescription" value={companyDescription} />
+          </p>
+        )}
+        {keyQuoteInvestor && <p className={cn(styles.quote, "mb-4")}>{keyQuoteInvestor}</p>}
```
desfero: just as a nitpick, maybe we should use `blockquote` tag here?

app/components/eto/public-view/CompanyDescription.tsx
```diff
@@ -0,0 +1,45 @@
...
+        {companyWebsite && (
+          <p>
+            <FormattedMessage id="eto.public-view.eto-find-out" />
+            <ExternalLink href={normalizedUrl(companyWebsite)}>
+              {companyWebsite.split("//")[1] || DEFAULT_PLACEHOLDER}
```
desfero: Does anyone know what's the reason for `DEFAULT_PLACEHOLDER` here? On line `32` we already check whether `companyWebsite` is defined or no

app/components/eto/shared/EtoView.tsx
```diff
@@ -60,9 +53,6 @@ const EtoViewSchema = EtoCompanyInformationType.toYup().concat(EtoPitchType.toYu
 // TODO: Refactor to smaller components
```
desfero: Can be probably removed now?

app/components/shared/Accordion.tsx
```diff
@@ -70,3 +68,18 @@ export const Accordion: React.FunctionComponent<IAccordionProps> = ({ children,
       )}
   </div>
 );
+
+const AccordionField: React.FunctionComponent<{
+  title: TTranslatedString;
+  value: string | undefined;
+}> = ({ title, value }) => (
+  <>
```
desfero: Maybe we can get rid of `<>`?

intl/locales/en-en.json
```diff
@@ -446,6 +445,7 @@
   "eto.public-view.dividend-rights": "Dividend rights",
   "eto.public-view.documents.legal-documents": "Legal documents",
   "eto.public-view.documents.marketing-links": "Marketing documents",
+  "eto.public-view.eto-find-out": "Find out at more ",
```
desfero: Maybe we can use `Find out at more {companyWebsite}` and then use `values` for `FormattedMessage` to nest `ExternalLink` into `companyWebsite`. Then it's easier to understand the whole context of the message.

app/components/eto/shared/EtoView.tsx
Moejoe90: Indeed

app/components/shared/Accordion.tsx
Moejoe90: I couldn't I had to add it as typescript was returning an error after further research it turns out the component had to return a `fragment`

intl/locales/en-en.json
Moejoe90: that's a good idea actually

app/components/eto/public-view/CompanyDescription.tsx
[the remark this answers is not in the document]
desfero: good point, maybe we can use `https://developer.mozilla.org/en-US/docs/Web/API/URL` here and parse it to just display `hostname`?

app/components/eto/public-view/CompanyDescription.tsx
Moejoe90: I wouldn't do it in this pr as I don't want to touch the styles

app/components/eto/public-view/CompanyDescription.tsx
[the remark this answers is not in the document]
Moejoe90: Hmmm regardless I wouldn't touch that in this PR

Moejoe90: It was probably a protection check done separately with the assumption that there is no protection

Issue #2890, opened 2019-05-08
Trans-History #4. Modal Content


Louisswryan:
Hello @desfero ,

This part is probably the trickiest so maybe we can conference about it via Slack.

It is only tricky because of the number of states and potential layouts. For now, let's apply the google doc pretty much as it is laid out.

https://docs.google.com/document/d/1ileQzHNis9RBHmZc6bHE2cF9kj7PCyUXAZEFYXSQpnY/edit

Trevor suggested that for now, the simplest way to order information from the API is alphabetically so let's do that first and we can develop an information hierarchy later.

Below is a guide for how the content should fit within the modal

Modal Content guide:
https://zpl.io/2Egn4ll

This design leaves us with space for 7 fields within the modal.

You will notice that investment in the Google doc has 8 fields. The "is ETH or nEUR Investment" does not apply as we should put the currency of the investment always in the event title. So "Eth Investment" is the title. 

Furthermore "in" transactions are represented with a 'success green' in bold amount value to show that funds came onto the investors' platform. This green value also applies to NEU rewards and token claims as they are also funds coming onto the platform. Conversely, funds leaving the platform are represented in regular black font.

There are quite a few states so let's talk more about it if you want anything cleared up.

Cheers!

Louisswryan:
Hello @desfero ,

This part is probably the trickiest so maybe we can conference about it via Slack.

It is only tricky because of the number of states and potential layouts. A solution that I discussed with Trevor, is to use 

Modal Content guide:
https://zpl.io/2Egn4ll

rudolfix:
Green vs. Black

all transactions which have amount and have direction `in` should be green. for `out` black.

**important elements transaction types**
as we do not have designs for transaction types, let's start with following
- for all transactions we show timestamp and link to etherscan
- transactions status is always Complete, but you can repurpose it for transactions that have updates (see below)
- from and to addresses sometimes have known interfaces. if we know interface we should show it's name instead (I will help you with translations)


**eto investment**
- investent basics: in company name, you can link to their profile, asset token amount and symbol and ICON OF THE TOKEN (ask Louis where to put it)
- investment currency and amount (always show euro equivalent)
- NEU reward
- flag if icbm funds were used (see below)
- a clear indication if tokens were claimed or refunded, if not we should show clearly that this is still pending (it may make sense to show eto_status here and some call to action if user can do it but didn't do it yet)
- if claimed/refunded we can show exact date if this happen
[...]

PR #3227, opened 2019-07-24
#2890 Eto investment and nEur redeem tx history modals

desfero:
This pr add ETO investment and nEur redeem modals to tx history list.


app/modules/tx-history/sagas.ts
```diff
@@ -47,12 +48,36 @@ export function* mapAnalyticsApiTransactionResponse(
         throw new Error("Invalid asset token metadata");
       }
 
+      const neuReward = transaction.extraData.neumarkReward!.toString();
```
drakkein: is it possible for `neumarkReward` to be undefined in this case? Should we assert it and throw an error if it's empty?

drakkein: Discussion settled on Slack

PR #3243, opened 2019-07-29
Allow to overwrite feature flag in runtime through local storage

desfero:
This is just to test whether the implementation is stable enough. 
If tests are green will add some info to the README


Moejoe90:
Temporary closing

PR #3272, opened 2019-08-05
Do not report compile errors during unit tests development flow

desfero:
This pr will fix our unit tests development flow as right now any unused variable will throw compile-time error and those block unit tests. Still `yarn tsc:app` or `yarn:lint` will catch any unused variables.


tsconfig.json
```diff
@@ -1,17 +1,14 @@
 {
   "compilerOptions": {
     "allowSyntheticDefaultImports": true,
-    "alwaysStrict": true,
```
desfero: `alwaysStrict`, `notImplicitAny` and `noImplicitThis` are included by default in `strict` (line `16`)

PR #3278, opened 2019-08-07
Fix NEUR Redeem history bug

desfero:


glupiawymowka:
it's hard to tell what was the bug exactly, but one thing I've noticed is that the fee in the summary is rounded differently than the one in the redeem flow itself: 

1. Redeeming
image

2. Redeemed
image

PR #3289, opened 2019-08-08
Issuer Flow Meeting Fixes

Moejoe90:


app/components/eto/registration/pages/CompanyInformation.tsx
```diff
@@ -63,19 +63,19 @@ const EtoRegistrationCompanyInformationComponent = (
 
       <FormTextArea
         label={<FormattedMessage id="eto.form.company-information.company-description" />}
-        placeholder="Describe your company 250 Characters"
+        placeholder="Please describe your company, this description will be displayed on the listing page (max. 250 characters)."
```
desfero: (max. 250 characters) where `charactersLimit` is set to `750`, is that correct?

app/components/eto/shared/ETOState.tsx
```diff
@@ -106,12 +104,8 @@ const getState = (
   isIssuer: boolean,
 ): EETOStateOnChain | EEtoState | EEtoSubState => {
   const state = isOnChain(eto) ? eto.contract.timedState : eto.state;
-
   if (eto.subState) {
-    return isIssuer &&
-      eto.isMarketingDataVisibleInPreview !== EEtoMarketingDataVisibleInPreview.VISIBLE
-      ? state
-      : eto.subState;
+    return isIssuer ? state : eto.subState;
```
desfero: is that correct? As it was a business requirement AFAIK

app/components/eto/registration/pages/LegalInformation.tsx
```diff
@@ -80,6 +80,7 @@ const EtoRegistrationLegalInformationComponent = ({ savingData }: IProps) => (
       <FormTextArea
         label={<FormattedMessage id="eto.form.legal-information.company-legal-description" />}
         name="companyLegalDescription"
+        placeholder="Please enter the legal purpose of the entity stated in your incorporation documents."
```
desfero: btw, do we need to move into translations all changes in `placeholder`?
And also i general IMO it's bad to put a description into placeholders as it's hard to work with this normally as after to start typings description disappears and on small devices, it's cut off. 
Check how bootstrap shows help text https://reactstrap.github.io/components/form/, we should probably do the same here (but i prefer to keep help text just below `label`)

app/lib/api/eto/EtoApi.interfaces.unsafe.ts
```diff
@@ -366,26 +367,34 @@ export const EtoInvestmentTermsType = YupTS.object({
...
   existingCompanyShares: YupTS.number().enhance(v => v.min(MIN_EXISTING_COMPANY_SHARES)),
   authorizedCapitalShares: YupTS.number().optional(),
-  newSharesToIssue: YupTS.number().enhance(v =>
-    v.when("minimumNewSharesToIssue", (value: number) =>
-      v.min(
-        value,
-        getMessageTranslation(
-          createMessage(ValidationMessage.VALIDATION_MAX_NEW_SHARES_LESS_THAN_MINIMUM),
+  newSharesToIssue: YupTS.number()
+    .enhance(v => v.required())
```
desfero: hm, it's required by default AFAIK?

app/lib/api/eto/EtoApi.interfaces.unsafe.ts
```diff
@@ -366,26 +367,34 @@ export const EtoInvestmentTermsType = YupTS.object({
...
       ),
     ),
-  ),
   minimumNewSharesToIssue: YupTS.number().enhance(v => v.min(MIN_NEW_SHARES_TO_ISSUE)),
   newSharesToIssueInWhitelist: YupTS.number()
     .optional()
     .enhance(v => v.min(NEW_SHARES_TO_ISSUE_IN_WHITELIST)),
-  whitelistDiscountFraction: YupTS.number().optional(),
-  publicDiscountFraction: YupTS.number().optional(),
+  whitelistDiscountFraction: YupTS.number()
```
desfero: please use for any kind of custom fields own schema helpers so we can reuse and have consistent behavior, please see `percentage` from `schemaHelpers.ts`

intl/locales/en-en.json
```diff
@@ -131,16 +131,16 @@
   "documents.legal-documents": "Legal documents for your ETO",
   "documents.legal-documents.description": "In this section, you can manage all your legal agreements and documents for your ETO. In the first section, we generate for you agreements based on the date you enter into the ETO forms. You can download them as word files, review, adjust and upload signed versions in the second section below. The second section navigates you through the order of documents that need to be uploaded. ",
   "documents.please-fill-the-eto-forms-in-order-to-generate-templates": "Please fill the ETO forms in order to generate templates",
-  "download-agreement-widget.download-and-sign": "download and sign",
+  "download-agreement-widget.download-and-sign": "UPLOAD SIGNED ISHA",
```
desfero: please, let's use normal case everywhere and style text appropriately when needed aka (`text-transform: touppercase`).

intl/locales/en-en.json
```diff
@@ -131,16 +131,16 @@
...
   "download-agreement-widget.sign-on-ethereum": "Sign Investment Agreement on Ethereum",
   "download-agreement-widget.sign-on-ethereum-text": " Now when you uploaded your signed Investment Agreement, it's time to announce and confirm your signature on Ethereum. Only then Nominee will be notified and able to verify and provide his/her own signature.",
   "download-agreement-widget.sign-on-ethereum-text-modal": "This is the final step! Announced agreements can only be confirmed by the nominee. Click below to sign the agreement on Ethereum.",
-  "download-agreement-widget.signing-text": "Your ETO is almost complete. Upload the signed Investment Agreement to your legal documents and then you will be guided to confirm all signatures on Ethereum.",
+  "download-agreement-widget.signing-text": "We’ve updated the investment summary for you. Please download it from the document section and update your Investment and Shareholder Agreement. Make sure to use the correct data presented in the investment summary. To upload your signed Investment and Shareholders Agreement go to Documents section. By uploading your signed ISHA you will be able to withdraw collected funds.",
```
desfero: this seems a little bit duplicated:
`Please download it from the document section and update your Investment and Shareholder Agreement.` and then 
`To upload your signed Investment and Shareholders Agreement go to Documents section.` but maybe it's correct
Also, we should link links here so a user can just click `Documents` and go to this section

intl/locales/en-en.json
```diff
@@ -987,12 +988,12 @@
...
   "settings.book-building-widget.reactivate-book-building": "Re-activate Whitelisting",
   "settings.book-building-widget.start-book-building": "Enable subscription for your whitelist",
   "settings.book-building-widget.stop-book-building": "Disable whitelisting",
-  "settings.choose-pre-eto-date.book-building-will-stop": "The whitelisting phase will stop automatically {minOffsetPeriod, plural, one {# day} other {# days}} before your chosen presale date",
+  "settings.choose-pre-eto-date.book-building-will-stop": "Your ETO smart contracts have been deployed. Please set the start date and time (UTC time) of your presale now. Public sale will automatically start after your presale ends. Unless you disable the subscription for the whitelist manually it will automatically close {minOffsetPeriod, plural, one {# day} other {# days}} days before your chosen presale date.",
```
desfero: `{minOffsetPeriod, plural, one {# day} other {# days}} **days** ` days is not neede here, `plural` filter will insert it automatically (either `day` or `days`)

intl/locales/en-en.json
```diff
@@ -1057,12 +1058,12 @@
...
   "settings.upload-investment-button.title": "UPLOAD INVESTMENT MEMORANDUM",
-  "settings.upload-investment-memorandum-please-upload-prospectus": "Please go to the documents section and upload your signed Investment Memorandum. This will enable you to start your ETO fundraising on the platform.",
-  "settings.upload-investment-memorandum.title": "Upload signed Investment Memorandum",
-  "settings.upload-prospectus-please-upload-prospectus": "Please go to the documents section and upload your approved Prospectus. After that will need to upload the approved Pamphlet to enable your ETO fundraising on the platform.",
-  "settings.upload-prospectus.title": "Upload approved Prospectus",
-  "settings.upload-term-sheet-please-upload-term-sheet": "Please go to the documents section, download the generated Term Sheet, sign it and upload it to the platform. This will enable you to publish your ETO listing page and start your fundraising campaign.",
-  "settings.upload-term-sheet.button-label": "Upload Term Sheet",
+  "settings.upload-investment-memorandum-please-upload-prospectus": "Please go to the documents section and upload your signed Investment Memorandum.",
```
desfero: it would be also cool to make `documents` a link to documents same with `settings.upload-prospectus-please-upload-prospectus` and `settings.upload-term-sheet-please-upload-term-sheet`

app/components/eto/registration/pages/LegalInformation.tsx
Moejoe90: I agree with you but this was a per bizdevs request

app/components/eto/shared/ETOState.tsx
Moejoe90: This was amended later on as per bizdevs request

app/lib/api/eto/EtoApi.interfaces.unsafe.ts
Moejoe90: I thought the same until i tested it out. 
I think there is a bug in `enhance` for sure it doesn't consider the field required without this

app/lib/api/eto/EtoApi.interfaces.unsafe.ts
Moejoe90: hmmmm this only changes them from `optional` to required I'm not sure i understand your idea exactly

intl/locales/en-en.json
Moejoe90: This was a direct Bizdev copy change request

intl/locales/en-en.json
Moejoe90: indeed you're right will change accordingly

intl/locales/en-en.json
Moejoe90: Indeed i agree but i don't want to get out of the scope. which was fixing as much as possible in the issuer documents

app/components/eto/registration/pages/CompanyInformation.tsx
Moejoe90: let me check this. Good catch

Moejoe90: done

intl/locales/en-en.json
Moejoe90: to be done in cleanup

intl/locales/en-en.json
Moejoe90: done

app/lib/api/eto/EtoApi.interfaces.unsafe.ts
Moejoe90: ok added to clean up 
https://github.com/Neufund/platform-frontend/issues/3328

Issue #3237, opened 2019-07-29
[Nominee task 1 of 3] Agree and Sign ETO Agreements


andywalton:
This ticket is part of the Nominee Dashboard epic #3203

**TL;DR**

We're building a new dashboard to allow our new user type Nominee to register/login, connect with an Issuer and thereafter to perform tasks required of them before, during and after an Issuers fundraising campaign.

**Tasks for ticket**
At this stage a nominee has a successfully linked themselves to an Issuer.

The Nominee will at this stage have nothing to do until they are told to complete their first Nominee task: accepting and signing the ETO Agreements

There are two agreements that needs to be signed in following order
* Token Holder Agreement (THA). 
* Reservation and Acquisition Agreement (RAAA)

Steps are identical for both. You execute RAAA signing when you see that THA is already signed (see technical details below)

The steps are as follows:

- [ ] We need to tell them they should do this (a card on the dashboard with a call to action)
- [ ] This will trigger a modal that will allow them to "Agree" - download and view this THA after
- [ ] They will then need to Agree via clicking a checkbox (which will enable the modal call to action allowing them Sign
- [ ] Signing will move the modal to the next step - ie. to Sign via entering their Light Wallet password.
- [ ] This will results in an on-chain transaction and we should showcase that this is happening using our transaction status flow and a summary screen that shows it's either in progress, has an error or is still pending.
- [ ] if THA is signed but RAAA is not signed you repeat steps above for RAAA

**Designs**
[See designs in Zeplin](https://app.zeplin.io/project/5d2b7e0bd1281e565d9d08e4)

Scroll down to section: 5) Nominee task #1 - Sign THA
RAAA is identical, you just need to download different template (as pdf) and execute transaction agains different contract.

**Acceptance**
Is as per the steps above - we notify the Nominee that they should sign the THA, the can view and download this THA, agree that they are happy with it and sign for this using our transaction signer. We notify them of the success of this.

**Notes:**
We aren't handling the case where they might not agree to the THA, this is for a later phase and should be handled via support or contacting an account manager. For now, they need to do it. :)

Also, if the THA changes, it will ask the Nominee to redo this step but for the new THA.

rudolfix:
here's tech information for THA signing and later steps. you can copy relevant part to tickets once they are available.

@andywalton THA situation is different - Nominee is an owner of THA. he can sign and update THA as s/he wants and then it's up to issuer to accept the final terms. This happens when setting the date.

[...]

PR #3317, opened 2019-08-13
#3237 Nominee agreements

drakkein:


packages/web/app/components/nominee-dashboard/NomineeDashboard.tsx
```diff
@@ -58,13 +68,28 @@ export const NomineeDashboard = compose<IDashboardProps, {}>(
...
         isBankAccountVerified,
+        nomineeTHAStatus,
+        nomineeRAAAStatus,
       ),
     }),
   ),
+  branch<IStateProps>(
+    props =>
+      props.isBankAccountVerified &&
+      props.nomineeRAAAStatus === undefined &&
```
AlexeyTuboltsev: they shouldn't  be undefined, see comment below.
Why do you need a loading state at all? If those are NOT_DONE, the dashboard should simply show the first task that's returned from `getNomineeTaskStep`. It can switch to the another one when nomineeTasks are loaded, so there might be some flickering. This is not good, but we need a general solution to it, maybe add some logic to `isLoading` prop. We can solve this problem later. But no undefineds pls 😁

drakkein: I'm not going to object :P

packages/web/app/lib/web3/Web3Manager/Web3Manager.ts
```diff
@@ -19,7 +19,7 @@ import { Web3Adapter } from "../Web3Adapter";
 import { Web3FactoryType } from "../Web3Batch/Web3Batch";
 
 const DEFAULT_UPPER_GAS_LIMIT = 2000000;
-const DEFAULT_LOWER_GAS_LIMIT = 21000;
+export const DEFAULT_LOWER_GAS_LIMIT = 21000;
```
rudolfix: please note that @Moejoe90 defined same value in his ETH withdraw branch.

PR #3321, opened 2019-08-14
Fix confirmation link validator in unsubscribe flow

drakkein:
This PR updated `isValidLink` regex to match latest specification provided by @banciur 

Not testable via codechecks


packages/web/app/modules/marketing-emails/utils.ts
```diff
@@ -1,6 +1,11 @@
 // Its possible for link to be undefined if user removed link prop from URL
-export const isValidLink = (link: string | undefined) =>
-  link &&
-  /^https?:\/\/(neufund\.(org|net|io)|localhost:\d+)\/api\/newsletter\/subscriptions\/\w+\/topics\/\w+$/.test(
-    link,
-  );
+export const isValidLink = (link: string | undefined) => {
```
desfero: can we make this function to accept only `string`? And filter any kind of `undefined` where it's called?

packages/web/app/modules/marketing-emails/sagas.ts
```diff
@@ -13,14 +13,15 @@ function* unsubscribeEmail({
   logger,
 }: TGlobalDependencies): Iterator<any> {
   const unsubscribeLink = yield select(selectUnsubscriptionLinkFromQueryString);
+  const decodedUnsubscribeLink = decodeURI(unsubscribeLink);
```
desfero: this should be moved to `getUnsubscriptionLinkFromQueryString`

desfero: Also, i see there is an option `decode` that you can pass to `query-string.parse` (https://www.npmjs.com/package/query-string#decode)

Issue #3257, opened 2019-08-01
[Issuer] - Investment terms can have negative min max amounts


andywalton:
**Description**
Editing the Investment Terms, we managed to create a situation where we have a negative (-) min and max investments amount.

Ooops. 

**Environment**
Chrome, Mac (latest versions/OS)

**Steps To Reproduce**
Login as using a fixture that allows you to act as an Issuer that can enter Investment Terms. Play around with the amounts as per the screengrabs below. 

**Actual behaviour**
Displays negative min max investment amounts. 

**Expected behavior**
Expect the values to validate correctly and warn you that this doesn't look right. You shouldn't be able to raise a negative amount.

**Screenshots**
![Screenshot 2019-08-01 17.27.34.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/5609184f-6925-4032-93b9-ce1cd27f9851)

Moejoe90:
I know that we copied a python implementation from the backend. 
The backend already has unit tests for that and AFAIK we also have unit test. 

Please check if we have all cases covered and all Unit tests are written from our side. 
- [  ] Check if we have all unit tests from the python implementation provided by @rudolfix 
- [  ] If yes all tests should pass 
- [  ] if yes do all forms have correct restrictions `max` `min`

glupiawymowka:
tested on stage

PR #3353, opened 2019-08-21
#3257 Fix possible negative values in Invest terms fllow

drakkein:


packages/web/app/components/eto/registration/pages/InvestmentTerms.tsx
```diff
@@ -104,6 +104,12 @@ const InvestmentCalculator: React.FunctionComponent<ICalculatorProps> = ({
 }) => (
   <FormikConsumer>
     {({ values }) => {
+      const fixedSlotsMaximumDiscountFraction = parseStringToFloat()(
```
Moejoe90: ok this seems to me as a point of break. 

What happens if the parsed string is `NaN`. 
We need to make sure this case is taken care off maybe the whole app should break and we report it to sentry

packages/web/app/components/eto/registration/pages/InvestmentTerms.tsx
```diff
@@ -112,11 +118,18 @@ const InvestmentCalculator: React.FunctionComponent<ICalculatorProps> = ({
         preMoneyValuationEur: parseStringToFloat()(values.preMoneyValuationEur),
         newSharesToIssueInFixedSlots: parseStringToFloat()(values.newSharesToIssueInFixedSlots),
         newSharesToIssueInWhitelist: parseStringToFloat()(values.newSharesToIssueInWhitelist),
-        fixedSlotsMaximumDiscountFraction: parseStringToFloat()(
-          values.fixedSlotsMaximumDiscountFraction,
-        ),
-        whitelistDiscountFraction: parseStringToFloat()(values.whitelistDiscountFraction),
-        publicDiscountFraction: parseStringToFloat()(values.publicDiscountFraction),
+        fixedSlotsMaximumDiscountFraction:
+          typeof fixedSlotsMaximumDiscountFraction === "number"
```
Moejoe90: why do we have this assertion if we are always parsing up?

drakkein: due of it's implementation `parseStringToFloat` can return `undefined`. I'm switching to `convertPercentageToFraction` utility function so this assertion would not be required

packages/web/app/components/eto/registration/pages/InvestmentTerms.tsx
drakkein: Right now `convertPercentageToFraction` gonna throw Invariant error

packages/web/app/components/eto/registration/pages/InvestmentTerms.tsx
drakkein: Also I'm gonna pass 0 to calculations while it's not a number. Is this ok?

Moejoe90: it would be good to log this case in sentry to know something went wrong. we don't need to break the interfect. 
What are the cases where `parseStringToFloat` can return undefined?

drakkein: We settled this on Slack

packages/web/app/components/eto/registration/pages/InvestmentTerms.tsx
```diff
@@ -112,11 +118,15 @@ const InvestmentCalculator: React.FunctionComponent<ICalculatorProps> = ({
         preMoneyValuationEur: parseStringToFloat()(values.preMoneyValuationEur),
         newSharesToIssueInFixedSlots: parseStringToFloat()(values.newSharesToIssueInFixedSlots),
         newSharesToIssueInWhitelist: parseStringToFloat()(values.newSharesToIssueInWhitelist),
-        fixedSlotsMaximumDiscountFraction: parseStringToFloat()(
-          values.fixedSlotsMaximumDiscountFraction,
-        ),
-        whitelistDiscountFraction: parseStringToFloat()(values.whitelistDiscountFraction),
-        publicDiscountFraction: parseStringToFloat()(values.publicDiscountFraction),
+        fixedSlotsMaximumDiscountFraction: fixedSlotsMaximumDiscountFraction
```
AlexeyTuboltsev: this is fine but I suggest to write it in a slightly better way, using those functions as they were intended to be used 😁 :
```
const conversionSpec = {
        newSharesToIssue: parseStringToFloat(),
        minimumNewSharesToIssue: parseStringToFloat(),
        equityTokensPerShare: parseStringToFloat(),
        existingCompanyShares: parseStringToFloat(),
        preMoneyValuationEur: parseStringToFloat(),
        newSharesToIssueInFixedSlots: parseStringToFloat(),
        newSharesToIssueInWhitelist: parseStringToFloat(),
        fixedSlotsMaximumDiscountFraction: [parseStringToFloat(), convertPercentageToFraction()],
        whitelistDiscountFraction: [parseStringToFloat(),convertPercentageToFraction()],
        publicDiscountFraction: [parseStringToFloat(), convertPercentageToFraction()],
      };

const calculatorValues = convert(values,conversionSpec );
```

`convert()` takes data and a conversion spec where each property that we convert will be run through the conversion fn or through the array of fns one by one.

packages/web/app/lib/api/eto/EtoUtils.ts
```diff
@@ -60,6 +60,14 @@ const getMaxInvestmentAmountWithDiscount = (
     return 0;
   }
 
+  if (
```
AlexeyTuboltsev: 👍

packages/web/app/components/eto/registration/pages/InvestmentTerms.tsx
```diff
@@ -446,4 +456,4 @@ const fromFormState = {
   authorizedCapitalShares: parseStringToInteger(),
 };
```
AlexeyTuboltsev: btw I know it's not your code, but if there's no hurry I'd move this bunch of small calculator fns to a one big calculation util function that returns an object with all those values, using the `getCapPercent` etc under the hood. Just to separate the calculations from the visual stuff.

packages/web/app/components/eto/registration/pages/InvestmentTerms.tsx
drakkein: I'd like to refactor this in separate PR, it's not kinda related to this bug, and we need it merged.

drakkein: https://github.com/Neufund/platform-frontend/issues/3361

PR #3354, opened 2019-08-22
Fix firefox date field input shrinking

desfero:
Seems there is some inconsistency between Chrome and Firefox flex implementation. This change should force input to shrink to the container size.
See https://bugzilla.mozilla.org/show_bug.cgi?id=1242914 for more information.

**Testing notes**
Please check whether it's fixed in **Firefox** and also recheck **Chrome** whether it's still okay after a fix


desfero:
This is how KYC looks right now on `.io`.
image

Issue #3359, opened 2019-08-23
Fixes to the Rich text editor images/image caption


glupiawymowka:
**1. Company description input, stage** 

- [ ]  Overlapping toolbars
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/4b0a06cf-3c1f-4af8-84bf-894a11c31fdd)
- [x] image added at the end of the text is disappearing just after it was added 
- [ ]  trying to add more than 1 character at once to the image caption results with replacing the image with regular text.
- [ ]  Deleting a character from text or caption deletes the whole image
- [ ] no character limit is shown, so the text is simply cut after being saved

**All inputs**
- [ ] no character limit for image caption
- [ ] image caption character limit is shared with the general limit for the input

desfero:
@glupiawymowka there is a pr that will fix `image added at the end of the text is disappearing just after it was added`. 
`Overlapping toolbars` bug is reported on the CKEditor bug tracker so let's wait for a fix there and then just update

pstroinski:
Please check if the fix already done on the Rich text editor

PR #3366, opened 2019-08-26
#3359 Fix rich-text editor content trimming

desfero:
**Testing notes**
This pr fixes:
- [ ] image added at the end of the text is disappearing just after it was added


packages/web/app/components/shared/forms/fields/FormTextArea.tsx
```diff
@@ -21,6 +22,9 @@ interface IFieldGroup {
   name: string;
   disabled?: boolean;
   placeholder?: string;
+  /**
+   * @deprecated Use `Yup` max validation to keep schema related validation in one place
+   */
```
AlexeyTuboltsev: 👍

packages/web/app/components/shared/forms/fields/FormTextArea.tsx
```diff
@@ -38,6 +42,13 @@ const RichTextArea: React.FunctionComponent<TFieldGroupProps & TFormikConnect> =
   charactersLimit,
   formik,
 }) => {
+  if (process.env.NODE_ENV === "development") {
```
Moejoe90: why not move this to a util function for example `forceDeprecation` that has `if (process.env.NODE_ENV === "development")` and an invariant check? 

I can see it being used more so better have one clear way to force

PR #3393, opened 2019-09-01
#3237 Nominee agreements E2E tests 

drakkein:
This PR adds E2E tests to Nominee agreements flow


packages/web/app/test-e2e/nominee-flow/NomineeFlow.spec.e2e.ts
```diff
@@ -0,0 +1,20 @@
+import { assertNomineeDashboard, goToDashboard, loginFixtureAccount } from "../utils/index";
+import { assertNoTasks, signRAAA, signTHA } from "./NomineeFlowUtils";
+
+describe("Nominee flow", () => {
```
desfero: can we rename to file to be `NomineeAgreementsAcceptanceFlow` as we already have the file named `NomineeFlow` and it's hard to understand at first what's the difference

packages/web/app/test-e2e/nominee-flow/NomineeFlow.spec.e2e.ts
```diff
@@ -0,0 +1,20 @@
...
+  it("should accept agreements", () => {
+    loginFixtureAccount("NOMINEE_SETUP_NO_ST", {
+      kyc: "business",
+      signTosAgreement: true,
+      clearPendingTransactions: true,
+    }).then(() => {
+      goToDashboard(false);
```
desfero: It's a function that we should use only for investor (btw, can you rename it to be `goToInvestorDashboard`? For nominee we should use `goToNomineeDashboard` and then we can remove `assertNomineeDashboard` below

packages/web/app/test-e2e/nominee-flow/NomineeFlowUtils.ts
```diff
@@ -10,3 +12,44 @@ export const linkEtoToNominee = (address: string) => {
...
+  successModalSelector: string,
+) => () => {
+  cy.get(containerSelector).should("exist");
+  cy.get(tid("eto-nominee-sign-agreement-action")).click();
+
+  cy.get(modalSelector).should("exist");
+
+  fillForm({
+    acceptAgreement: {
+      type: "checkbox",
```
desfero: please use `type: "toggle"` for boolean checkbox as `{ false: true }` look akward

PR #3397, opened 2019-09-02
[wip] onfido prototype

huesforalice:
For working on this, please check out the "onfido implementation" branch of the backend and run that. We will later figure out how to merge all of this to master.

The general flow for onfido is the following:
* Collect some upfront data from the user. In our case only the first and last name. There may be more to collect later.
* Let the user upload some documents to onfido. Onfido has a javascript sdk for that, you just need to aquire a webtoken from the backend to initialise it.
* Request that onfido runs a check on the uploaded data.
* If all is good, the kyc request will go to accepted

In particular you need to do the following:
* Upload first and last name to PUT /kyc/individual/data
* Get an onfido webtoken from PUT /kyc/individual/request/onfido/upload-request
* Start the onfido sdk with the aquired webtoken. There will be an overlay over our site, which the user uses to upload his passport and a foto of himself (this may also change, but this is not important for now)
* From the success handler of the onfido sdk, start an onfido check-request with PUT /kyc/individual/request/onfido/check-request.
* If the check passes, GET /kyc/individual/request will switch the status to accepted

Additional states and notes:
* After you do the first onfido upload-request, GET /kyc/individual/request will switch its status to "outsourced" and the instant_id_provider attribute to "onfido". Similarly if you start and idnow request, there will be an instant_id_provider attribute that is set to "idnow".
* You can only start an upload request, if the  GET /kyc/individual/request is in "draft" or in "outsourced" with the instant_id_provider set to "onfido".
* You can only start an onfido check request if GET /kyc/individual/request is in in "outsourced" with the instant_id_provider set to "onfido". Also you can't start a check request, if there is one currently ongoing.
* To see wether there is a check request ongoing, do GET /kyc/individual/request/onfido/check-request. If this returns 404, it means there has never been a check. If it returns and object with the status attribute set to "complete", the last check has been completed. There will also be a result attribute. If the result attribute is set to "clear", it means the last check has passed. In this case the GET /kyc/individual/request will also have the status switched to "approved". If the result attribute is set to "consider", this means the last onfido check has failed.
* If the last onfido check has failed, you can let the user upload documents again by just creating a new upload-request. After completion of this upload request, you can again start a check-request.


Moejoe90:
Will temp close

Issue #3230, opened 2019-07-26
Metamask signer is shown after visiting `login/browser` when already logged in


desfero:
**Description**
Metamask signer is shown after visiting manually `login/browser` when already logged in.

**Environment**
-

**Steps To Reproduce**
1. Log in
2. Visit `/login/browser` url 
3. Metamask message signer appears

**Actual behaviour**
Metamask signer appears 

**Expected behavior**
We should not show any kind of signing request. 

**Comments, context, logs**

**Screenshots**
![Jul-26-2019 13-00-09](https://user-images.githubusercontent.com/7031578/61947392-5a3e2e80-afa5-11e9-9d5a-da9c666c49d1.gif)

glupiawymowka:
another similar case I had is: 
1. There was an open metamask + investor session in one browser window
2. Forgetting about it, I tried to open a new metamask + issuer account by using the secret link.
 The action was done in a new browser window, with just created metamask wallet 
3. The dashboard with my new wallet address opened, but the account was an investor

left: old browser window,      right: new browser window
![image](https://user-images.githubusercontent.com/18315589/62205058-d7471a80-b38e-11e9-91bd-0bf38a503cff.png)

glupiawymowka:
tested on .io

glupiawymowka:
tested & released

PR #3409, opened 2019-09-04
#3230 Fix only public route to wait for auth state

drakkein:
This PR fixes issue with MetaMask signer showing up when you navigate to `login/browser` page, whether user is logged in or not and overwriting session with selected MetaMask wallet without signing. 

For the next step I will add some automated tests using https://github.com/decentraland/dappeteer


packages/web/app/modules/auth/reducer.ts
```diff
@@ -31,6 +31,10 @@ export const authReducer: AppReducer<IAuthState> = (
         ...state,
         currentAgreementHash: action.payload.currentAgreementHash,
       };
+    case actions.web3.personalWalletDisconnected.getType():
```
drakkein: Quick explanation why this was added. Since `logoutUser()` from `auth/sagas.ts` was only clearing localStorage and leaving state unchanged, while recovery saga performs whole login -> logout cycle. Because of that `OnlyPublicRoute` was getting stucked, since initial state was set with empty auth and during saga processing it got changed, so it made `OnlyPublicRoute` to not invoke onEnter redirection and later on change displayed component to loading.

packages/web/app/components/App.tsx
```diff
@@ -94,6 +92,8 @@ const App = compose<React.ComponentClass>(
     }),
     options: { pure: false }, // we need this because of:https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/blocked-updates.md
   }),
+  branch<IStateProps>(state => !!state.error, renderComponent(CriticalError)),
+  branch<IStateProps>(state => !state.done, renderComponent(LoadingComponent)),
```
desfero: hm, i dont' know why do we use `LoadingComponent` from `MyPortfolioWidget`. Also we can get rid of `done` here completly as it's the reverse of `inProgress`. Also we can now remove `LoadingIndicator` from the `render` method above

packages/web/app/components/shared/routing/OnlyPublicRoute.tsx
```diff
@@ -25,6 +28,7 @@ export const OnlyPublicRoute = compose<IComponentProps, RouteProps>(
   appConnect<IStateProps, {}, RouteProps>({
     stateToProps: s => ({
       isAuthorized: selectIsAuthorized(s.auth),
+      appInitDone: selectIsInitDone(s.init),
```
desfero: not used

packages/web/app/components/shared/routing/OnlyPublicRoute.tsx
```diff
@@ -25,6 +28,7 @@ export const OnlyPublicRoute = compose<IComponentProps, RouteProps>(
   appConnect<IStateProps, {}, RouteProps>({
```
desfero: in general i think out guards should also implement `componentDidUpdate` and in case `isAuthorized` was set to true redirect automatically.

Or we even can rewrite `OnlyPublicRouteComponent` to
```
   isAuthenticated === true
      ? <Component {...rest} />
      : <Redirect to={{
          pathname: '/dashboard',
        }} />
```

drakkein:
Not testable via codechecks, requires to manually visit `login/browser` route

PR #3414, opened 2019-09-06
User inactivity logout

Moejoe90:


packages/web/app/di/setupBindings.ts
```diff
@@ -229,6 +234,13 @@ export function setupBindings(config: IConfig): Container {
     .inSingletonScope();
 
   container.bind(symbols.intlWrapper).toConstantValue(new IntlWrapper());
+  container.bind(symbols.broadcastChannel).toConstantValue(
+    config.isMochaRunning
```
Moejoe90: there was an issue while running tests with default imports.

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
```diff
@@ -32,8 +44,83 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
...
+        });
+      });
+    });
+
+    it("should not logout automatically when a user has activity", () => {
+      createAndLoginNewUser({
+        type: "investor",
+        kyc: "business",
+      }).then(() => {
+        goToDashboard();
```
desfero: We can simplify code to
```
        goToDashboard();
        const now = Date.now();
        cy.clock(now);

        cy.tick(AUTH_INACTIVITY_THRESHOLD / 2);

        cy.get(tid("authorized-layout-wallet-button")).trigger("mousemove");

        cy.tick(AUTH_INACTIVITY_THRESHOLD / 2);

        assertDashboard();
```

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
```diff
@@ -32,8 +44,83 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
...
+        cy.clock(now).then(clock => {
+          clock.tick(AUTH_INACTIVITY_THRESHOLD / 2);
+          assertDashboard();
+          clock.tick(AUTH_INACTIVITY_THRESHOLD / 2);
+          assertLogin();
+        });
+      });
+    });
+
+    it("should not logout automatically when a user has activity", () => {
```
desfero: There is something wrong with this test as it's a false positive (try to rereun a couple of times and you may see that you land on a login page but test is marked as green)

desfero: From what i can see `jwt` is timing out here as it's the same amount of time

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
```diff
@@ -32,8 +44,83 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
...
+    clock.tick(AUTH_INACTIVITY_THRESHOLD / 2);
+    return channel.postMessage({
+      status: "user-active",
+    });
+  });
+
 describe("auto-logout/auto-login", () => {
-  it("should logout automatically when a user logs out from another tab", () => {
+  describe("Automatic Actions", () => {
+    it("should logout automatically when a user logs out from another tab", () => {
```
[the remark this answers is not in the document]
desfero: Also, it's a flaky test as it's testing `jwt` timeout. Just log in `watchTimeoutActions` what action was resolved

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
```diff
@@ -32,8 +44,83 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
...
+        });
+      });
+    });
+    it("should not logout automatically when a user has activity in another tab", () => {
+      createAndLoginNewUser({
+        type: "investor",
+        kyc: "business",
+      }).then(() => {
+        goToDashboard();
+        const now = Date.now();
```
desfero: I'm wondering whether it's worth to refactor like this:
```
        cy.clock(now);

        cy.tick(AUTH_INACTIVITY_THRESHOLD / 2);

        doSomeActionInAnotherTab();

        cy.tick(AUTH_INACTIVITY_THRESHOLD / 2);

        doSomeActionInAnotherTab();

        assertDashboard();
```
`doSomeActionInAnotherTab` will just broadcast event without any `ticks`. What do you think?

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
```diff
@@ -32,8 +44,83 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
...
+                assertDashboard();
+              });
+            });
+        });
+      });
+    });
+  });
+});
+describe("User Driven Actions", () => {
+  it("should logout automatically", () => {
```
desfero: should be `should logout automatically when a user logs out from another tab`

packages/web/app/modules/auth/user/sagas.ts
```diff
@@ -12,15 +14,47 @@ import {
...
+  userActivityChannel,
+}: TGlobalDependencies): Iterator<any> {
+  while (true) {
+    const { logout, active } = yield race({
+      logout: safeDelay(AUTH_INACTIVITY_THRESHOLD),
+      active: take(actions.auth.userActive.getType()),
+      refresh: take(actions.auth.refreshTimer.getType()),
+    });
+    // safeDelay sometimes returns 0 or 1
+    if (typeof logout !== "undefined") {
```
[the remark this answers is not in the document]
Moejoe90: both can work i figured asserting by type is better. I have no hard opinions here its the same

packages/web/app/modules/auth/user/sagas.ts
```diff
@@ -12,15 +14,47 @@ import {
...
+      logger.info("User is inactive logging out");
+      yield put(actions.auth.userActivityTimeout());
+    }
+    if (active) {
+      yield userActivityChannel.postMessage({
+        status: "user-active",
```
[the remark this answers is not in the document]
Moejoe90: This is a good question. 
I will keep like this for now. I have a plan to start using `status` more and more. I also didn't want to add more code in this PR.

If you think the code should be changed in anyway that indicates this please suggest and lets do it. but i would keep status data type.

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
```diff
@@ -32,8 +44,83 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
...
+            cy.get(tid("authorized-layout-wallet-button")).trigger("mousedown");
+            return cy.get(tid("authorized-layout-wallet-button")).trigger("mouseleave");
+          }).then(() => {
+            clock.tick(AUTH_INACTIVITY_THRESHOLD / 2);
+            assertDashboard();
+          });
+        });
+      });
+    });
+    it("should not logout automatically when a user has activity in another tab", () => {
```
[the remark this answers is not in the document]
Moejoe90: done

packages/web/app/modules/auth/watcher/sagas.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+/**
+ * Saga that watches for timeout events and logs out the
+ * user when a timeout actions is dispatched */
+export function* watchTimeoutActions(): any {
+  while (true) {
+    yield race({
+      userTimeout: take(actions.auth.userActivityTimeout.getType()),
```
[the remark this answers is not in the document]
Moejoe90: done

packages/web/app/modules/auth/user/sagas.ts
```diff
@@ -12,15 +14,47 @@ import {
...
-import { loadPreviousWallet } from "../../web3/sagas";
 import { EWalletSubType, EWalletType } from "../../web3/types";
+import { AUTH_INACTIVITY_THRESHOLD } from "../constants";
 import { createJwt } from "../jwt/sagas";
 import { selectUserType } from "../selectors";
+import { ELogoutReason } from "../types";
+import { logoutUser } from "./external/sagas";
+
+/**
+ * Refresh jwt before timing out.
```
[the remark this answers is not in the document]
Moejoe90: oops missed that one, good catch

packages/web/app/modules/auth/watcher/sagas.ts
```diff
@@ -0,0 +1,89 @@
...
+        break;
+    }
+    yield delay(REDIRECT_CHANNEL_WATCH_DELAY);
+  }
+}
+
+/**
+ * Saga that watches for timeout events and logs out the
+ * user when a timeout actions is dispatched */
+export function* watchTimeoutActions(): any {
```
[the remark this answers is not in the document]
Moejoe90: done

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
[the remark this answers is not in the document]
Moejoe90: Hmmm intresting because in theory we should always refresh JWT, 
Maybe i need to `tick` exactly to the point where jwt gets refreshed. 

How can we do that?

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
```diff
@@ -32,8 +44,83 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
...
+declare type Message = {
+  status: string;
+};
+const channel: BroadcastChannel<Message> = new BroadcastChannel(
+  symbols.userActivityChannel.toString(),
+);
+
+const pushTimeThenPostMessage = (clock: any) =>
+  cyPromise(() => {
+    cy.wait(1000);
```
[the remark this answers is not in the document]
Moejoe90: The issue comes from the time difference between posting a message and ticking. 
It gets apperantly worse the faster the machine is. I had to introduce it after testing on CI. 

The test will pass locally btw but not on CI

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
Moejoe90: I tried with 
```
        cy.get(tid("authorized-layout-wallet-button")).trigger("mousemove");
```
It doesn't work. 

Also there is an issue with `cy.tick` with tiral and error i realized i had batch all the commands i want and then do a `then` 

I assume this was not tested correct?

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
Moejoe90: done

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
Moejoe90: As mentioned above this doesn't work i had to batch the commands and wrap then around a then.

Now after looking at your comments though i realized that indeed the JWT timeout might be interfearing let me test with that and see

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
Moejoe90: I solved it in a quick way for now and setup the timeout time to 5 mins for inactivity this should be ok as it will not effect other tests. 

The downside means that if the backend decided to change for whatever reason to 5 mins then the tests will fail. At the same time i don't  want to inroduce complex logic of ticking until the moment where jwt refreshes and await a refresh as this will make the tests slower.

packages/web/app/test-e2e/general-app/autoLoginLogoutActions.spec.e2e.io.ts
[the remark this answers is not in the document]
desfero: can you add some comment then why we need `wait`? I imagine in month or two someones will try to remove `wait` because tests are green locally and then will spend some time understanding what's wrong

packages/web/app/lib/dependencies/broadcast-channel/types.ts
```diff
@@ -0,0 +1,8 @@
+export declare type UserActivityChannelMessage = {
+  status: EUserActivityMessage;
+};
+
+export enum EUserActivityMessage {
+  USER_ACTIVE,
```
desfero: Can we use `string` enums here? I think it's then easier to understand what's the value really is in the final bundle as `0` doesn't say anything

packages/web/app/di/setupBindings.ts
```diff
@@ -229,6 +235,15 @@ export function setupBindings(config: IConfig): Container {
     .inSingletonScope();
 
   container.bind(symbols.intlWrapper).toConstantValue(new IntlWrapper());
+  container
+    .bind(symbols.userActivityChannel)
+    .toConstantValue(
+      config.isMochaRunning
```
desfero: Let's get rid of `config.isMochaRunnin` as rebind in `integrationTestUtils` to 
```
  container.rebind(symbols.userActivityChannel).toConstantValue(new BroadcastChannelMock());
```
as with the current aprroach `BroadcastChannelMock` will get included into production bundle. Also to get rid of `.default` mocha errors let's move it from `toConstantValue` to `toDynamicValues` so `createNewBroadcastChannel` is not even called.
```
container
    .bind(symbols.userActivityChannel)
    .toDynamicValue(() =>
      createNewBroadcastChannel<UserActivityChannelMessage>(symbols.userActivityChannel.toString()),
    )
    .inSingletonScope();
```
I hope this should work :D

PR #3417, opened 2019-09-06
[Nominee] Add ability to change active ETO

### LINKED ISSUE #3415: [Nominee dashboard] - all the completed steps loading after refresh
**Description**
Even though the user has completed all the currently available steps, they show up, one by one, before "you're all up to date" screen appears. 

This gives a strong impression that something is broken

**Environment**
all
**Steps To Reproduce**
1. Login with a fulfilled nominee fixture (NOMINEE_NEUMINI works well)
2. Refresh the dashboard
https://files.slack.com/files-pri/T1X3Q3L2F-FMQ93T5JP/sep-05-2019_14-38-39.gif

**Actual behavior**
after refresh, the modal shows all the previously completed states for a brief moment

**Expected behavior**
only the applicable step should be shown 

**Comments, context, logs**

**Screenshots**

desfero:
**What has be changed**
- Adds an ability to force another ETO to be active (by default for nominee we select first one)
- Refactors nominee dashboard to not show intermediate steps while we load all required information to calculate next task (fixes #3415)
- Minor refactoring (forcing all query selectors to receive whole state as a param) and dead code elimination 
 - Fixing race condition that leads to `Missing Nominee ETO` (https://sentry.neusrv.net/sentry/development/issues/6940/)

**Testing notes**
To change nominee eto you need to use `eto` query param and pass `previewCode`. (If it's successful you will receive proper notification, if preview code is invalid we just select first valid eto from nominee etos list)
Example: `https://platform.neufund.io/dashboard?eto=f92ce30d-c54c-48f1-a9d7-e6f85cb88269`
Also would be good to do whole regression for nominee flow.


packages/web/app/components/nominee-dashboard/NomineeDashboard.tsx
```diff
@@ -48,40 +34,18 @@ export const DashboardTitle: React.FunctionComponent<IDashboardTitleProps> = ({
...
+      nomineeTaskStep: selectNomineeTaskStep(state),
     }),
   }),
-  onEnterAction<IStateProps>({
-    actionCreator: (dispatch, { verificationIsComplete }) => {
-      if (verificationIsComplete) {
-        dispatch(actions.nomineeFlow.loadNomineeTaskData());
-      }
+  onEnterAction({
+    actionCreator: dispatch => {
```
Moejoe90: so you can't access this component if `verificationIsComplete` is not true or was it not needed before?

packages/web/app/modules/nominee-flow/reducer.ts
```diff
@@ -129,6 +73,23 @@ export const nomineeFlowReducer: AppReducer<INomineeFlowState> = (
...
+      return {
+        ...state,
+        activeNomineeEtoPreviewCode: action.payload.previewCode,
+      };
+    case actions.nomineeFlow.setNomineeEtos.getType():
+      return {
+        ...state,
+        nomineeEtos: {
+          ...state.nomineeEtos,
+          ...action.payload.etos,
```
Moejoe90: set nominee etos should take all the etos correct and update the latest.
does it make sense to still keep the `nomineeEtos` and spread the new ones?

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -198,9 +206,85 @@ function* loadAgreementStatus(
...
+      typeof selectActiveEtoPreviewCodeFromQueryString
+    > = yield select(selectActiveEtoPreviewCodeFromQueryString);
+
+    if (isEmpty(etos)) {
+      if (previewCode !== undefined) {
+        yield put(actions.nomineeFlow.setActiveNomineeEto(undefined));
+      }
+    } else {
+      if (
+        forcedActiveEtoPreviewCode !== undefined &&
```
Moejoe90: maybe a constant with a name that explaines the logic here

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -198,9 +206,85 @@ function* loadAgreementStatus(
...
+        forcedActiveEtoPreviewCode !== previewCode &&
+        etos[forcedActiveEtoPreviewCode] !== undefined
+      ) {
+        yield put(actions.nomineeFlow.setActiveNomineeEto(forcedActiveEtoPreviewCode));
+
+        notificationCenter.info(
+          createMessage(EEtoNomineeActiveEtoNotifications.ACTIVE_ETO_SET_SUCCESS),
+        );
+      } else if (previewCode === undefined || etos[previewCode] === undefined) {
```
Moejoe90: so the fallback ETO is always the first?

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -198,9 +206,85 @@ function* loadAgreementStatus(
...
+        const firstEto = nonNullable(Object.values(etos)[0]);
+        yield put(actions.nomineeFlow.setActiveNomineeEto(firstEto.previewCode));
+      }
+    }
+  } catch (e) {
+    logger.fatal("Could not set active eto", e);
```
Moejoe90: when do you use `fatal` and when do you use `error`?

packages/web/app/modules/nominee-flow/utils.ts
```diff
@@ -84,3 +93,38 @@ export const nomineeIsEligibleToSignAgreement = (nomineeEto: TEtoWithCompanyAndC
   isOnChain(nomineeEto) &&
   nomineeEto.contract.timedState === EETOStateOnChain.Setup &&
   nomineeEto.contract.startOfStates[EETOStateOnChain.Whitelist] === undefined;
+
+export const getNomineeTaskStep = (
```
Moejoe90: good step moving this here.
but it would be really goot if it was more a saga that can access store and get all required informtaion

packages/web/app/test-e2e/nominee-flow/NomineeChangeActiveEto.spec.e2e.io.ts
```diff
@@ -0,0 +1,27 @@
...
+    });
+
+    const issuerProspectusApprovedId = accountFixtureAddress("ISSUER_PROSPECTUS_APPROVED");
+    goToNomineeDashboard();
+    cy.get(tid(`eto-overview-${issuerProspectusApprovedId}`)).should("exist");
+
+    const issuerListedId = accountFixtureAddress("ISSUER_LISTED");
+    getEto(issuerListedId).then(eto => {
+      goToNomineeDashboard(eto.previewCode);
+      cy.get(tid(`eto-overview-${issuerListedId}`)).should("exist");
```
Moejoe90: hmm does this test actually change the active eto?

packages/web/app/components/nominee-dashboard/NomineeDashboard.tsx
desfero: In general, it worked before because nominee `loadNomineeEto` was dispatched somewhere below in the tree, now nominee etos call is done inside `loadNomineeTaskData` as we need it right there to calculate additional states. I hope it explains the reason, if no, catch me on slack.

packages/web/app/modules/nominee-flow/reducer.ts
desfero: for nominee etos that's probably a good point, fixed

packages/web/app/modules/nominee-flow/sagas.ts
desfero: yes, it's how it is right now

packages/web/app/modules/nominee-flow/sagas.ts
desfero: done

packages/web/app/modules/nominee-flow/sagas.ts
desfero: `error` when it's still possible to use an app (like when transactions history list failed to load, you still can use other parts of the platform)
`fatal` when it's not possible to use the app as something critical is missing, in this case without active eto it's not possible to use nominee`

packages/web/app/modules/nominee-flow/utils.ts
desfero: completely agree, but I don't want to introduce a lot of changes in single pr, the same problem is with issuer flow logic where we pass whole logic as parameters. IMO it should be all moved to selectors and then get all we need from sub selectors

packages/web/app/test-e2e/nominee-flow/NomineeChangeActiveEto.spec.e2e.io.ts
desfero: if tests are green then it does, `goToNomineeDashboard` call `cy.visit` with `eto` query param, or what's your point here?

Issue #3404, opened 2019-09-03
redeem nEUR fee not calculated on confirmation and incorrectly rounded earlier


rudolfix:
**Description**
1. Fee deduced by the bank is incorrectly computed and does not add to the total
2. On confirmation dialog, fee is always 0

**Environment**
all

**Steps To Reproduce**
Try to redeem any amount and go to confirmation dialog

**Actual behaviour**
as per description

**Expected behavior**
1. fee is calculated as follows. 
there are two sources of purchase amount
- value entered by the user
- a full balance of nEUR wallet if transfer all is used. full balance is an integer like `1000000000000000000` for 1 EUR
a. first step is to quantize amount to ISO20022 (SWIFT) format: scale is 5 (5 decimal places), rounding is HALF_UP
b. in case of wei value you first convert it to decimal 1.0 nEUR and then quantize
c. on that number you compute the final total: amount - amount*0.005
d. now you quantize it to scale 2 with ROUND_DOWN (that is the value that bank frick accepts)
e. this is a final value you display to the user as 'Total Redeemed`
f now you compute display fee. you subtract `Total Redeemed` from `Amount to Redeem` display value. That ensures that total always adds with display fee!
2. you display exactly same data on confirmation dialog
3. extract methods to ISO2022 rounding and bank rounding to some utility place and write some unit tests

**Screenshots**
![image](https://user-images.githubusercontent.com/17202864/64193042-47f4c180-ce7c-11e9-9f60-bbdaefc19ce2.png)
![image](https://user-images.githubusercontent.com/17202864/64193112-6490f980-ce7c-11e9-8d58-9cbeda322871.png)

glupiawymowka:
point 2 is for sure fixed

PR #3427, opened 2019-09-09
#3404 Fix nEUR redeem fee calculation

drakkein:


packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+import { MONEY_DECIMALS } from "../../../../config/constants";
+import { getPossibleMaxUlps } from "../../../../utils/Number.utils";
+import { convertFromUlps, isEmptyValue, isValidNumber } from "../../../shared/formatters/utils";
+
+export const getValidAmount = (amount: string, maxAmount: string): BigNumber => {
+  if (!isValidNumber(amount) || (isEmptyValue(amount) && 0)) {
+    return new BigNumber(0);
+  }
+
+  const amountUlps = getPossibleMaxUlps(maxAmount, amount, 2);
```
Moejoe90: This should be a constant

packages/web/app/utils/Number.utils.ts
```diff
@@ -43,3 +50,63 @@ function normalizeValue(options: TNormalizeOptions, value: number): number {
...
+  const valueBn = value instanceof BigNumber ? value : new BigNumber(value);
+
+  if (valueBn.isZero()) {
+    return "0";
+  }
+
+  return valueBn.toFixed(DISPLAY_AMOUNT_SCALE, BigNumber.ROUND_DOWN);
+};
+
+export const getPossibleMaxUlps = (
```
Moejoe90: please move this to redeem utils

packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  return convertFromUlps(new BigNumber(amountUlps), MONEY_DECIMALS);
```
Moejoe90: This is actually not returing full precision because 
Also since this is for display purpose 
hvae you used `toFixedPrecision`?

packages/web/app/modules/tx/transactions/redeem/saga.ts
```diff
@@ -68,11 +61,11 @@ export function* startNEuroRedeemGenerator(_: TGlobalDependencies): any {
     throw new Error("During redeem process user should have bank account");
   }
 
-  const bankFee: string = yield select(selectBankFeeUlps);
+  const bankFee: string = yield select(selectRedeemFeeUlps);
 
   const additionalDetails = {
     bankFee,
-    amount: txDataFromUser.value,
+    amount: convertFromUlps(new BigNumber(redeemAmountUlps), MONEY_DECIMALS).toString(),
```
Moejoe90: This should be already in state so we can use a selector

packages/web/app/modules/tx/transactions/redeem/saga.ts
```diff
@@ -45,16 +45,9 @@ export function* startNEuroRedeemGenerator(_: TGlobalDependencies): any {
...
-    .div(new BigNumber(10).pow(MONEY_DECIMALS))
-    .toFixed(2, BigNumber.ROUND_DOWN);
-
   // Whole precision number should be passed when there is whole balance redeemed
   // also when user provided value has been used, then it have to be converted to Q18 via convertToBigInt
-  const redeemAmountUlps =
-    compareBigNumbers(selectedAmount, nEURBalance) === 0
-      ? nEURBalanceUlps
-      : convertToBigInt(selectedAmount);
+  const redeemAmountUlps = getPossibleMaxUlps(nEURBalanceUlps, selectedAmount, 2);
```
Moejoe90: by the time we reach this point the conversion should be correct we don't need to do the conversion again.
The generator should be called with the correct params.

Also it would be good to add checks to this.

packages/web/app/utils/Number.utils.ts
```diff
@@ -43,3 +50,63 @@ function normalizeValue(options: TNormalizeOptions, value: number): number {
 }
 
 export const normalize = curry(normalizeValue);
+
+export const iso2002Quantize = (value: number | BigNumber | string): string => {
```
Moejoe90: It would be better to have this in a separate file actually. 
Since they are more specific would be easier to find

packages/web/app/utils/Number.utils.spec.ts
```diff
@@ -26,3 +35,65 @@ describe("isZero", () => {
     expect(isZero("364458900000000000")).to.be.false;
   });
 });
+
+describe("iso2002Quantize", () => {
```
Moejoe90: Thanks for the tests :+1:

packages/web/app/modules/tx/transactions/redeem/saga.ts
```diff
@@ -43,23 +40,15 @@ export function* startNEuroRedeemGenerator(_: TGlobalDependencies): any {
   const txDataFromUser = action.payload.txDraftData;
   const selectedAmount = txDataFromUser.value;
 
-  const nEURBalanceUlps = yield select(selectLiquidEuroTokenBalance);
```
Moejoe90: good that we got rid of all this into one place

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,61 @@
...
+import { multiplyBigNumbers, subtractBigNumbers } from "./BigNumberUtils";
+
+export const iso2002Quantize = (value: number | BigNumber | string): string => {
+  const valueBn = value instanceof BigNumber ? value : new BigNumber(value);
+
+  if (valueBn.isZero()) {
+    return "0";
+  }
+
+  return toFixedPrecision({
```
Moejoe90: How many places share this exact settings? 
if more than one place use one util method

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const bankQuantize = (value: number | BigNumber | string): string => {
+  const valueBn = value instanceof BigNumber ? value : new BigNumber(value);
+
+  if (valueBn.isZero()) {
+    return "0";
+  }
+
+  return toFixedPrecision({
```
Moejoe90: Also this

packages/web/app/modules/bank-transfer-flow/sagas.ts
```diff
@@ -107,6 +118,46 @@ export function* completeBankTransfer({ apiKycService, logger }: TGlobalDependen
...
+
+  const bankFeeFractionDec = toFixedPrecision({
+    value: bankFeeFractionUlps,
+    decimalPlaces: ISO2022_AMOUNT_SCALE,
+  });
+
+  // by default use BANKING_AMOUNT_SCALE as decimal precision
+  const providedAmountUlps = getPossibleMaxUlps(walletBalanceUlps, amount);
+  // use all decimal places without rounding
+  const providedAmountDec = toFixedPrecision({
```
Moejoe90: Then why do you use `toFixedPrecision` in the first place?

packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { BANKING_AMOUNT_SCALE } from "../../../../config/constants";
+import { compareBigNumbers } from "../../../../utils/BigNumberUtils";
+import { convertToBigInt } from "../../../../utils/Number.utils";
+import { ERoundingMode, toFixedPrecision } from "../../../shared/formatters/utils";
+
+export const getPossibleMaxUlps = (
+  maxUlpsValue: string,
+  providedDecimal: number | string,
+  decimalPrecision = BANKING_AMOUNT_SCALE,
```
Moejoe90: why do we give the option to choose percision? 
In our case its always `BANKING_AMOUNT_SCALE` correct just force it to be like that this is only `redeem` related

packages/web/app/modules/bank-transfer-flow/sagas.ts
```diff
@@ -107,6 +118,46 @@ export function* completeBankTransfer({ apiKycService, logger }: TGlobalDependen
   }
 }
 
+export function* calculateRedeemData(amount: string): Iterator<any> {
+  const walletBalanceUlps: string = yield select(selectLiquidEuroTokenBalance);
+  const bankFeeFractionUlps: string = yield select(selectRedeemFeeUlps);
+
+  const bankFeeFractionDec = toFixedPrecision({
```
Moejoe90: So there is no settings for the number here from fixed precision really? rounding mode for example output mode etc?

packages/web/app/components/modals/tx-sender/redeem/BankTransferRedeemInit.tsx
```diff
@@ -205,15 +209,34 @@ const BankTransferRedeemLayout: React.FunctionComponent<IProps> = ({
                 </Heading>
               </Tooltip>
               <span className="text-warning">
-                {"-"} {isValid && <CalculatedFee bankFee={bankFee} amount={values.amount} />}
+                {"- "}
+                {isValid && calculatedData && (
```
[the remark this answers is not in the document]
drakkein: Actually no, it wont be shown until form is not valid, so we can populate state with 0 without any problem

packages/web/app/modules/bank-transfer-flow/reducer.ts
```diff
@@ -25,6 +26,7 @@ export interface IBankTransferState {
     minEuroUlps: string;
     bankFeeUlps: string;
   };
+  calculatedRedeemData?: ICalculatedRedeemData;
```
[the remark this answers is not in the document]
drakkein: there is none

packages/web/app/utils/BankArithmetics.ts
drakkein: Both this an the one below are used only in this file (note that they are slightly different)

packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+) => {
+  const nEURBalance = toFixedPrecision({
```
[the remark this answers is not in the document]
drakkein: yeah, now I see it :D

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -19,43 +24,67 @@ export const iso2002Quantize = (value: number | BigNumber | string): string => {
...
+    if (new BigNumber(feeFraction).comparedTo(convertToBigInt(1)) === 1) {
+      throw new Error("FeeFraction must be fraction number");
+    }
   }
 
-  const iso2002Amount = iso2002Quantize(amount);
-  const feeBn = multiplyBigNumbers([iso2002Amount, feeFraction]);
+  const fee =
+    inputFormat === ENumberInputFormat.ULPS
+      ? toFixedPrecision({
```
Moejoe90: Please be careful with the use of `toFixedPrecision`

For example here
```
toFixedPrecision({
          value: feeFraction,
          decimalPlaces: ISO2022_AMOUNT_SCALE,
        })
```
You don't have a rounding mode set up and here
https://github.com/Neufund/platform-frontend/pull/3427/files#diff-68d4b3455ffb9a2610463755294eced5R23
You do

Define clearly which one is right and use only one util method to get the correct percision

packages/web/app/modules/bank-transfer-flow/sagas.ts
Moejoe90: This is still not resolved

packages/web/app/utils/BankArithmetics.ts
Moejoe90: How is this slightly different than
https://github.com/Neufund/platform-frontend/pull/3427/files#diff-68d4b3455ffb9a2610463755294eced5R61
There the rounding mode is missing and i doubt it should be a different rounding mode

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export const calculateBankFee = (
+  amount: number | BigNumber | string,
+  feeFraction: number | BigNumber | string,
+  inputFormat: ENumberInputFormat,
+): string => {
+  const amountVal =
+    inputFormat === ENumberInputFormat.ULPS
+      ? toFixedPrecision({
```
Moejoe90: Why doesn't this use 
`toFixedBankingPrecision` and where is the RoundingMode

packages/web/app/components/modals/tx-sender/redeem/BankTransferRedeemInit.tsx
```diff
@@ -205,15 +209,34 @@ const BankTransferRedeemLayout: React.FunctionComponent<IProps> = ({
...
               <span className="text-warning">
-                {"-"} {isValid && <CalculatedFee bankFee={bankFee} amount={values.amount} />}
+                {"- "}
+                {isValid && (
+                  <Money
+                    data-test-id="bank-transfer.redeem.fee"
+                    value={calculatedData.bankFee}
+                    inputFormat={ENumberInputFormat.FLOAT}
+                    valueType={ECurrency.EUR}
+                    outputFormat={ENumberOutputFormat.FULL}
```
Moejoe90: Also here please make sure of the `outputFormat`
```
                    outputFormat={ENumberOutputFormat.FULL}
```
Should it be full or `nonZeroDecimals`

packages/web/app/modules/bank-transfer-flow/sagas.ts
```diff
@@ -107,6 +119,49 @@ export function* completeBankTransfer({ apiKycService, logger }: TGlobalDependen
...
+  // use all decimal places without rounding
+  const providedAmountDec = toFixedPrecision({
+    value: providedAmountUlps,
+    decimalPlaces: undefined,
+  });
+
+  const bankFee = calculateBankFee(
+    providedAmountUlps,
+    bankFeeFractionUlps,
+    ENumberInputFormat.ULPS,
```
Moejoe90: do you use `calculateBankFee` ever as `ENumberInputFormat.FLOAT`? 
I don't think so so the tests should handle if the numbers were `ULPS` not Float

packages/web/app/utils/BankArithmetics.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+        convertToBigInt("14873.97"),
+        convertToBigInt("0.004"),
+        ENumberInputFormat.ULPS,
+      ),
+    ).to.be.eq("14814.47");
+  });
+});
+
+describe("calculateBankFee", () => {
+  it("should correctly calculate value with fee subtracted", () => {
```
Moejoe90: I think `calculateBankFee` shouldn't be tested as `ENumberInputFormat.FLOAT` since the use case is only for ULPS
is this correct? will `calculateBankFee` ever get used as a FLOAT instead of ULPS

drakkein: Probably never, I will remove this param and force ULPS

drakkein:
Conflicts fixed

packages/web/app/modules/bank-transfer-flow/sagas.ts
```diff
@@ -107,6 +121,38 @@ export function* completeBankTransfer({ apiKycService, logger }: TGlobalDependen
...
 }
 
+export function* calculateRedeemData(amount: string): Iterator<any> {
+  const walletBalanceUlps: string = yield select(selectLiquidEuroTokenBalance);
+  const bankFeeFractionUlps: string = yield select(selectRedeemFeeUlps);
+
+  // by default use BANKING_AMOUNT_SCALE as decimal precision
+  const providedAmountUlps = getPossibleMaxUlps(walletBalanceUlps, amount);
+  // use all decimal places without rounding
+  const providedAmountDec = toFixedBankingPrecision(providedAmountUlps, ENumberInputFormat.ULPS);
```
rudolfix: this is not how it was described in the ticket:

there are two sources of purchase amount
- value entered by the user
- a full balance of nEUR wallet if transfer all is used. full balance is an integer like `1000000000000000000` for 1 EUR
a. first step is to quantize amount to ISO20022 (SWIFT) format: scale is 5 (5 decimal places), rounding is HALF_UP
b. in case of wei value you first convert it to decimal 1.0 nEUR (`from_wei`) and then `iso_quantize`
c. on that number you compute the final total: amount - amount*0.005
d. now you quantize it to scale 2 with ROUND_DOWN (that is the value that bank frick accepts) (`bank_quantize`)
e. this is a final value you display to the user as `Total Redeemed`
f. now you compute **display fee**. you subtract `Total Redeemed` from `Amount to Redeem` display value. That ensures that total always adds with display fee!

what's wrong below is that you operate on bank scale. you should operate on SWIFT scale and convert to bank scale for display purposes. again:

redeemed amount ulps -> from_wei -> iso_quantize -> calculate bank fee that is iso quantized -> total_redeemed: iso quantized amount - iso quantized fee -> display total redeemed which is bank quantize -> display fee which is Amount to Redeem display value - bank quantized total redeemend

in transaction you use redeemed amount ulps

please fix it

packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+import { compareBigNumbers } from "../../../../utils/BigNumberUtils";
+import { convertToBigInt } from "../../../../utils/Number.utils";
+import {
+  ENumberInputFormat,
+  ERoundingMode,
+  toFixedPrecision,
+} from "../../../shared/formatters/utils";
+
+export const toFixedBankingPrecision = (
+  value: number | BigNumber | string,
```
rudolfix: number cannot be accepted to any function that expects ULPS. please!

packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+import { BANKING_AMOUNT_SCALE } from "../../../../config/constants";
+import { compareBigNumbers } from "../../../../utils/BigNumberUtils";
+import { convertToBigInt } from "../../../../utils/Number.utils";
+import {
+  ENumberInputFormat,
+  ERoundingMode,
+  toFixedPrecision,
+} from "../../../shared/formatters/utils";
+
+export const toFixedBankingPrecision = (
```
rudolfix: function name is misleading. it expects Ulps as input. why not separate into two function
`from_wei` that converts to decimal with WEI SCALE and then bank quantize?

packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  value: number | BigNumber | string,
+  inputFormat: ENumberInputFormat = ENumberInputFormat.ULPS,
+) =>
+  toFixedPrecision({
```
rudolfix: why formatting function is used for arithmetics?

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,74 @@
...
+  value: number | BigNumber | string,
+  inputFormat: ENumberInputFormat,
+): string => {
+  const valueBn = value instanceof BigNumber ? value : new BigNumber(value);
+
+  if (valueBn.isZero()) {
+    return "0";
+  }
+
+  return toFixedBankingPrecision(valueBn, inputFormat);
```
rudolfix: and here you are passing decimal, not ulps. I think I may not understand stuff

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export const subtractBankFee = (
+  amount: number | BigNumber | string,
+  feeFraction: number | BigNumber | string,
```
rudolfix: fee fraction is ULPS so please never allow number here

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,74 @@
...
+  const iso2002Amount = iso2002Quantize(amount, ENumberInputFormat.ULPS);
+  const feeBn = multiplyBigNumbers([iso2002Amount, feeDec]);
+  const totalBn = subtractBigNumbers([iso2002Amount, feeBn]);
+
+  // Calculation result is always FLOAT
+  return bankQuantize(totalBn, ENumberInputFormat.FLOAT);
+};
+
+export const calculateBankFee = (
+  amount: number | BigNumber | string,
```
rudolfix: amountUlps and never number

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,74 @@
...
+  feeFraction: number | BigNumber | string,
```
rudolfix: number not allowed

packages/web/app/utils/BankArithmetics.ts
```diff
@@ -0,0 +1,74 @@
...
+): string => {
+  const amountDec = toFixedBankingPrecision(amount, ENumberInputFormat.ULPS);
```
rudolfix: it does not work like this, wrong quantization order

packages/web/app/components/modals/tx-sender/redeem/utils.ts
```diff
@@ -0,0 +1,31 @@
+import BigNumber from "bignumber.js";
+
+import { BANKING_AMOUNT_SCALE } from "../../../../config/constants";
+import { compareBigNumbers } from "../../../../utils/BigNumberUtils";
+import { convertToBigInt } from "../../../../utils/Number.utils";
+import {
+  ENumberInputFormat,
+  ERoundingMode,
+  toFixedPrecision,
```
rudolfix: just reviewed FixedPrecision code, this should not be used in any arithmetic operations. please provide basic set of methods to convert from wei to decimal and back +  iso and bank quantizers. you do not need anything else.

also operations on numbers should not be allowed.

packages/web/app/components/modals/tx-sender/redeem/utils.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { expect } from "chai";
+
+import { convertToBigInt } from "../../../../utils/Number.utils";
```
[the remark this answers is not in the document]
rudolfix: also you use `toFixedPrecision` for the same purpose!

packages/web/app/modules/bank-transfer-flow/sagas.ts
rudolfix: also rename `getPossibleMaxUlps` to `getRedeemValueUlps` because this is what it does! it does not return any max value

Moejoe90:
As per discussion with @rudolfix We will close this PR for now

Issue #2425, opened 2019-02-06
Unsuccessful login and registration on iOS 12 / Safari


Desperado:
**Description**
It is not possible to register or sign-in with iOS 12 and 11

**Environment**
Use physical device or browserstack, both Safari and Chrome 

**Expected behaviour**
Normal ability to login and signup (at least on iOS12) 

**Screenshots**
![bs_realios_Mobile_iPhone XS-12.0 (2).jpg](https://images.zenhubusercontent.com/5c3ddac32c893f1dad8d7a9a/eb7205ed-022b-42fc-bd78-456ad571fd5a)

ios 12 
![image](https://user-images.githubusercontent.com/18315589/58802234-f6cefa80-860c-11e9-9039-3f7c568dfbb3.png)

ios 11 
![image](https://user-images.githubusercontent.com/18315589/58802379-53321a00-860d-11e9-84d9-88cf86a7ca51.png)

**Inspect**
Network: 

headers: 

Request URL:https://platform.neufund.io/api/user/user/me
Request Method:GET
Status Code:401
cf-ray:4e12203b3f6ba6c3-DUB
Content-Length:2517
Content-Type:application/json
Date:Mon, 03 Jun 2019 13:48:39 GMT
expect-ct:max-age=604800, report-uri=“https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct”
Server:cloudflare
X-XSS-Protection:1; mode=block
Provisional headers are shown
Accept:application/json, text/plain
[...]

PR #3431, opened 2019-09-10
#2425 Add custom X-NF-Authorization header to AuthHttpClient

drakkein:
It is possible to be checked by @glupiawymowka on codechecks. Just check the request headers to the paths that requires authorization

image


packages/web/app/lib/api/client/AuthHttpClient.ts
```diff
@@ -24,6 +24,7 @@ export abstract class AuthorizedHttpClient implements IHttpClient {
       headers: {
         ...config.headers,
         Authorization: `Bearer ${this.objectStorage.get()}`,
+        "X-NF-Authorization": `Bearer ${this.objectStorage.get()}`,
```
desfero: Can we add a comment why it's needed there as it's quite not obvious

Issue #3285, opened 2019-08-08
Cleanup for #3209 no.1


AlexeyTuboltsev:
**Original Ticket**
#3209 

**Link to code Review**
https://github.com/Neufund/platform-frontend/pull/3276/files

**Technical notes**
this ticket includes small code fixes, storybook tests and naming changes
the bigger issue with moving the logic to sagas will be handled in a separate ticket

**Testing notes**

Issue #3209, opened 2019-07-18
[Nominee setup 1] Link nominee to issuer


andywalton:
This ticket is part of the Nominee Dashboard epic #[3203](https://github.com/neufund/platform-frontend/issues/3203).

**TL;DR**

We're building a new dashboard to allow our new user type Nominee to register/login, connect with an Issuer and thereafter to perform tasks required of them before, during and after an Issuers fundraising campaign.

---

**Task**
At this stage a nominee has a verified Neufund account.

The next thing it should do is to link itself to an Issuer company that is fundraising on the platform so that it may perform the required tasks expected of a nominee.

The steps are as follows:
1. The nominee is told they should enter the Ethereum wallet address of the company they are wanting to link with. We should verify this wallet address is an Issuer user on the platform.
2. There will be a form that they can input this address into and submit it.
3. Submitting the form results in an email being sent via the platform to the relevant Issuer wallet address.
4. The nominee then is told this request is pending acceptance.
5. The Issuer will then be notified that there is a new nominee request pending (ie. via the email sent to them, via a dashboard notification card.
6. Clicking on this notification will open the ETO form called Token Holder Rights where they can see the details of this pending nominee request. They can then accept / reject this notification.
7. On acceptance, the issuer is asked to sign/confirm this action using our signer and the Nominee is notified via email and via their dashboard. This then changes the state and the nominee is now setup to be a nominee for the issuer.
8. On rejection, the issuer will be told they should share their wallet address with the company they're wanting to act as a nominee. They should also then tell this company to register for a nominee account on Neufund via `https://platform.neufund.org/nominees` (ie. the secret URL we have for the nominees). The nominee will also be notified that their nominee request was rejected by the company and that they should try another company.

---

**Designs**
See designs in [Zeplin](https://zpl.io/ameO4vv)

Nominee screens:
- 3a) A nominee requests to link with an issuer

Issuer screens:
- 3b) What the Issuer will see via their Dashboard (ISSUER VIEW)

---
**Tasks**

_Please edit this ticket with relevant tasks to best break up the work._

Note, this requires both work on the Nominee and Issuer dashboards.

---

**Acceptance**
- [x] A Nominee can request a connection with an Issuer company by submitting this company's Ethereum wallet address (this wallet address is validated to check that it's correct and exists on our platform as a verified Issuer.
 - [x] We handle the pending state of this request and tell the Nominee that it's pending. 

[...]

PR #3437, opened 2019-09-11
#3285 cleanup for #3209

AlexeyTuboltsev:
clean up all stuff not fixed in the #3209 except for moving logic to sagas
so mostly storybook tests


packages/web/app/modules/tx/transactions/sagas.ts
```diff
@@ -185,8 +185,6 @@ export function* etoSignInvestmentAgreementSaga(
       extraParam: action.payload,
     });
     logger.info("Signing investment agreement was successful");
-    // cleanup & refresh eto data
```
Moejoe90: :+1:

Issue #3382, opened 2019-08-28
[WHITELISTING] - fixes to the current flow 


glupiawymowka:
**Description**
The whitelisting needs a total re-design, but we could fix some problems before we get there. 

**Tasks**

1.  We're validating the min whitelist commitment, but not the max one 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/23007195-0014-4ed3-80bf-11bc3286c4d0)

status: 400
title: "Pledge with amount 50000000 is not within bounds of 100000 and 15920398.00995024876000 for eto id 0x2e2eb728c6974C93a564Fe615847573603ADDaDa"
type: "exceptions.PledgeNotWithinBoundsException"

ACCEPTANCE CRITERIA:
- [ ] the similar information as for the min amount shows up, instead of the unclear error bar 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/f6b43828-5e9b-45e7-8262-935d5227bf10)

2. See the screen above. We should not add .00 to full numbers, so in this case we just show 100 000
ACCEPTANCE CRITERIA:
- [ ] Full number is not represented with non existing decimals 

3. Widget size changes once the user starts typing 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/cca76853-d223-42da-9f33-79562a03b310)

4. There's a place to show commited amount in the incoming email, but nothing was added there. 
If this is too much work, please let @vizmab know so she deletes this part from the email 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/f657aa58-bc07-4937-aec9-ad95edd43bd9)

ACCEPTANCE CRITERIA: 
- [ ] `Amount commited` either show the valid amount, or is deleted completely 

5. User should not be able to enter dot `.` in the input field. `1000.` is possible to enter. only round numbers, guys, without decimals!
- [ ] only digits accepted in input

rudolfix:
Technical information

ad. 1 backend will provide `effective_max_ticket` value against which you should validate maximum value
ad 2. this applies only to `amount` not prices or rates, I really hope you've made this difference in your formatters. all amounts that are round (without decimal places) are displayed without decimal dot

andywalton:
Hello, so I have some initial ideas about improving the UX. This needs more ideation and design but in the meantime some of the validation - ie. min commitment size => min ticket size etc. could commence.

[...]

PR #3439, opened 2019-09-11
#3382 Fix whitelisting flow

drakkein:


packages/web/app/components/eto/overview/EtoOverviewStatus/CampaigningWidget/CampaigningActivatedInvestorApprovedWidgetLayout.tsx
```diff
@@ -92,22 +94,43 @@ const CampaigningActivatedInvestorApprovedWidgetLayout: React.FunctionComponent<
...
+                    setFieldTouched("amount", true);
+                  }}
+                  returnInvalidValues={true}
+                  showUnits={true}
+                  decimals={0}
+                />
+              </div>
+              <div className={cn(styles.value, styles.backNow)}>
+                <Button
+                  data-test-id="eto-bookbuilding-back-now"
```
Moejoe90: should be changed to commit instead of `back-now` since its not going back

packages/web/app/components/eto/overview/EtoOverviewStatus/EtoStatusManager.tsx
```diff
@@ -48,13 +48,13 @@ const EtoStatusComponentChooser: React.FunctionComponent<IStateProps & IExternal
 
   switch (timedState) {
     case EETOStateOnChain.Setup: {
-      if (isAuthorized) {
+      if (isAuthorized && eto.investmentCalculatedValues) {
```
Moejoe90: is there a case where `investmentCalculatedValues` cannot exist? is that what's coming from the swagger defenition?

packages/web/app/components/shared/MaskedNumberInput.tsx
```diff
@@ -41,12 +41,17 @@ interface IProps {
   reverseMetaInfo?: boolean;
   prefix?: TTranslatedString;
   suffix?: TTranslatedString;
+  size?: EInputSize;
+  decimals?: number;
```
Moejoe90: no need to add this `decimals` and all this logic since you already picked 
```
                  outputFormat={ENumberOutputFormat.INTEGER}
```

Or it doesn't work?

packages/web/app/lib/api/eto/EtoApi.interfaces.unsafe.ts
```diff
@@ -425,6 +440,7 @@ interface IAdditionalEtoType {
   product: TEtoProduct;
   nomineeDisplayName?: string;
   hasDividendRights?: boolean;
+  investmentCalculatedValues?: IInvestmentCalculatedValues;
```
Moejoe90: this is coming from the backend correct? is it really optional?

packages/web/webpack/loadAppEnv.js
```diff
@@ -2,6 +2,7 @@ const { join } = require("path");
 const { mapValues } = require("lodash");
 const dotenv = require("dotenv");
 const fs = require("fs");
+const url = require("url");
```
Moejoe90: what happened here?

packages/web/app/components/eto/overview/EtoOverviewStatus/EtoStatusManager.tsx
drakkein: Yeah, this is from swagger definition

packages/web/app/lib/api/eto/EtoApi.interfaces.unsafe.ts
drakkein: According to swagger definition it is optional

packages/web/webpack/loadAppEnv.js
drakkein: If I try to run `start:remote:serve-on-network` using node `8.12.0` it will throw an error due of missing `URL` definition, so we need to import `url` module. If you think that we should leave it this way I will revert.

packages/web/app/components/eto/overview/EtoOverviewStatus/EtoStatusManager.tsx
drakkein: But since @rudolfix added typings on master I will reuse it that form.

packages/web/webpack/loadAppEnv.js
Moejoe90: All good

packages/web/app/components/shared/MaskedNumberInput.tsx
```diff
@@ -91,7 +92,9 @@ export class MaskedNumberInput extends React.Component<IProps> {
   };
 
   changeValue = (value: string) => {
-    const validValue: string | null = parseInputToNumber(value);
+    const processedValue =
```
Moejoe90: Do we really need `processedValue`?

drakkein: If we want completely block user from writing `.` it's required. Otherwise user will be able to write for example `20.` (`20.1` will not be allowed tho).

packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  getMessageTranslation,
+  ValidationMessage,
+} from "../../components/translatedMessages/messages";
+import { createMessage } from "../../components/translatedMessages/utils";
+
+export const isPledgeAboveMinimum = (minPledge: number): TestOptions => ({
+  name: "minAmount",
+  message: getMessageTranslation(
+    createMessage(ValidationMessage.VALIDATION_MIN_PLEDGE, minPledge),
+  ) as any,
```
Moejoe90: Just make it as a `string`

packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  test: function(this: TestContext, value: string) {
+    return isValidNumber(value) && new BigNumber(value).comparedTo(minPledge) === 1;
+  },
+});
+
+export const isPledgeNotAboveMaximum = (maxPledge?: number): TestOptions => ({
+  name: "minAmount",
+  message: getMessageTranslation(
+    createMessage(ValidationMessage.VALIDATION_MAX_PLEDGE, maxPledge),
+  ) as any,
```
Moejoe90: also here just cast as a `string`

packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+  // tslint:disable-next-line
+  test: function(this: TestContext, value: string) {
```
Moejoe90: There is no need for 
```
// tslint:disable-next-line
```
Just do 
```
function(this: TestContext, value: string) : Boolean
```
You need to return a type

packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    return isValidNumber(value) && new BigNumber(value).comparedTo(maxPledge || Infinity) <= 0;
```
Moejoe90: just to make sure that what is returned is a boolean why not wrap 
```
isValidNumber(value) && new BigNumber(value).comparedTo(maxPledge || Infinity)
```
Around 
```
!!(isValidNumber(value) && new BigNumber(value).comparedTo(maxPledge || Infinity))
```

If you don't want to do then please prove that this works without with a unit test

packages/web/app/components/eto/overview/EtoOverviewStatus/CampaigningWidget/CampaigningActivatedInvestorApprovedWidgetLayout.tsx
```diff
@@ -87,27 +92,56 @@ const CampaigningActivatedInvestorApprovedWidgetLayout: React.FunctionComponent<
         </div>
       </div>
     ) : (
-      <Formik<{ amount: number | "" }>
+      <Formik<{ amount: string }>
         initialValues={{ amount: pledgedAmount }}
         onSubmit={({ amount }) => backNow(Number(amount))}
-        validationSchema={generateCampaigningValidation(minPledge, maxPledge)}
+        validate={values => {
```
Moejoe90: my bad that i didn't mention this to you but now if you put 
```
        validationSchema={generateCampaigningValidation(minPledge, maxPledge)}
```
Back 
instead of 
```
validate={values => {
  try {
            generateCampaigningValidation(minPledge, maxPledge).validateSync(values, {
              abortEarly: false,
            });
          } catch (errors) {
            return yupToFormErrors(errors);
          }
          return undefined;
        }}
```
It will work since we don't need to do cross form validations

packages/web/app/modules/investor-portfolio/selectors.ts
```diff
@@ -140,6 +140,13 @@ export const selectCalculatedEtoTicketSizesUlpsById = (state: IAppState, etoId:
 
   if (min && max) {
     if (eto && investorTicket) {
+      if (!eto.investmentCalculatedValues) {
+        return {
+          minTicketEurUlps: zero,
```
Moejoe90: what is zero where is it coming from?

packages/web/app/modules/tx/transactions/investment/sagas.ts
```diff
@@ -165,6 +165,10 @@ export function* investmentFlowGenerator(_: TGlobalDependencies): any {
   const etherPriceEur: string = yield select(selectEtherPriceEur);
   const isIcbm: boolean = yield select(selectIsICBMInvestment);
 
+  if (!eto.investmentCalculatedValues) {
+    throw new Error("ETO investment calculated values are empty");
```
Moejoe90: Since this is empty please also log it as an error to sentry because if you check the parent saga we don't log any information

packages/web/app/modules/bookbuilding-flow/utils.ts
drakkein: ahhh... I added this because of error `error TS2730: An arrow function cannot have a 'this' parameter.`  now and forgot to remove it before pushing.

packages/web/app/modules/bookbuilding-flow/utils.ts
drakkein: I like unit tests, so I will add them :D

packages/web/app/modules/investor-portfolio/selectors.ts
drakkein: zero is just `new BigNumber(0)` defined few lines above;

packages/web/app/modules/tx/transactions/investment/sagas.ts
drakkein: 👍

drakkein:
@glupiawymowka Texts are aligned now, spacing has been reduced to have more space for errors to show up, it's helping in case of short messages, but with longer widget still changes it's size.

Issue #3395, opened 2019-09-02
[Issuer] Add new step: Sign ETO legal agreements


desfero:
**Background**
After Investment memorandum was uploaded we should transition eto dashboard state to `YOUR NOMINEE IS CURRENTLY SIGNING ETO AGREEMENTS`.

State for this step: Your current ETO status is:  <green> Whitelisting/Campaigning
Step text: Your ETO smart contracts are deployed on-chain now. We have notified your Nominee to sign the legal agreements on-chain. Once all agreements you will be able to review all ETO terms, sign legal agreements from your side and set the date and time for the ETO.

**Designs**

**Acceptance criteria**
- [ ] Given I'm logged in as an Issuer When smart contracts were deployed then step should be changed to `YOUR NOMINEE IS CURRENTLY SIGNING ETO AGREEMENTS` with text as above

**Technical notes**

**Testing notes**

glupiawymowka:
tested on stage

PR #3444, opened 2019-09-12
#3395 Issuer steps 14, 15, 16. 

desfero:
**Testing notes**
This pr fixes a couple of steps for issuer dashboard.
- Adds a step `YOUR NOMINEE IS CURRENTLY SIGNING ETO AGREEMENTS` (https://docs.google.com/presentation/d/1v2gQFfd_gCI1XDgoDPCVFQpjEpqXbzYg-ehOHILHUlQ/edit#slide=id.g5f03e622ab_0_70)
- Adds a step `AGREE TO TERMS OF YOUR ETO AND SET THE DATE OF YOUR ETO` (https://docs.google.com/presentation/d/1v2gQFfd_gCI1XDgoDPCVFQpjEpqXbzYg-ehOHILHUlQ/edit#slide=id.g51aa7b02f8_0_181 and https://docs.google.com/presentation/d/1v2gQFfd_gCI1XDgoDPCVFQpjEpqXbzYg-ehOHILHUlQ/edit#slide=id.g5f03e622ab_0_47)
- Partial adds a step `YOUR ETO STARTS SOON` (only step name is implemented) (https://docs.google.com/presentation/d/1v2gQFfd_gCI1XDgoDPCVFQpjEpqXbzYg-ehOHILHUlQ/edit#slide=id.g51aa7b02f8_0_186)

Also fixes nominee dashboard bug where we call `nominee/etos` endpoint before checking whether the user is fully verified.


packages/web/app/components/issuer-dashboard/utils.spec.ts
```diff
@@ -59,13 +69,17 @@ describe("selectEtoStep", () => {
       selectEtoStep(
         true,
         EEtoState.PREVIEW,
+        undefined,
+
```
Moejoe90: we have an extra line here?

packages/web/app/components/issuer-dashboard/utils.ts
```diff
@@ -98,7 +104,24 @@ export const selectEtoStep = (
   }
 
   if (etoState === EEtoState.ON_CHAIN) {
-    return EEtoStep.REQUEST_THA_SIGN;
+    if (etoOnChainState === EETOStateOnChain.Setup) {
```
Moejoe90: This is really getting hard to understand

packages/web/app/components/nominee-dashboard/AcceptAgreement.tsx
```diff
@@ -72,15 +58,6 @@ const acceptAgreement = compose<IComponentProps, {}>(
       };
     },
   }),
-  onEnterAction<IStateProps>({
```
Moejoe90: :+1: 
Thanks for this. this was one of the points mentioned that should be moved to sagas instead of having the logic here

packages/web/app/modules/bank-transfer-flow/sagas.ts
```diff
@@ -71,7 +71,7 @@ function* stop(): any {
 function* downloadNEurTokenAgreement({ contractsService, intlWrapper }: TGlobalDependencies): any {
   const [, , agreementHashWithIpfs] = yield contractsService.euroToken.currentAgreement();
 
-  const agreementHash = agreementHashWithIpfs.replace(`${IPFS_PROTOCOL}:`, "");
```
Moejoe90: :+1:

packages/web/app/modules/eto-flow/actions.ts
```diff
@@ -12,6 +12,7 @@ export const etoFlowActions = {
   loadIssuerEto: createActionFactory("ETO_FLOW_LOAD_ISSUER_ETO"),
   loadDataStop: createActionFactory("ETO_FLOW_LOAD_DATA_STOP"),
   loadProducts: createActionFactory("ETO_FLOW_LOAD_PRODUCTS"),
+  loadIssuerStep: createActionFactory("ETO_FLOW_LOAD_ISSUER_STEP`"),
```
Moejoe90: why do we have ` in "ETO_FLOW_LOAD_ISSUER_STEP'"

packages/web/app/components/issuer-dashboard/EtoDashboard.tsx
```diff
@@ -300,14 +319,30 @@ const EtoDashboard = compose<React.FunctionComponent>(
...
     }),
     dispatchToProps: dispatch => ({
       initEtoView: () => {
-        dispatch(actions.etoFlow.loadIssuerEto());
-        dispatch(actions.kyc.kycLoadIndividualDocumentList());
+        dispatch(actions.etoFlow.loadIssuerStep());
       },
     }),
   }),
+  onEnterAction<IStateProps & IDispatchProps>({
```
Moejoe90: I'm thinking if we would be even more strict about logic inside components and call `loadIssuerStep` everytime the component updates of course the saga will know if it should load the eto or not or if the eto is loaded not change it. 

This can be part of a future plan

packages/web/app/modules/eto-flow/sagas.ts
```diff
@@ -286,6 +288,22 @@ export function* publishEtoData({
   }
 }
 
+export function* loadIssuerStep(): Iterator<any> {
+  yield neuCall(loadIssuerEto);
```
Moejoe90: What do you think of adding a caching layer. Do we really need to `loadIssuerEto` everytime maybe we should indeed.

packages/web/app/modules/eto-flow/sagas.ts
```diff
@@ -286,6 +288,22 @@ export function* publishEtoData({
...
+
+  const issuerEto: ReturnType<typeof selectNomineeEtoWithCompanyAndContract> = yield select(
+    selectIssuerEtoWithCompanyAndContract,
+  );
+
+  if (issuerEto === undefined) {
+    throw new Error("Issuer eto should be defined before loading eto agreements");
+  }
+
+  yield put(actions.eto.loadEtoAgreementsStatus(issuerEto));
```
Moejoe90: I'm not sure if its a good idea to put the issuer files in `eto` vs putting them strictly to `issuer-eto`. I think we need to try to seperate between both flows more. This might create indeed some double code but the model will be decoupled at least

packages/web/app/modules/eto-flow/selectors.ts
```diff
@@ -318,3 +324,29 @@ export const selectIsMarketingDataVisibleInPreview = createSelector(
...
+
+  return undefined;
+};
+
+export const selectAreAgreementsSignedByNominee = createSelector(
+  selectIssuerEtoAgreementsStatus,
+  documents => {
+    if (documents) {
+      return (
+        documents[EAgreementType.THA] === EEtoAgreementStatus.DONE &&
```
Moejoe90: so both agreements need to be signed. 
I was reviewing this before and was under the imporession that one gets signed or the other

packages/web/app/modules/eto/reducer.ts
```diff
@@ -81,6 +83,14 @@ export const etoReducer: AppReducer<IEtoState> = (
           [action.payload.previewCode]: action.payload.tokenData,
         },
       };
+    case actions.eto.setAgreementsStatus.getType():
```
Moejoe90: I think this really needs be in the eto-issuer model explcitly or nominee model not `eto`

packages/web/app/modules/eto/sagas.ts
```diff
@@ -492,11 +501,65 @@ function* verifyEtoAccess(
   }
 }
 
+// TODO: Move to generic eto module (the one that's not specific to investor/issuer/nominee eto)
```
Moejoe90: Lets discuss this because we can have a general module that includes only `sagas` maybe also high order `selectors` but no `actions` no `reducers` etc. to not have double code. 

But for sure every user model should have its own reducer and selectors that use these meta selectors.  to decouple these dependancies

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -55,39 +45,49 @@ export function* loadNomineeTaskData({
...
+      const nomineeRequestsConverted: TNomineeRequestStorage = nomineeApiDataToNomineeRequests(
+        taskData.nomineeRequests,
+      );
+      // todo convert data
+      // const linkBankAccountConverted = ...
+      // const acceptThaConverted = ...
+      // const redeemShareholderCapitalConverted = ...
+      // const uploadIshaConverted = ...
+
+      yield put(
```
Moejoe90: does this action set `loadingDone`?

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -246,33 +211,25 @@ export function* guardActiveEto({
...
+      > = yield select(selectActiveEtoPreviewCodeFromQueryString);
+
       // For testing purpose we can force another ETO to be active (by default it's the first one)
       const shouldForceSpecificEtoToBeActive =
-        forcedActiveEtoPreviewCode !== undefined &&
-        forcedActiveEtoPreviewCode !== previewCode &&
-        etos[forcedActiveEtoPreviewCode] !== undefined;
-
-      const doesActiveEtoExist = previewCode === undefined || etos[previewCode] === undefined;
+        forcedActiveEtoPreviewCode !== undefined && etos[forcedActiveEtoPreviewCode] !== undefined;
```
Moejoe90: looks like a good case for a selector

packages/web/app/types.d.ts
```diff
@@ -108,10 +108,15 @@ type SelectPropertyNames<T, R> = { [K in keyof T]: T[K] extends R ? K : never }[
 type PickProperties<T, R> = Pick<T, SelectPropertyNames<T, R>>;
 
 /**
- * In T, mark as required properties from K
+ * Make all properties in T required and non nullable
+ */
+export type Required<T> = { [P in keyof T]-?: NonNullable<T[P]> };
```
Moejoe90: I don't think we should call this time Required `typescript` already has this time why should we override it?

packages/web/app/test-e2e/eto-registration/EtoNomineeAcceptAgreements.spec.e2e.ts
```diff
@@ -0,0 +1,35 @@
...
+    loginFixtureAccount("ISSUER_SETUP_NO_ST", {
+      kyc: "business",
+      signTosAgreement: true,
+      clearPendingTransactions: true,
+    });
+
+    goToIssuerDashboard();
+
+    assertWaitForNomineeAgreementsStep();
+
```
Moejoe90: why not `    cy.saveLocalStorage();` 

them just restore it after logout. this will take out all the overhead related with communicating with the backend

packages/web/app/components/issuer-dashboard/EtoDashboard.tsx
```diff
@@ -342,32 +377,21 @@ const EtoDashboard = compose<React.FunctionComponent>(
...
     };
   }),
-  onEnterAction<IStateProps & IDispatchProps>({
-    actionCreator: (_, props) => {
-      if (props.userHasKycAndEmailVerified) {
-        props.initEtoView();
-      }
-    },
-  }),
-  lifecycle<IStateProps & IDispatchProps, {}>({
```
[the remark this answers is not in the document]
desfero: It's moved just above `withProps` to get correct step calculated before we reach `branch` with `LoadingIndicator` as otherwise without invoking actions and branching before we are stuck with always pending loading indicator.

packages/web/app/components/issuer-dashboard/chooseEtoStartDateWidget/ChooseEtoStartDateWidget.stories.tsx
```diff
@@ -27,5 +29,9 @@ storiesOf("ETO-Flow/ChooseEtoStartDateWidget", module)
     <EtoStartDateWidgetComponent {...data} etoDate={new Date("2028-12-20")} />
   ))
   .add("starting date cannot be changed anymore", () => (
-    <EtoStartDateWidgetComponent {...data} etoDate={new Date("2018-12-24")} />
+    <EtoStartDateWidgetComponent
+      {...data}
+      etoDate={new Date("2018-12-24")}
+      canChangeEtoStartDate={false}
```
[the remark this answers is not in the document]
desfero: yes, it's the above one (`starting date may be changed`)

packages/web/app/components/issuer-dashboard/chooseEtoStartDateWidget/ChooseEtoStartDateWidget.tsx
```diff
@@ -252,20 +258,16 @@ class DateChooser extends React.PureComponent<IDateChooserProps, IDateChooserSta
   };
 
   render(): React.ReactNode {
-    const { etoDate } = this.props;
-    const canChangeDate =
```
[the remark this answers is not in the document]
desfero: yeah, let's get rid of business logic inside components 🤗

packages/web/app/components/issuer-dashboard/utils.spec.ts
desfero: my fault, I thought prettier will fix that kind of stuff

packages/web/app/components/nominee-dashboard/AcceptAgreement.tsx
desfero: I was forced to do this to imlement it for issuer too :D

packages/web/app/modules/eto-flow/actions.ts
desfero: ouu, good catch

packages/web/app/modules/eto-flow/selectors.ts
desfero: both of them needs to be signed on nominee side to unblock issuer flow

packages/web/app/modules/nominee-flow/sagas.ts
desfero: yes, that's why I'm not a fan of keeping `loading: boolean` as you never know what's really happening and when with this `loading` :D

packages/web/app/test-e2e/eto-registration/EtoNomineeAcceptAgreements.spec.e2e.ts
desfero: fixed

packages/web/app/types.d.ts
desfero: Renamed to `RequiredNonNullable`

packages/web/app/modules/nominee-flow/sagas.ts
desfero: Do you think we should move this one to selector? It seems quite specific to this saga and it's only getting `forcedActiveEtoPreviewCode` and checking if it exists

packages/web/app/modules/eto/reducer.ts
desfero: discussed on slack

packages/web/app/modules/eto/sagas.ts
desfero: discussed on slack

packages/web/app/modules/eto-flow/sagas.ts
desfero: discussed on slack

packages/web/app/modules/eto-flow/sagas.ts
desfero: like would be good to have a caching layer, but it's something bigger to do :(

Issue #3447, opened 2019-09-13
thr form hasDividendRights fix


AlexeyTuboltsev:
**Description**

hasDividendRights in the THR form is bound to a wrong name

**Environment**

**Steps To Reproduce**

**Actual behaviour**

**Expected behavior**

**Comments, context, logs**

**Screenshots**

PR #3448, opened 2019-09-13
#3447 hasDividendRights field name fix

AlexeyTuboltsev:


packages/web/app/components/shared/forms/Form.tsx
```diff
@@ -28,7 +32,7 @@ const useFocusFirstInvalid = (isSubmitting: boolean, isValid: boolean) => {
       if (invalidInput) {
         invalidInput.focus();
       } else {
-        logger.warn(`It's not possible to focus invalid field`);
+        logger.warn(new Error("It's not possible to focus invalid field"), { errors });
```
desfero: so I have two concerns here: 
- we can potentially expose user data. An `error` object contains `key: message` where a message in some extreme situations can look like: `You have entered 1331 1313 1231 1321 3239 which is invalid credit card number` (i know it's really extreme but it's something we can easily expose to sentry by mistake)
- We use react components for message almost everywhere and after serialization, it contains a lot of messy data: 
Screenshot at Sep 15 15-15-44

Moejoe90:
I removed the QA label as there is not much to test

PR #3455, opened 2019-09-15
fixes token formatting portfolio

rudolfix:
several small improvements: please see commit list


packages/web/app/components/eto/overview/EtoOverviewThumbnail/EtoStatusManager.tsx
```diff
@@ -139,6 +139,9 @@ const EtoStatusManager = ({ eto }: IExternalProps) => {
       );
     }
 
+    case EEtoState.SUSPENDED:
+      return <></>;
```
Moejoe90: We need a ticket for this probably

packages/web/app/components/issuer-dashboard/EtoDashboard.tsx
```diff
@@ -155,6 +155,7 @@ const EtoDashboardStateViewComponent: React.FunctionComponent<IEtoStateRender> =
         </>
       );
     case EEtoState.PENDING:
+    case EEtoState.SUSPENDED:
```
Moejoe90: So `SUSPENDED` is just like `PENDING` from the issuer side really?

rudolfix: no - linter does not allow for code duplications!!

SUPSENDED is a dead end where I had to remove ETO from Universe contract. it's dead. and we do not need to cater to this too much

rudolfix: I will explain what suspended is on standup tomorrow quickly so you have all the info

glupiawymowka: we will need a ticket for this indeed, cause this state is causing minor errors, but currently, this is an internal state anyway (and let's hope we don't have to suspend any company soon 😅 )

PR #3473, opened 2019-09-18
Fixes showing false ETO dates

Moejoe90:
This fixes the bug also adds storybook for visual regression


packages/web/app/components/eto/overview/EtoTimeline/EtoTimeline.tsx
```diff
@@ -61,10 +61,17 @@ const POINTER_WIDTH = 26;
 const POINTER_CAMPAIGNING_POSITION = 20;
 const MAX_POINTER_POSITION = TIMELINE_WIDTH - 15;
 
+const MINIMUM_VALID_TIME_EPOCH = 1;
```
Moejoe90: We can increase the `MINIMUM_VALID_TIME_EPOCH` to whatever timestamp we want

PR #3478, opened 2019-09-18
fixes pre-sale visibility

rudolfix:
this does two things
- allows investor to see pre-eto data when investor has a pledge OR has locked wallet
- fixes bug to allow pre-eto data visibilty both in Setup and Whitelisting eto state and even before eto-goes on-chain

Testing notes:
User without pledge, eto state: listed, on_chain but not yet public sale
1. sees countdown to sale
2. does not see ticket discount

User pledges
1. should start seeing ticket discount and possibly counter to sale, change should happen immediately

disable whitelisting to see the countdown!

For eto in public sale state:
- ticket discount should not be shown

Affected areas
image
image


packages/web/app/modules/investor-portfolio/selectors.ts
```diff
@@ -167,13 +168,16 @@ export const selectIsWhitelisted = (state: IAppState, etoId: string) => {
...
+  const isOnBookbuildingWhitelist = selectMyPledge(state, etoId) !== undefined;
 
-  return isLockedWalletConnected || isWhitelisted;
+  return isLockedWalletConnected || isOnBookbuildingWhitelist;
 };
 
 export const selectShouldShowWhitelistDiscount = (state: IAppState, eto: TEtoSpecsData) => {
-  const isPreEto = selectEtoOnChainStateById(state, eto.etoId) === EETOStateOnChain.Whitelist;
+  const isPreEto =
+    selectEtoOnChainStateById(state, eto.etoId) ||
```
Moejoe90: If this is pre eto shouldn't the condition actually be 
`selectEtoOnChainStateById(state, eto.etoId) && EETOStateOnChain.Setup <= EETOStateOnChain.Whitelist;`

Issue #3308, opened 2019-08-12
recently claimed token not visible in MY ASSETS nor PAST INVESTMENTS until refresh


rudolfix:
**Description**
After claiming NOMERA TECH token, it is not visible in MY ASSETS, but is visible in PAST INVESTMENTS immediately after claim. You must refresh the page to see NOM in MY ASSETS

**Environment**
On `io`

**Steps To Reproduce**
Use fixture "INV_HAS_EUR_HAS_KYC" (`0xfA1Af2E251ee739F83e14d7daCfd77B3d0E930b7`)
1. log in
2. go to portfolio
3. claim NOMERA TECH
4. wait for SUCCESS
5. click "go to portfolio"

**Actual behaviour**
NOM appears in PAST INVESTMENTS but not in MY ASSETS.

**Expected behavior**
NOM appears in both places

**Technical notes**
Please add/modify e2e tests to cover this edge case

glupiawymowka:
Again, the claimed token is not appearing in "Your assets" instantly. 
This time on **stage** env 

If it's the first token that user claims, only NEU appears, every other token claimed shows up in the correct modal after changing tab/refreshing manually. 

![image](https://user-images.githubusercontent.com/18315589/72794129-c8556c80-3c3b-11ea-97bd-4f0885166e24.png)

glupiawymowka:
As the latest fixes didn't change anything on stage, waiting for the PR #3977 as the last hope

Moejoe90:
@glupiawymowka  Actually please push it back to `Ready for dev` and ping @drakkein

Moejoe90:
@glupiawymowka we need a last check if this is still relevant

Issue #3307, opened 2019-08-12
assets are missing in the portfolio


rudolfix:
**Description**
QTT token which I have balance > 0 and which is visible in PAST INVESTMENTS and in metamask is not visible in MY ASSETS

**Environment**
Found on `io`

**Steps To Reproduce**
Use fixture "INV_HAS_EUR_HAS_KYC" (`0xfA1Af2E251ee739F83e14d7daCfd77B3d0E930b7`)
1. log in
2. go to portfolio

**Actual behaviour**
there's no QTT token in MY ASSETS, but it's in past investments

**Expected behavior**
QTT token visible in MY ASSETS and PAST INVESTMENTS

**Comments, context, logs**
additionally confirm by adding QTT to metamask as custom token (token address ` 0x2d96e2a0f117232debc9ea4e780a4b5c1ee5aa73`)

also note that MY ASSET balance of QTT should be less than PAST INVESTMENT (purchased tokens) due to token transfer
![image](https://user-images.githubusercontent.com/17202864/62884363-9fe55000-bd36-11e9-9181-5b47569120c3.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/17202864/62880481-be931900-bd2d-11e9-9e34-0cb99e1094af.png)

desfero:
@rudolfix this is because we show only assets that reached `payout` state

rudolfix:
so it must be fixed!

PR #3496, opened 2019-09-20
#3308 #3307 Fix My Assets 

drakkein:
SSH token should now be showed 
image

**Testing notes**
Both #3308 and  #3307 are fixed here


packages/web/app/test-e2e/portfolio/TokenClaim.spec.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+        { submit: false },
+      );
+
+      cy.get(tid("modals.tx-sender.user-claim-flow.summary.accept")).click();
+
+      acceptWallet();
+
+      cy.get(tid("modals.tx-sender.user-claim-flow.success")).should("exist");
+      cy.get(tid("modals.tx-sender.user-claim-flow.success.go-to-portfolio")).click();
+
```
desfero: Would be good to add `assertPortfolio` here IMO

packages/web/app/test-e2e/portfolio/TokenClaim.spec.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+      clearPendingTransactions: true,
+    }).then(() => {
+      goToPortfolio();
+
+      cy.get(tid(`modals.portfolio.portfolio-asset-action.claim-${ETO_ID}`)).click();
+
+      fillForm(
+        {
+          readDocuments: {
+            type: "checkbox",
```
desfero: use `toggle` type, `{false: true}` is really hard to read

packages/web/app/test-e2e/portfolio/TokenClaim.spec.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+import { acceptWallet } from "../utils/index";
+
+describe("Token claim", () => {
+  it("claimed token should show in My Assets table", () => {
+    const ETO_ID = etoFixtureAddressByName("ETOInPayoutState");
+
+    loginFixtureAccount("INV_HAS_EUR_HAS_KYC", {
+      kyc: "business",
+      clearPendingTransactions: true,
+    }).then(() => {
```
desfero: there is no need to `then` here we should await nesting when it's possible

packages/web/app/test-e2e/portfolio/TokenClaim.spec.e2e.ts
```diff
@@ -0,0 +1,38 @@
+import { etoFixtureAddressByName, goToPortfolio, loginFixtureAccount, tid } from "../utils";
+import { fillForm } from "../utils/forms";
+import { acceptWallet } from "../utils/index";
```
desfero: Remove `index`, you even can combine all imports here into one

packages/web/app/test-e2e/portfolio/TokenClaim.spec.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      cy.get(tid("modals.tx-sender.user-claim-flow.summary.accept")).click();
+
+      acceptWallet();
+
+      cy.get(tid("modals.tx-sender.user-claim-flow.success")).should("exist");
+      cy.get(tid("modals.tx-sender.user-claim-flow.success.go-to-portfolio")).click();
+
+      cy.get(tid(`portfolio-my-assets-token-${ETO_ID}`)).should("exist");
+      cy.get(tid(`past-investments-${ETO_ID}`)).should("exist");
```
desfero: Let's also assert that it's not in the `RESERVED ASSETS` anymore

packages/web/app/modules/investor-portfolio/selectors.ts
```diff
@@ -66,7 +66,7 @@ export const selectMyAssets = (state: IAppState): TEtoWithCompanyAndContract[] |
   const etos = selectEtos(state);
 
   if (etos) {
-    return etos.filter(eto => eto.contract && eto.contract.timedState === EETOStateOnChain.Payout);
+    return etos.filter(eto => eto.contract);
```
desfero: Hm, this one is probably bad for performance as `selectMyAssets` is also used in `loadTokensData`. Maybe we can filter `Payout` and `Claim` state only (please confirm with @rudolfix )

rudolfix: this was my quick fix locally to see claimed assets
`return etos.filter(eto => eto.contract && [EETOStateOnChain.Payout, EETOStateOnChain.Claim].indexOf(eto.contract.timedState));`

desfero: This is what we should use in this case IMO, but @drakkein, please use `includes` over `indexOf`. Maybe even for readability `return etos.filter(isOnChain).filter(eto => [EETOStateOnChain.Payout, EETOStateOnChain.Claim].includes(eto.contract.timedState))`

Issue #3421, opened 2019-09-08
Company tags input is broken


desfero:
**Description**
After some recent changes we have introduced a regression bug into our `Tags` component.
![image](https://user-images.githubusercontent.com/7031578/64494853-96e69080-d29a-11e9-9b9b-6f4449a097ec.png)
 

**Environment**
production

**Steps To Reproduce**
Login as an `Issuer`
Go to `Company Information` form
Scroll to `Company tags` field
Click on an `input`

**Actual behavior**
UI is broken

**Expected behavior**
Should look as it was before

**Comments, context, logs**

**Screenshots**
![image](https://user-images.githubusercontent.com/7031578/64494863-cb5a4c80-d29a-11e9-8ada-11a87449638c.png)

glupiawymowka:
tested on stage

PR #3498, opened 2019-09-20
#3421 Fix tags import

drakkein:


packages/web/app/components/eto/shared/EtoTagWidget.unsafe.tsx
```diff
@@ -11,6 +11,9 @@ import { FormFieldError } from "../../shared/forms";
 import { FormFieldLabel } from "../../shared/forms/fields/FormFieldLabel";
 import { ETagSize, Tag } from "../../shared/Tag";
 
+import "react-select/dist/react-select.css";
```
desfero: Let's do a proper fix here as it's easy to break again in the future. Let's move all usages of `react-virtualized-select` `Select` component as a separate component inside our `forms` folder and replace all imports to use own implementation.

PR #3500, opened 2019-09-20
[WIP] Fix an error thrown while loading equity tokens data

desfero:


packages/web/app/modules/investor-portfolio/selectors.ts
```diff
@@ -222,9 +222,10 @@ export const selectPayoutAvailable = (state: IAppState) => {
 };
 
 export const selectMyAssetsWithTokenData = (state: IAppState): TETOWithTokenData[] | undefined => {
-  const myAsssets = selectMyAssets(state);
-  if (myAsssets) {
-    return myAsssets.map((asset: TEtoWithCompanyAndContract) => ({
+  const myAssets = selectMyAssets(state);
```
desfero: Just typo fix (`myA*sss*ets`)

packages/web/app/modules/eto/sagas.ts
```diff
@@ -396,13 +396,8 @@ function* loadTokensData({ contractsService }: TGlobalDependencies): any {
...
     // obtain nominal value of a share from IEquityToken
-    let shareNominalValueUlps;
-    try {
-      shareNominalValueUlps = yield equityToken.shareNominalValueUlps;
-    } catch (e) {
-      // make it backward compatible with FF ETO, which is always and forever Q18 and does not provide method above
-      shareNominalValueUlps = Q18;
-    }
+    // make it backward compatible with FF ETO, which is always and forever Q18 and does not provide method above
+    const shareNominalValueUlps = yield defaultTo(equityToken.shareNominalValueUlps, Q18);
```
rudolfix: `equityToken.shareNominalValueUlps` in fact will call smart contract via `eth_call` of `shareNominalValueUlps` which in case FF ETO does not implement it. `VM Exception` will be returned and this should be filtered out here to be 100% correct. `defaultTo` will not help in this case.

Please check the code above for ELDO token on **stage**. you can use DEPLOYER account with address 0x8a194c13308326173423119F8dCb785CE14C732B to log in.

desfero:
closing for now

Issue #3436, opened 2019-09-11
Transaction history amounts aren't visible on screens smaller than 768 pixels in width


andywalton:
**Description**
Hello, here's a simple bug.

The transaction history items have an "amount" field that isn't visible on screens smaller than 768 pixels.

You therefore see this:
![Screenshot 2019-09-11 12.20.22.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/412b26f1-9caf-44af-8151-5a5b0ca840ae)

This is a very quick CSS fix.

The problem is that one of the NewTable classes has a min-width set to 200% as follows:
```
.NewTable-module__table___3H9Gm {
    min-width: 200%;
}
```

Changing this back to 100% fixes it. That 200% must've been a type?

**Environment**
Responsive web and mobile

**Steps To Reproduce**
Visit https://platform.neufund.io/wallet and view your Wallet transaction history. Resize the browser so that the screen is less than 768 pixels in width.

You'll see you can no longer see the transaction amounts.

**Actual behaviour**
Transaction history amounts are hidden on small screens.

**Expected behavior**
You can see the transaction amount, floated and visible to the right of the transaction item description.

The entire amount should always be visible.

**Comments, context, logs**
~

**Screenshots**
Before fix:
![Screenshot 2019-09-11 12.20.22.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/06a356c9-92aa-4188-8d7c-56a5a4fdd2c8)

After fix:
![Screenshot 2019-09-11 14.16.47.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/653bfd74-28cf-4027-babf-c8ac88e53542)

![Screenshot 2019-09-11 14.17.32.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/bf90396a-05a3-4a73-87dc-c2efeac0eaea)

glupiawymowka:
tested on stage

Issue #3424, opened 2019-09-09
Add transaction direction to our Wallet transaction history items


andywalton:
**Background**
Currently our transaction history on the wallet view is missing a transaction direction indicator which makes it hard to know if the transaction direction is out or in. We highlight transactions that are incoming by changing the colour to green, which creates a bit of meaning. However, colour alone isn't an ideal solution IMO. What if you're colour blind?

Anyway, the solution I'm recommending is to display a minus (negative) symbol that displays for all transactions with that are marked as `transaction direction = out`.

**Designs**
The below shows a "-" next to the out transaction.
<img width="967" alt="Screenshot 2019-09-09 09 49 55" src="https://user-images.githubusercontent.com/609657/64512671-316cc100-d2e7-11e9-8a84-6a8c5c32780a.png">

**Acceptance criteria**
- [ ] All transactions that have `transaction direction = out` show a minus character to the left of the value as per the Design screenshot above.

**Technical notes**
To determine when to show the minus listen on the transaction object for the property:
`"transaction_direction": "out",`

For `"transaction_direction": "in",` we show the value of the transaction as green text.

Please let me know if there is a 3rd / 4th transaction direction value I might need to consider how we handle. Ie. What do we do for a transfer transaction? I assume it will show the direction = out?

@rudolfix - please will you review this ticket and add any more technical thoughts to it.

**Testing notes**
As per the acceptance criteria.

Please speak to me @andywalton if you have any questions.

glupiawymowka:
tested on stage

PR #3523, opened 2019-09-24
#3424 #3436 Minor fixes to transactions history

desfero:
**Testing notes**
#3424 #3436 should be fixed


packages/web/intl/locales/en-en.json
```diff
@@ -1459,4 +1459,4 @@
   "withdraw-flow.support-link": "here",
   "your-ethereum-address-widget.explanation": "This is the wallet address you are registered with on Neufund.",
   "your-ethereum-address-widget.header": "Wallet address"
-}
\ No newline at end of file
+}
```
Moejoe90: Why did this happen do you have a different linter / formatter version?

desfero: I really don't know, maybe it's the difference between Mac and Linux? Will try to investigate soon

Issue #3170, opened 2019-07-02
[REFUND] - flow improvements and copy changes


glupiawymowka:
**Background**

We've got a working refund flow now, but as the initial ticket was not taking some cases into consideration, we still do need to improve wording and add some features to it. 

Current issues: 
- nEUR investment is not taken into consideration, the refund summary only shows ETH value 
- it is unclear to the user where the refunded money goes 

**Designs**

*please add if needed*

**What should be changed** 

1. Please implement a variable that shows separately ETH AMOUNT REFUNDED if there was an ETH investment, nEUR AMOUNT REFUNDED if there was an nEUR investment, or both if applicable. 
Currently, we always show the amounts in ETH, even if nEUR is also being refunded 
Fixture to test it: 
0xfA1Af2E251ee739F83e14d7daCfd77B3d0E930b7
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/7c232f2b-8c5b-456e-a28e-09ef612c0e2b)

2. Please change the copy as follows: 
"Unfortunately, the 'COMPANY NAME' offering you invested in did not reach the soft cap (?? I think this is the word, but maybe we're not using it for a reason, please correct me if I'm wrong). Please continue to refund your invested funds back to the wallet you invested with." 

3. On success screen, please add "go to wallet" button 

**Acceptance criteria**
- [x] go to wallet button available on success 
- [x] variable AMOUNT REFUNDED depending on investment currency 
- [x] a proper /CURRENCY/ AMOUNT REFUNDED visible on transaction summary, pending and success state

**Technical notes**

**Testing notes**

PR #3530, opened 2019-09-25
#3170 Improve refund flow

drakkein:


packages/web/app/components/modals/tx-sender/investor-refund/RefundDetails.tsx
```diff
@@ -25,28 +25,35 @@ const RefundTransactionDetails: TransactionDetailsComponent<ETxSenderType.INVEST
       value={`${additionalData.tokenName} (${additionalData.tokenSymbol})`}
     />
 
-    <InfoRow
-      caption={<FormattedMessage id="user-refund-flow.amount" />}
-      value={
-        isZero(additionalData.amountEth) ? (
+    {!isZero(additionalData.amountEurUlps) && (
```
Moejoe90: why do you check if this value is `Zero`? 
What's the idea behind this

packages/web/app/components/modals/tx-sender/investor-refund/RefundSummary.stories.tsx
```diff
@@ -32,19 +33,31 @@ const additionalDataNEur: TAdditionalDataByType<typeof ETxSenderType.INVESTOR_RE
   amountEurUlps: Q18.mul(150).toString(),
 };
 
+const bothAdditionalData: TAdditionalDataByType<typeof ETxSenderType.INVESTOR_REFUND> = {
```
Moejoe90: `bothAdditionalData` is confusing name what does this mean?

packages/web/app/components/modals/tx-sender/shared/TxAction.tsx
```diff
@@ -0,0 +1,45 @@
...
+        >
+          <FormattedMessage id="modals.shared.tx-action.go-to-wallet" />
+        </ButtonArrowRight>
+      );
+    default:
+      return null;
+  }
+};
+
+const TxAction = compose<TComponentProps, IExternalProps>(
```
Moejoe90: the name `TxAction`  sounds very general while in reality this button specializes on going to wallet. 

please make the correct changes in terms of name to reflect that. 

Also maybe its a good idea actually to make `TxActionLayout` a general component and just export a couple of actions including `txActionGoToWallet` also while you're at it please refactor all existing goToWallet call to actions to use tha me component

packages/web/app/test-e2e/eto/Refund.spec.e2e.ts
```diff
@@ -52,4 +53,62 @@ describe("Refund", () => {
...
     });
   });
+
+  it("Refund from ETO Page with wallet investment (nEUR and ETH)", () => {
+    const ETO_ID = etoFixtureAddressByName("ETOInRefundState");
+
+    loginFixtureAccount("INV_HAS_EUR_HAS_KYC", {
+      kyc: "business",
+      clearPendingTransactions: true,
+    }).then(() => {
```
Moejoe90: you can get rid of `then` for `loginFixtureAccount` we don't need to use it anymore

packages/web/app/test-e2e/utils/index.ts
```diff
@@ -322,8 +322,10 @@ export const parseAmount = (amount: string) => new BigNumber(amount.replace(/\s|
 /**
  * Get eth wallet balance
  */
-export const getWalletEthAmount = () => {
-  goToWallet();
+export const getWalletEthAmount = (navigate: boolean = true) => {
```
Moejoe90: change `navigate` to `navigateToWallet`

packages/web/app/test-e2e/eto/Refund.spec.e2e.ts
```diff
@@ -52,4 +53,62 @@ describe("Refund", () => {
...
+            // balance after refund should be increased by the refund amount
+            // there can be rounding issue so we assert balance with small delta
+            expect(balanceAfter.minus(balanceBefore.plus(amount))).to.be.bignumber.lessThan(0.01);
+          });
+        });
+      });
+
+      cy.visit(etoPublicViewByIdLinkLegacy(ETO_ID));
+
+      cy.get(tid("eto-overview-refund-claimed")).should("exist");
```
Moejoe90: This and line 29-53 can be extracted into one method to have less code 
this method takes [] of data-test-id and just maps through the flow again and again 
No need to adding all these lines of code

packages/web/app/components/modals/tx-sender/investor-refund/RefundDetails.tsx
drakkein: Basically I'm checking if row should be displayed if amount is not `0` or empty string. `0` should never happen so this can be changed to `!== ""` but this is safer IMO.

packages/web/app/components/modals/tx-sender/shared/TxAction.tsx
drakkein: I was thinking about extending it in future to contain more actions for generic success modal of different transaction types, so that i came to TxAction name, let's talk about it a little bit.

Moejoe90: ok but then lets name it like that when it is generic. as currently the name doesn't reflect the action

drakkein: ok, I renamed it to `TxGoToAction` as it only changes route for now

PR #3538, opened 2019-09-26
Fixes Pledge Loading state

Moejoe90:


packages/web/app/components/eto/overview/EtoOverviewStatus/EtoStatusManager/PublicSaleTimer.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react";
```
Moejoe90: This is a temporary fix in order to load the correct state when the state changes and we have out of date pledge data

packages/web/app/components/eto/overview/EtoOverviewStatus/EtoStatusManager/EtoStatusManager.tsx
```diff
@@ -82,9 +82,9 @@ const EtoStatusComponentChooser: React.FunctionComponent<IStateProps & IExternal
         return <InvestmentWidget eto={eto} isEmbedded={isEmbedded} />;
       } else {
         return (
-          <PublicSaleTimer
```
Moejoe90: @rudolfix I decided to revert this change after thinking about it. 
If the user pledged at any point before the whitelisting. then the already loaded `pledgeData` is updated and the swtich should happen correctly. 

There is a small underlying issue here which was what created the bug in the first place. The fact is we assume even pledge is loading that the user is not elgible. this causes this component `PublicSaleTimer` to load. 
I started making the change but found that this selector we use is already in a couple of places. 

I can start an overhaul now but i think this is not best time to do it. In all cases tell me what you think is best and I will push forward accordingly.

Issue #1497, opened 2018-10-05
Calculate timeline pointer in campaigning state


desfero:
Background:
Currently, ETO timeline campaigning state day pointer is always set to a constant value.
![image](https://user-images.githubusercontent.com/7031578/46521098-bb620d80-c87e-11e8-90b7-8068f26668e4.png)

But if the dates are know we can correctly calculate date pointer.

**technical details**
* We can calculate the pointer only when we know start date (returned by the contract)
* We known the whole span of the ETO in seconds and also in pixels
* We can calculate the visible span before the WHITELISTED starts in pixesl and we know how many seconds it covers
* If current date < start of WHITELISTED - calculated span then we display as on the image above (initial position)
* if we are in range we start to move the cursor

This is rough idea implement something akin to this. The cursor starts moving when we are in displayed range

SergejSoft:
is it still relevant?

desfero:
probably yes, campaigning pointer position is still hardcoded on timeline

rudolfix:
here's test case on `io` where indicator should not be "parked"
https://platform.neufund.io/eto/view/LI/ce390a89-be03-45a6-9360-af7d1e019085

glupiawymowka:
tested on stage

PR #3557, opened 2019-10-01
#1497 calculate timeline pointer in campaigning

desfero:
**Technical notes**
Promise related utils were moved to single `Promise.utils` file for consistency.

**Testing notes**
There is an ETO where pointer in the campaigning state shouldn't be parked. https://localhost:9090/eto/view/LI/9cc3a0dd-729e-4b0c-846e-06d82334781d (`ISSUER_SETUP LTD.`). Please to make cursor moving modify the start date for the given ETO so it's easier to see the changes in position


rudolfix:
TESTING:
please use ISSUER SETUP on `io` and change the start date to 8 or 7 days to pre-sale.

the pointer should move from the default position

rudolfix:
btw. check on codechecks that it MOVES. so moving to QA column for `io` check

Issue #2118, opened 2018-11-28
Test and Upgrade Redux-Saga


Moejoe90:
**Background**
We are in `pre-release` version 0.16.x 
the latest version is `1.0.1` There are some breaking changes but many great features and implementations that will help make the whole application tighter 
**Task**
- [ ] Upgrade to the latest version and make sure everything is running
- [ ] Remove from `neu*` wrappers (`neuCall`, etc) casting `saga as CallEffectFn<any>` as it should be no longer needed 
- [ ] Refactor our `neu*` wrappers (`neuCall`, etc) to spread `args` to the effect. (At the moment they are passed as `args[0], args[1], args[2], args[3], args[4]`.

Moejoe90:
There is currently no clear API interface of documentation for changes things like 
`delay` are not there. 
It doesn't make sense to upgrade now until the stable version is released at some point

desfero:
Ok let's push this as the stable version was already released more month ago

Moejoe90:
Update there is already V1.1.3 
https://github.com/redux-saga/redux-saga/releases
Pushing to Christmass tech depth

PR #3561, opened 2019-10-01
Improve generated types for neu* saga wrappers

desfero:
**Technical notes**
You may see some casting `saga as CallEffectFn<any>` in a couple of places. We need it until we migrate to `redux-saga` 1.*, therefore, I have added a note in #2118


packages/web/app/modules/tx/transactions/investment/sagas.ts
```diff
@@ -32,7 +32,7 @@ export const INVESTMENT_GAS_AMOUNT = "600000";
 function* getEtherTokenTransaction(
   { contractsService }: TGlobalDependencies,
   etoId: string,
-  investAmountUlps: string,
+  investAmountUlps: BigNumber,
```
Moejoe90: good catch

packages/web/app/modules/tx/user-flow/withdraw/sagas.ts
```diff
@@ -48,7 +48,7 @@ export function* detectMaxWithdraw(
 
   if (fixedEther === value) {
     try {
-      yield neuCall(isAddressValidAcceptsEther, modifiedTo);
+      yield neuCall(isAddressValidAcceptsEther, modifiedTo, modifiedValue);
```
Moejoe90: :scream:

PR #3572, opened 2019-10-02
Metamask Fix

gregzaitsev:


desfero:
@gregzaitsev thanks a lot for pointing where we need to dive deep and search for a problem/bug. 

After quick searching through the code I found that our `promisify` function is only expecting `error` to be either `null` or `Error` object (and that's valid due to the way callback works in node js) but for some reason, metamask mobile is returning `undefined` as an `error` in case we don't have any errors. Fixed in #3665

Issue #3565, opened 2019-10-02
Change wording in document sections of nominee, issuer and listing page


pstroinski:
Please change the following:
- "Preview of Investment and Shareholder Agreement" to "Investment and Shareholder Agreement" in the document section & listing page (ETO terms / legal documents section)
After the ETO we should display "Signed Investment and Shareholder Agreement" in the document section & listing page (ETO terms / legal documents section)

glupiawymowka:
tested on stage

PR #3582, opened 2019-10-06
#3565 Change wording of Preview of ISHA to just ISHA

desfero:


packages/web/app/lib/api/eto/EtoFileApi.interfaces.ts
```diff
@@ -80,5 +80,5 @@ export const immutableDocumentName: { [key in EEtoDocumentType]: string } = {
   approved_investor_offering_document: "Approved Offering Document",
   signed_termsheet: "Signed Termsheet",
   signed_investment_and_shareholder_agreement: "Signed Investment and Shareholder Agreement",
-  investment_and_shareholder_agreement_preview: "Preview of Investment and Shareholder Agreement",
+  investment_and_shareholder_agreement_preview: "Investment and Shareholder Agreement",
```
desfero: in general, we should get rid of `immutableDocumentName` as it's a duplicate, so let's keep that in mind

Issue #3581, opened 2019-10-05
show countdowns to (pre)sale instead of whitelist limit reached


rudolfix:
**Background**
Current implementation of states in `getEtoSubState` puts `WHITELISTING_LIMIT_REACHED` over any other state. in on chain this will prevent countdowns to be shown.

here's how we should fix it (final fix would decouple LIMIT REACHED from states but let's not touch it).

1. if there's start date and whitelisting is disabled (isBookbuilding===false) we show countdowns no matter what
2. else we have our current code

intention of (1) is to show countdowns when we are within 7 days (so you cannot change start date anymore) from pre-sale.

**Designs**
We do not change UI

**Testing notes**
OK - we do not have a fixture exactly in that state. you can produce one by setting start date to minimum possible value and then wait for a few minutes.
Also the api mock call to change whitelist limit is coming

desfero:
Fixed in https://github.com/Neufund/platform-frontend/pull/3583

glupiawymowka:
tested on stage

Issue #3568, opened 2019-10-02
ETO card should make the Whitelist status (closed) more visible


pstroinski:
**Background**
We should have a more visible status of the closed whitelist with the committed amount. Something that will correspond more to the one we have on the website (Charlene design)

**Designs**
Charlene or to be created

**Acceptance criteria**
Visible amount of funds committed during the Whitelisting period

**Testing notes**
As on design

andywalton:
@pstroinski 

Here are is a solution I'd recommend for this:

We should create a new component based off our ETO success state card.

So create a **new component** based off this:
<img width="405" alt="Screenshot 2019-10-02 17 10 55" src="https://user-images.githubusercontent.com/609657/66056565-9d81c400-e537-11e9-8e00-75a673df10da.png">

ie. the `EtoStatusManager-module__successful-info` component.

And make it look like so:
<img width="391" alt="Screenshot 2019-10-02 17 04 11" src="https://user-images.githubusercontent.com/609657/66058814-4da4fc00-e53b-11e9-8d4c-f916c0eda603.png">

This only css change required is:
`background: #E7EBEE;`

And the text:
Info text for the case of all whitelisting slots filled:
`Whitelist closed - all X slots filled`

If not all slots are filled then:
`Whitelist closed - X of Y slots filled`

Then add how much was committed during the whitelist:
`X EUR committed`

**Show this when:**
- [ ] show this when the whitelist is closed and no campaign dates have been set for sales (ie. no countdowns are active)
- [ ]

Issue #3485, opened 2019-09-19
[ETO cards] Visual layout fix


andywalton:
Increase the width of the `.EtoOverviewThumbnail-module__label` to 60% so that labels for the example image above aren't wrapping onto two lines.

For example look at our test ETO below:
![Screenshot 2019-09-19 15.26.40.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/26d492c0-7613-4678-9c31-0ee877f5c12f)

After the CSS width of the class is set at 60% if visually looks better as Company Headquarters for example is on a single line:
![Screenshot 2019-09-19 15.25.39.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/6a096b62-df1b-4133-b15a-427abfc896e8)

desfero:
Fixed in https://github.com/Neufund/platform-frontend/pull/3583

glupiawymowka:
tested on stage

PR #3583, opened 2019-10-07
#3568 Add grey info box after whitelisting limit was reached 

desfero:
**Testing notes**
This one includes a couple of tickets at once: #3568 #3581 #3485


packages/web/app/components/eto/overview/EtoOverviewThumbnail/Info.module.scss
```diff
@@ -1,7 +1,10 @@
 @import "../../../shared/Panel.module";
 
+$grey-info-background: #e7ebee;
```
drakkein: Maybe we should add this color to our `colors.scss` file

packages/web/app/components/eto/overview/EtoOverviewThumbnail/EtoCardStatusManager.tsx
```diff
@@ -10,53 +9,53 @@ import {
...
   switch (state) {
     case EEtoState.LISTED:
     case EEtoState.PROSPECTUS_APPROVED:
     case EETOStateOnChain.Setup: {
+      // if start date was already set show countdown over whitelist components
+      if (
+        eto.subState === EEtoSubState.COUNTDOWN_TO_PRESALE ||
+        eto.subState === EEtoSubState.COUNTDOWN_TO_PUBLIC_SALE
+      ) {
+        const nextState =
```
drakkein: I'm not sure, is this the only place where we calculating next state? Maybe it is worth to create some utility function to compute it?

packages/web/app/components/eto/overview/EtoOverviewThumbnail/Whitelist/Whitelist.tsx
```diff
@@ -94,6 +141,18 @@ const Whitelist = compose<React.FunctionComponent<IExternalProps>>(
       };
     },
   }),
+  withProps<IWithProps, IStateProps & IExternalProps>(({ eto, investorsCount }) => {
+    const bookbuildingLimitReached = eto.maxPledges - investorsCount === 0;
```
drakkein: I know it should not be possible to have negative value here, but I remember that there was some issue with data not locked on time, maybe we should change it to `<=` to have it covered in case?

packages/web/app/components/eto/overview/EtoOverviewThumbnail/Info.module.scss
desfero: ok, I used a similar color from our `colors` pallet to not duplicate

packages/web/app/components/eto/overview/EtoOverviewThumbnail/EtoCardStatusManager.tsx
desfero: in this context, it's only here so lets for now keep it as it is, ok?

packages/web/app/components/eto/overview/EtoOverviewThumbnail/Whitelist/Whitelist.tsx
desfero: good catch, changed to `<=`

rudolfix:
test results are OK
- #3581 checked directly on production data
- #3568 checked on various combination (on chain/not on chain/no start date/with start date) on io.

Issue #3486, opened 2019-09-19
[ETO cards] - Add conditional copy for whitelisted investor for presale state


andywalton:
When an Issuers campaign's Presale starts, please update the ETO card to inform logged in Investors who are on the Whitelist that they can Invest Now via viewing the offer page.

Ie. the below is what you would see if you weren't logged into the platform and/or you are logged in as an Investor but you haven't subscribed to the ETO Whitelist.

![image](https://user-images.githubusercontent.com/17202864/66398319-d9a4a100-e9dd-11e9-99fb-89b42a3e030c.png)

However, if you have subscribed to the whitelist and you are logged in Investor, you should see the following:
![Screenshot 2019-09-19 15.38.37.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/b173475c-3912-46d0-9434-bf4c1e25340b)

**Done when**
- [ ] Logged in Investors who are on the whitelist for a campaign, will see an ETO card for that campaign that tells them this message: "PRESALE STARTED - VIEW OFFER"
- [x] For users who are NOT logged in as well as Investors who ARE logged but they ARE NOT subscribed to the campaign Whitelist, they will see a message that current reads: "X DAYS TO GO UNTIL PUBLIC SALE STARTS". Please also change this copy to rather read: "PUBLIC SALE STARTS IN X DAYS"

X = the number of days until the Public Sale starts.

rudolfix:
this is same problem as with main widget - state machine is convoluted and probably this state was not taken into account.

glupiawymowka:
waiting for presale on stage

PR #3588, opened 2019-10-08
corrects #3486 - displays countdown to public sale for unlogged/ineligible investors

rudolfix:
ticket #3486 specified that pre-sale stats should be visible to unlogged/ineligible investors. that was a mistake that was corrected, please check
https://github.com/Neufund/platform-frontend/issues/3486
for the correction (correct design was put there0

this PR fixes that problem. tested manually on `io` and expected state is there:
image


glupiawymowka:
waiting for presale on stage

PR #3600, opened 2019-10-09
[DO NOT MERGE] removes ledger watcher

rudolfix:
objectives
1. read the code review in the PR
2. we have too many watchers for the ledger connection started from many different places. like 5?
3. some of the watchers do not wait for action to be completed and just reissue connection. this should not happen
4. this applies to ledger and browser wallet IMO
5. those actions and connection establishing in the loop - does not seem there's a way to cancel them!

ledger fix
1. minimize transport creation as per comments in PR. we can cache opened transport and reopen it seamlessly on error. that will minimize popups on windows (and is also a good practice)
2. we check firmware and do other checks on every transport open. does not make sense. IMO this is why ledger slowed down recently (if you cahce transport then we can check it for every new transport obtained to detect connecting different ledger)
3. consider dropping testing connection on ledger altogether but for sure on windows machines

the windows popup happens when transport.create() is called. this needs to be minimized

we also have a timeout parameter when creating transport. please investigate if we can use long timeout periods


desfero:
Fixes are gonna be handles as a separate PR's

Issue #3585, opened 2019-10-08
Bring back whitelisting stats till end of pre-sale


pstroinski:
**Background**
We need to enable download of CSV file (whitelisted users) till the end of pre-sale stage.

**Acceptance criteria**
- [ ] Available whitelisting module till the end of pre-sale stage

PR #3605, opened 2019-10-10
#3585 Fix whitelisting stats before presale

desfero:
**Testing notes**
You can use `ISSUER_SETUP_NO_ST` to check the state after whitelisting is disabled


packages/web/app/components/issuer-dashboard/bookBuildingWidget/BookBuildingWidget.tsx
```diff
@@ -303,4 +298,23 @@ export const BookBuildingWidget = compose<TProps, IExternalProps>(
       };
     },
   ),
+  branch<IStateProps & IWithProps>(props => {
```
drakkein: I would move it to utility function

Issue #3597, opened 2019-10-09
Issuer flow - presale started


pstroinski:
**Background**
Issuer should see a proper status of the ETO (Pre-sale) during the presale stage.
We should also display two modules:
- whitelisting closed (and csv available to download)
- countdown to the end of presale/start public sale

Additionally, we should show at least simple stats of the investments (whitelist vs invested or full tab with stats from monitor)

**Designs**

**UPDATED with designs, please see designs and notes in Zeplin here: [https://zpl.io/VO1J7dL](https://zpl.io/VO1J7dL)**

Countdown and whitelist should look more less like this (we have those modules already designed in different state, whitelist module is already implemented with the proper closed status) - PLEASE USE UPDATED Zeplin design:
![should_be_in_this_format_with_different_data](https://user-images.githubusercontent.com/27957897/66492589-f4961480-eab4-11e9-8428-e9cafc5ec39a.png)

**Acceptance criteria**
- [ ] Available module with whitelist closed and with option to download csv
- [ ] Countdown to the end of the presale and start of the public sale 
- [ ] Statistics of current investment status

PR #3613, opened 2019-10-11
#3597 Add fundraising statistics and presale time counter to ETO Dashboard

drakkein:
image

Need confirmation from @rudolfix if usage of `totalEquivEurUlps` from `contract.totalInvestment` is correct, since on testing ETO data `euroTokenBalance + (etherTokenBalance * etherPriceEur)` gives different value here. 
ex. `578.0026 ETH * 168.229 + 100 000.43 nEUR = 97 236.80 EUR + 100 000.43 nEUR = 197 237.23 EUR` while  value from `totalEquivEurUlps` is `308 648.93 EUR`.


rudolfix:
you are right. `totalEquivEurUlps` is calculated from particular investment tickets and spot price of ether is used for each ticket - at the moment of making transaction. ether price varies over time and the result is different then just multiply of total ether * ether price after ETO.

so all looks good,

packages/web/app/components/issuer-dashboard/EtoDashboard.tsx
```diff
@@ -199,6 +203,13 @@ const EtoDashboardStateViewComponent: React.FunctionComponent<IEtoStateRender> =
     case EEtoState.ON_CHAIN:
       return (
         <>
+          {etoStep === EEtoStep.FUNDRAISING_IS_LIVE && (
+            <>
+              <ETOFundraisingStatistics eto={eto} columnSpan={EColumnSpan.ONE_AND_HALF_COL} />
```
rudolfix: this will also display in public sale, right? we can ignore for now but, if so, we need to fix it in two weeks

drakkein: It should not, I'm gonna add e2e to cover this case.

packages/web/app/components/issuer-dashboard/EtoDashboard.tsx
```diff
@@ -199,6 +203,15 @@ const EtoDashboardStateViewComponent: React.FunctionComponent<IEtoStateRender> =
     case EEtoState.ON_CHAIN:
       return (
         <>
+          {etoStep === EEtoStep.FUNDRAISING_IS_LIVE &&
+            eto.contract &&
+            eto.contract.timedState === EETOStateOnChain.Whitelist && (
```
desfero: maybe it's enough if we check and always show widget for `timedState === `EETOStateOnChain.Whitelist`?

packages/web/app/components/issuer-dashboard/ETOPresaleCounterWidget.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import * as styles from "./ETOPresaleCounterWidget.module.scss";
+
+interface IExternalProps {
+  eto: TEtoWithCompanyAndContract;
+}
+
+type TProps = IExternalProps & IPanelProps;
+
+const ETOPresaleCounterWidget: React.ComponentType<TProps> = ({ eto, columnSpan }) => {
```
desfero: Just for consistency, let's use `React.FunctionComponent`

packages/web/app/components/issuer-dashboard/ETOPresaleCounterWidget.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  & td {
+    padding-right: .5rem;
+  }
+}
+
+.date-bold {
+  font-weight: bold;
+}
+
+.eto-time-left-start {
```
desfero: probably this one is not used? (as a side note, we should probably research a way to make generated styles type-safe so we can get warnings when something is not used as right now all unused styles are bundled into the code

packages/web/app/components/issuer-dashboard/ETOPresaleCounterWidget.module.scss
```diff
@@ -0,0 +1,31 @@
...
+  align-items: center;
+  width: 100%;
+}
+
+.date {
+  font-weight: normal;
+  width: auto;
+  border: 0;
+
+  & td {
```
desfero: `&` is not needed probably. Btw, it's always better to provide special classes and it's also better to use `em` in the below case

packages/web/app/components/issuer-dashboard/ETOPresaleCounterWidget.module.scss
```diff
@@ -0,0 +1,31 @@
+.eto-public-start-date {
+  display: flex;
+  flex-flow: column;
+  justify-content: center;
+  align-items: center;
+  width: 100%;
+}
+
+.date {
+  font-weight: normal;
```
desfero: font-weight: normal is not needed?

packages/web/app/components/issuer-dashboard/ETOPresaleCounterWidget.module.scss
```diff
@@ -0,0 +1,31 @@
+.eto-public-start-date {
+  display: flex;
+  flex-flow: column;
+  justify-content: center;
+  align-items: center;
+  width: 100%;
```
desfero: Seems it's not needed

packages/web/app/components/issuer-dashboard/ETOPresaleCounterWidget.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <DashboardWidget
+      data-test-id="settings.presale-counter"
+      title={<FormattedMessage id="settings.presale-counter.title" />}
+      text={<FormattedMessage id="settings.presale-counter.text" />}
+      columnSpan={columnSpan}
+    >
+      <div className={styles.etoPublicStartDate}>
+        {timeLeft && <FancyTimeLeft finalTime={countdownDate} asUtc={true} refresh={true} />}
```
desfero: if we hide timer when there is no `timeLeft` then we also should hide/change text `Presale has started. It ends in:` otherwise, it looks strange`. Btw what's the exact case when there is no `timeLeft`?

packages/web/app/components/issuer-dashboard/ETOFundraisingStatistics.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const ETOFundraisingStatistics = compose<IProps, IExternalProps & IPanelProps>(
+  appConnect<IStateProps, {}>({
+    stateToProps: state => ({
+      etherPriceEur: selectEtherPriceEur(state),
+    }),
+  }),
+  withProps<IComputedProps, IProps & IStateProps>(props => ({
+    etherTokenEurEquivUlps: multiplyBigNumbers([
+      props.eto.contract!.totalInvestment.etherTokenBalance,
```
desfero: check for `isOnChain` and throw `InvalidEtoState` exception so then we don't have any casting here

packages/web/app/components/issuer-dashboard/ETOFundraisingStatistics.tsx
```diff
@@ -0,0 +1,136 @@
...
+      <section className={styles.groupWrapper}>
+        <span className={styles.label}>
+          <FormattedMessage id="settings.fundraising-statistics.total-investment" />
+        </span>
+        <Money
+          value={totalEquivEurUlps}
+          valueType={ECurrency.EUR}
+          inputFormat={ENumberInputFormat.ULPS}
+          outputFormat={ENumberOutputFormat.ONLY_NONZERO_DECIMALS}
+          currencySymbol={ECurrencySymbol.CODE}
```
desfero: do we need to specify `CODE` it's a default one (applies to all usages here)

packages/web/app/components/issuer-dashboard/ETOFundraisingStatistics.tsx
```diff
@@ -0,0 +1,136 @@
...
+    totalInvestors,
+  } = eto.contract.totalInvestment;
+
+  return (
+    <DashboardWidget
+      data-test-id="settings.fundraising-statistics"
+      title={<FormattedMessage id="settings.fundraising-statistics.title" />}
+      columnSpan={columnSpan}
+    >
+      <section className={styles.groupWrapper}>
```
desfero: image
IMO from the UI point of view, this one should be synced with whitelisting stats

packages/web/app/components/issuer-dashboard/ETOFundraisingStatistics.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+interface IExternalProps {
+  eto: TEtoWithCompanyAndContract;
+}
+
+interface IStateProps {
+  etherPriceEur: string;
+}
+
+interface IComputedProps {
```
desfero: hm, we need to probably agree on convention here as I see it's often called `IWithProps`

packages/web/app/components/issuer-dashboard/ETOFundraisingStatistics.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+  .addDecorator(
+    withStore({
+      tokenPrice: {
+        tokenPriceData: {
+          etherPriceEur: "168.71",
+        },
+      } as ITokenPriceState,
+    }),
+  )
+  .add("default", () => <ETOFundraisingStatistics eto={testEto} />);
```
desfero: Just use `ETOFundraisingStatisticsLayout` so we don't need to mock store in any way

rudolfix:
looks good however this state is visible only in PRE-SALE. in public sale we have empty dashboard again.

@drakkein @andywalton @pstroinski make sure there's ticket for PUBLIC sale as well

packages/web/app/components/issuer-dashboard/ETOPresaleCounterWidget.tsx
drakkein: Let's not hide it now

packages/web/app/components/issuer-dashboard/ETOFundraisingStatistics.tsx
drakkein: CODE is default, so we can remove it

packages/web/app/components/issuer-dashboard/ETOFundraisingStatistics.tsx
drakkein: I'm not sure what do you mean here

Issue #3593, opened 2019-10-09
Investment flow - invest step 1 quick fixes


andywalton:
### Background
Minor improvements to the investment modal step 1.

### Tasks
**1. Lets make it more clear that the what the ticket price is, and if it includes a discount, that we highlight that**

So, before the change we have this:

![platform neufund io_eto_view_DE_abf7e218-1a21-4423-a600-20504e40f3c2 (2)](https://user-images.githubusercontent.com/609657/66472716-ccb0ac00-eac8-11e9-90a0-90527290a67d.png)

After the change we have this:

![Invest with ETH changes to show token price + discount more clearly](https://user-images.githubusercontent.com/609657/66472740-d76b4100-eac8-11e9-90ac-859e9f7bb250.png)

Note this has some styling changes (mainly font sizing, positioning and some colours) - I will send you a separate hacked HTML file @desfero.

*Copy changes:*

Wallet selector changes to:
- [x] `Which balance would you like to invest from?`

Investment amount changes to:
- [x] `How much would you like to invest?`

The paragraph above the amount input field should read:
- [x] `You'll invest at a token price of xxx EUR which includes a x% discount.`

If no discount exists. The sentence should read:
- [x] `You'll invest at a token price of xxx EUR.`

Note both the token price and discount should be dynamic.

**2. Fix below token price unclear error message**

@desfero picked up this bug that we aren't highlighting clearly the token price correctly during the investment flow - it doesn't make sense particularly for ETH investment.

Screenshot showing the issue:
![image](https://user-images.githubusercontent.com/609657/66472486-4b591980-eac8-11e9-84e7-b1b94f800215.png)

A questions you might ask looking at that issue: Is that the ETH price or EUR. It's actually EUR?

- [x] Please update the error message firstly to be:
1. for ETH investing
`The amount is below the minimum ticket size of 100 000 EUR ≈ 607.1084 ETH`

2. for nEUR investing
`The amount is below the minimum ticket size of 100 000 EUR`

After this change it should look as follows:

![Invest with ETH changes to show token price + discount more clearly--with eth below ticket size error](https://user-images.githubusercontent.com/609657/66472839-0aadd000-eac9-11e9-859f-b56b500ec61b.png)

**3. Add some EUR conversion for ETH**

In two places:

- [x] 1. Amount input field
[...]

PR #3614, opened 2019-10-11
#3593 investment step quick fixes

desfero:
**Testing notes**
Fixes all from the #3593 except price per token and discount.


rudolfix:
- texting changes look good
- ETH/EUR conversions look good

PR #3635, opened 2019-10-15
Cleans up transactions sagas folder

Moejoe90:


packages/web/app/modules/tx/sender/sagas.ts
```diff
@@ -330,7 +330,27 @@ function* updateWalletValues(): Iterator<any> {
   yield put(actions.wallet.loadWalletData());
 }
 
+function* removePendingTransaction({ logger }: TGlobalDependencies): Iterator<any> {
```
desfero: IMO, this one should be moved under `tx/monitor/sagas` as it's related to the transaction `monitor` flow

Moejoe90: I was also thinking like you but then i realized we are actaully mainly touching the window sender hence why i choose to put it in sender. In any case after thinking about it again i will move it to monitor

Issue #3628, opened 2019-10-14
[ETO forms, Legal] - fix the Shareholder capital chart 


glupiawymowka:
**Description**
The chart was edited to show the % values from max to min. This is correct behavior, except the case of "others" that easily can be a higher value than any of the added positions. 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/99cf4263-151d-47ac-b14c-cb9ea157f4a5)

**Expected behavior** 
- [ ] No matter the % values, `Others` always stay at the last position
- [ ] all the other values are sorted as for now (max to min)

PR #3637, opened 2019-10-15
#3628 Always move Others to last position in shareholders chart

drakkein:


packages/web/app/components/eto/public-view/LegalInformationWidget.tsx
```diff
@@ -134,6 +134,8 @@ export const LegalInformationWidget: React.FunctionComponent<IProps> = ({
           )}
         </div>
 
+        {JSON.stringify(shareholdersData)}
```
desfero: some leftovers

packages/web/app/components/eto/utils.ts
```diff
@@ -7,6 +7,7 @@ import { formatFlexiPrecision } from "../../utils/NumberUtils";
 import { TShareholder } from "./public-view/LegalInformationWidget";
 
 const HUNDRED_PERCENT = new BigNumber(100);
+const OTHERS_NAME = "Others";
```
desfero: Can we translate `Others`? We can do it on the component level by checking if `name === OTHERS_NAME`

Issue #3627, opened 2019-10-14
Ledger login issue


pstroinski:
**Description**
Since the latest Windows update we are unable to use Ledger Nano on our platform. We have a constant popup that cannot be removed.

**Environment**
Production

**Steps To Reproduce**
Connect the ledger and open the login page with the Ledger Tab

**Actual behaviour**
Constant pop-up that blocks any actions on a browser

**Expected behavior**
Possibility to log into the platform and invest on Windows with ledger

**Comments, context, logs**

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5beab1a20cefa317ea424319/7a927872-ec71-4db4-a7f0-22334c5133c5)

pstroinski:
Related to https://github.com/Neufund/platform-frontend/pull/3600

Issue #2819, opened 2019-04-29
OnlyAuthorizedRoute component will always redirect to Wallet Selector during logout


Moejoe90:
**Description**
Currently `OnlyAuthorizedRoute` relies on the fact the user is verified to show correct component or redirect to login. That is ok except for the fact when you need to log out on purpose. 

In the case of manual logout, the component will quickly redirect to the wallet selector then to the main page. 
**Environment**
All
**Steps To Reproduce**
Try to verify user with wrong link using Metamask
**Actual behaviour**
You will have a split second where the router is on the wallet selector only noticeable with `console.log`
**Expected behaviour**
App should redirect to root without passing through the wallet selector
**Comments, context, logs**

**Screenshots**

PR #3638, opened 2019-10-16
#3627 Reduce the amount of u2f calls for ledger

desfero:
**Technical notes**
It's not possible to eliminate popup on windows in any way as each `send` command triggers the modal. What we can do is to reduce the number of calls and remove all kinds of polling so it's in some way usable on windows.
I'm still not fully happy about the code there as we can further refactor `wallet-selector` and `access-modal` modules to be more in sync but for now, let's push what we have. 

**Testing notes**
This one will be pretty time consuming to test as we need to go through login/register flow for light/ledger/metamask and also do the same for signing the message (for e.g. changing email) and sending transactions (for e.g.investment) flow.

From important changes:
- [ ] There is no pooling for a connection right now so in case there is some communication error `TRY AGAIN` button will appear to retry the connection. The same button should appear for login/register with metamask and ledger and also while signing a transaction in case metamask was disconnected it should not retry but rather show try again button.
image
- [ ] It should not be possible to log in again in case the user moved from the login page to anywhere else and then accepted a signing message either on metamask or ledger.
This is how it is right now:
Oct-21-2019 10-43-44
- [ ] There should not be any flash of a white page while transitioning from unauthorized to authorized state. This was the bug we introduced in #3230 and now it should be fixed.
This is how it is right now:
Oct-21-2019 10-47-27


packages/web/app/lib/web3/Web3Manager/Web3Manager.ts
```diff
@@ -79,22 +78,15 @@ export class Web3Manager extends EventEmitter {
...
-
     this.emit(EWeb3ManagerEvents.NEW_PERSONAL_WALLET_PLUGGED, {
-      isUnlocked,
+      isUnlocked: this.personalWallet.isUnlocked(),
       metaData: this.personalWallet.getMetadata(),
     });
 
-    this.web3ConnectionWatcher.start();
+    // TODO: discuss code on review
+    // this.web3ConnectionWatcher.start();
```
desfero: @rudolfix @Moejoe90 I really don't know what do to with connection watcher we have here. I don't see a reason to watch for a connection as right now we don't reuse it in any way (during tx flow wallet is plugged again). And if we want to add support for windows we should get rid of connection watching as otherwise modal is triggered every second 5 seconds.

desfero: i'm sure we compare address, let me check recheck with `networkId`

packages/web/app/modules/web3/sagas.ts
```diff
@@ -63,14 +63,14 @@ export function* cancelLocking(): Iterator<any> {
...
   let storageData = walletStorage.get();
+
   if (storageData) {
     yield put(actions.web3.loadPreviousWallet(storageData));
   }
 }
 
 export function* personalWalletConnectionLost({ notificationCenter }: TGlobalDependencies): any {
   yield put(actions.walletSelector.reset());
-  yield put(actions.walletSelector.ledgerReset());
```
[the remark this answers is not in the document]
desfero: yes, if we remove the watcher then yes, this code gonna be removed, just wanted to keep it until we confirm that watcher is safe to remove

packages/web/app/components/wallet-selector/WalletSelectorLayout.tsx
```diff
@@ -0,0 +1,188 @@
+import * as cn from "classnames";
```
desfero: There is no need for a full review here, it's just copy-paste from `WalletSelector`

packages/web/app/lib/web3/Web3Manager/Web3Manager.ts
desfero: ok, we compare `networkId` too before connecting

rudolfix:
my review:
1. metamask works correctly I didn't check other wallets like opera or gnosis but they should not be affected
2. nano ledger on linux works correctly
3. nano ledger on windws is usable but shows a lot of popups

my take on what to do:
1. we can safely remove is unlocked check - we have long timeout so there's plenty of time to unlock. if not we have retry button now
2. we merge what we have. Ledger is suggesting to move to webHID transport so we just going to wait and switch when it's ready
3. for browser wallets let's implement this: https://eips.ethereum.org/EIPS/eip-1193 so we can detect account/network change and log out user immediately. @desfero could you add the ticket?

packages/web/app/lib/web3/Web3Manager/Web3Manager.spec.ts
```diff
@@ -72,128 +48,5 @@ describe("Web3Manager", () => {
       },
       isUnlocked: true,
     });
-    expect(disconnectListener).to.not.be.called;
-    expect(asyncIntervalSchedulerMock.start).to.be.calledOnce;
-  });
-
-  it("should throw when plugging not connected wallet", async () => {
```
Moejoe90: If only u2f and windows worked without any hassle

packages/web/app/lib/web3/Web3Manager/Web3Manager.ts
```diff
@@ -5,14 +5,8 @@ import * as Web3 from "web3";
 
 import { symbols } from "../../../di/symbols";
 import { calculateGasLimitWithOverhead, encodeTransaction } from "../../../modules/tx/utils";
-import {
-  AsyncIntervalScheduler,
```
Moejoe90: maybe we should get rid fully off `AsyncIntervalScheduler` its only used here and `withActionWatcher`. 
I think `withActionWatcher` can be sagatized what do you think

packages/web/app/lib/web3/ledger-wallet/LedgerConnector.ts
```diff
@@ -42,9 +42,8 @@ export class LedgerWalletConnector {
     derivationPath: string,
     networkId: EthereumNetworkId,
   ): Promise<LedgerWallet> {
-    if (!this.getTransport) {
-      throw new Error("Can't finish not started connection");
-    }
+    if (!this.getTransport) throw new LedgerNotAvailableError();
```
Moejoe90: :+1:

packages/web/app/lib/web3/ledger-wallet/LedgerConnector.ts
```diff
@@ -72,42 +73,35 @@ export class LedgerWalletConnector {
...
   public async getMultipleAccounts(derivationPaths: string[]): Promise<IDerivationPathToAddress> {
-    const accounts: IDerivationPathToAddress = {};
-
-    for (const derivationPath of derivationPaths) {
-      if (!this.getTransport) throw new LedgerNotAvailableError();
-
-      const account = await this.getMultipleAccountsFromHdPath(derivationPath, 0, 1);
-      Object.assign(accounts, account);
-    }
+    const accounts = await Promise.all(
```
Moejoe90: :1st_place_medal:  
I wonder if this optimizes the time the ledger needs to process all this information

packages/web/app/lib/web3/ledger-wallet/LedgerWallet.ts
```diff
@@ -92,4 +92,8 @@ export class LedgerWallet implements IPersonalWallet {
       this.waitingForCommand = false;
     }
   }
+
+  public isUnlocked(): boolean {
```
Moejoe90: Now that i see `isUnlocked` for the third time i realize if `isUnlocked` is the correct name. 
As we use AFAIK to assume if the user needs to enter a password or not and not really check if its unlocked. 
maybe we change the naming?

packages/web/app/lib/web3/ledger-wallet/ledgerUtils.ts
```diff
@@ -52,18 +52,6 @@ export const testConnection = async (getTransport: () => any): Promise<boolean>
   }
 };
 
-export const testIfUnlocked = async (getTransport: () => any): Promise<void> => {
```
Moejoe90: I wonder if we should really remove it vs just not use it. 

As this function still works. In all cases i have no strong feelings about this

packages/web/app/modules/access-wallet/sagas.ts
```diff
@@ -159,34 +161,27 @@ export function* accessWalletAndRunEffect(
...
-}
+  // always hide the current modal
+  yield put(actions.accessWallet.hideAccessWalletModal());
 
-/**
- * Use only as a part of another saga. This won't trigger modal.
- */
-export function* connectWallet(): any {
-  yield connectWalletAndRunEffect(call(() => {}));
+  return yield effect;
```
Moejoe90: Now that i see this change i think it makes sense that run the effect actually after hiding the access modal

packages/web/app/modules/wallet-selector/browser-wizard/sagas.ts
```diff
@@ -59,5 +40,10 @@ export function* tryConnectingWithBrowserWallet({
 }
 
 export function* browserWalletSagas(): Iterator<any> {
-  yield fork(neuTakeEvery, "BROWSER_WALLET_TRY_CONNECTING", browserWalletConnectionWatcher);
+  yield fork(
+    neuTakeUntil,
```
Moejoe90: what happens if you use `neuTakeLatestUntil`?

packages/web/app/test-e2e/link-triggered-flows/WalletVerify.spec.e2e.io.ts
```diff
@@ -58,8 +58,6 @@ describe("Verify Wallet", () => {
       // try to activate previous user when second one is logged in
       cy.visit(activationLink);
 
-      cy.get(tid("light-wallet-login-with-email-email-field")).contains(email);
```
Moejoe90: why remove this?

packages/web/app/test-e2e/kyc/KYCFlowIndividualManual.spec.e2e.io.ts
```diff
@@ -9,6 +9,7 @@ import { tid } from "../utils/selectors";
 import { createAndLoginNewUser } from "../utils/userHelpers";
 import { kycInvidualForm } from "./fixtures";
 
+// Unskip after US investor is done
```
Moejoe90: is this still relavent?

packages/web/app/test-e2e/kyc/KYCFlowIndividualIDnow.spec.e2e.io.ts
```diff
@@ -21,6 +21,7 @@ const initiateIDNowKyc = () => {
   cy.get(tid("kyc-panel-outsourced")).should("exist");
 };
 
+// Unskip after US investor is done
```
Moejoe90: is this still relavent?

packages/web/app/test-e2e/kyc/KYCFlowIndividualOutsourced.spec.e2e.io.ts
```diff
@@ -4,6 +4,7 @@ import { tid } from "../utils/selectors";
 import { createAndLoginNewUser } from "../utils/userHelpers";
 import { kycInvidualForm } from "./fixtures";
 
+// Unskip after US investor is done
```
Moejoe90: is this still relavent?

packages/web/app/modules/web3/sagas.ts
```diff
@@ -62,15 +62,15 @@ export function* cancelLocking(): Iterator<any> {
...
 export function* loadPreviousWallet({ walletStorage }: TGlobalDependencies): Iterator<any> {
-  let storageData = walletStorage.get();
+  const storageData = walletStorage.get();
+
   if (storageData) {
     yield put(actions.web3.loadPreviousWallet(storageData));
   }
 }
 
 export function* personalWalletConnectionLost({ notificationCenter }: TGlobalDependencies): any {
```
Moejoe90: do we remove this now?

packages/web/app/modules/web3/sagas.ts
```diff
@@ -93,39 +93,30 @@ export function* personalWalletConnectionLost({ notificationCenter }: TGlobalDep
   }
 }
 
-interface IChannelTypes {
-  type: EWeb3ManagerEvents;
-  payload?: { isUnlocked: boolean; metaData: TWalletMetadata };
-}
+type TChannelTypes = {
+  type: EWeb3ManagerEvents.NEW_PERSONAL_WALLET_PLUGGED;
```
Moejoe90: :+1:

Moejoe90:
Do you also confirm that this is solved with your pr?
`https://github.com/Neufund/platform-frontend/issues/2819`

packages/web/app/components/modals/wallet-access/AccessWalletModal.tsx
```diff
@@ -84,7 +89,7 @@ export const AccessWalletContainerComponent: React.FunctionComponent<
     )}
     {walletType === EWalletType.LEDGER && (
       <div>
-        <img src={ledgerConfirm} className="mt-1 mb-3" />
+        <img src={ledgerConfirm} className="mt-1 mb-3" alt="" />
```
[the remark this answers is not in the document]
desfero: yes, it's a proper way of adding images when you don't need any description, otherwise screen reader will read the file name

packages/web/app/lib/web3/Web3Manager/Web3Manager.ts
desfero: yeah, if we can then it's would be good to remove

packages/web/app/lib/web3/ledger-wallet/LedgerWallet.ts
desfero: agree, the name is confusing, I just used the name we had for the variable name for consistency

packages/web/app/test-e2e/link-triggered-flows/WalletVerify.spec.e2e.io.ts
desfero: it's a duplicate, see two lines below

packages/web/app/test-e2e/kyc/KYCFlowIndividualManual.spec.e2e.io.ts
desfero: ah, sorry some leftover will remove

packages/web/app/components/wallet-selector/WalletSelector.integration.spec.tsx
```diff
@@ -307,7 +313,12 @@ describe("Wallet selector integration", () => {
     await waitForTid(mountedComponent, "browser-wallet-error-msg");
 
     // there is no wallet in browser (BrowserWallet thrown BrowserWalletMissingError)
-    expect(mountedComponent.find(tid("browser-wallet-error-msg")).text()).to.be.eq(
+    expect(
+      mountedComponent
+        .find(tid("browser-wallet-error-msg"))
+        .first()
```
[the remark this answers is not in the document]
desfero: It's because if you pass `data-test-id` to a component and the component under the hood pass `data-test-id` to the dom you have inside enzyme two dom nodes if you search for `data-test-id`. Then if `enzyme` sees two nodes and you are trying to get the `text()` it will throw an error. I think we used that approach previously (to select `first()`) so I just tried to keep consistency

packages/web/app/components/wallet-selector/browser/WalletBrowser.tsx
```diff
@@ -94,13 +107,15 @@ export const WalletBrowser = compose<React.FunctionComponent>(
...
-        dispatch(actions.walletSelector.browserWalletResetApprovalRequest());
+      tryConnectingWithBrowserWallet: () => {
         dispatch(actions.walletSelector.tryConnectingWithBrowserWallet());
       },
     }),
   }),
   onEnterAction({
-    actionCreator: dispatch => dispatch(actions.walletSelector.tryConnectingWithBrowserWallet()),
+    actionCreator: dispatch => {
+      dispatch(actions.walletSelector.reset());
```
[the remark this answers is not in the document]
desfero: This one is just moving `reset` action from the `WalletSelector` to each route separately to reset all previous pending signings otherwise if you go to the metamask and then to light wallet code related to metamask signign is still not canceled.

packages/web/app/components/wallet-selector/light/WalletLight.tsx
```diff
@@ -17,8 +20,15 @@ export const WalletLightComponent: React.FunctionComponent<IStateProps> = ({ isL
   </section>
 );
 
-export const WalletLight = appConnect<IStateProps>({
-  stateToProps: s => ({
-    isLoginRoute: selectIsLoginRoute(s.router),
+export const WalletLight = compose<IStateProps, {}>(
+  onEnterAction({
```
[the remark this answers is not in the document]
desfero: This one is a good idea exactly, will do

packages/web/app/modules/wallet-selector/browser-wizard/sagas.ts
desfero: hm, we can use `neuTakeLatestUntil` but I just tried to keep consistency with `neuTakeEvery`

packages/web/app/modules/web3/sagas.ts
desfero: good catch, we can remove it's not used anymore

Issue #3624, opened 2019-10-14
Add token icon and symbol to investment summary modal


desfero:
**Background**
Right now there is a placeholder token image inside investment summary:
![image](https://user-images.githubusercontent.com/7031578/66738564-1e618980-ee6f-11e9-8690-65132aefb88e.png)
We need to replace it with a proper token icon with a symbol.

**Designs**
not needed

**Acceptance criteria**
- [ ] Token icon and a token symbol should be displayed inside `your equity tokens` summary, success and rejected modals

**Technical notes**

**Testing notes**
Please note that you can access success/rejected modals also from pending transaction monitor (button in header)

PR #3644, opened 2019-10-17
#3624 add token and symbol to investment summary

desfero:
**Testing notes**
In case there is a pending transaction created by an old frontend ui (so the one without token symbol and image) we are gonna delete this transaction from the pending list and throw a warning in the console. You can test it by creating an investment transaction on `io` and closing the page before reaching successful state (so you have a pending transaction icon in the header) then on codechecks just refresh the page and you should see a warning in the console and no pending transaction in the header.


packages/web/app/modules/tx/monitor/sagas.ts
```diff
@@ -99,22 +100,38 @@ export function* markTransactionAsPending(
   return transactionTimestamp;
 }
 
-export function* makeSureWithdrawSchemaIsUpToDate(
-  pendingTransaction: TxPendingWithMetadata | undefined,
+export function* ensumrePendingTransactionSchemaIsValid(
+  pendingTransaction: TxPendingWithMetadata,
 ): Iterator<any> {
   // THIS IS A TEMPORARY PATCH A GENERAL SOLUTION THAT INCLUDES VERSIONING SHOULD COVER ALL TX TYPES
```
desfero: I'm not sure we really need versioning for schema mismatch, as it's a really rare case and on top of that the amount of pending transactions is really small (close to zero) so it will just complicate frontend with unneeded `if`'s.

packages/web/app/utils/opaqueTypesUtils.ts
```diff
@@ -0,0 +1,5 @@
+import { EquityToken } from "../types";
+
+const toEquityTokenSymbol = (symbol: string) => symbol as EquityToken;
```
desfero: let's hide all `as *` for our opaque types under some functions so it's easier to work with

packages/web/app/utils/opaqueTypesUtils.ts
```diff
@@ -0,0 +1,5 @@
+import { EquityToken } from "../types";
```
Moejoe90: Maybe we should put it in a specific folder called Opaque to make it easier to manage these Opaques in one place

packages/web/app/types.d.ts
```diff
@@ -26,6 +26,7 @@ export type EthereumTxHash = Opaque<"EthereumTxHash", string>;
 export type EthereumAddress = Opaque<"EthereumAddress", string>;
 export type EthereumAddressWithChecksum = Opaque<"EthereumAddressWithChecksum", string>;
 export type FunctionWithDeps = Opaque<"FunctionWithDeps", Function>;
+export type EquityToken = Opaque<"EquityToken", string>;
```
Moejoe90: I would also put all these types into an Opaque folder as they will get easier to manage

packages/web/app/components/modals/tx-sender/investment-flow/InvestmentTransactionDetails.tsx
```diff
@@ -165,9 +169,7 @@ const InvestmentTransactionDetails: TransactionDetailsComponent<ETxSenderType.IN
     additionalData.equityTokens,
   );
 
-  const fullTokenPrice =
-    additionalData.eto.investmentCalculatedValues.sharePrice! /
-    additionalData.eto.equityTokensPerShare;
+  const fullTokenPrice = additionalData.eto.sharePrice / additionalData.eto.equityTokensPerShare;
```
Moejoe90: is this operation done with `floats`? 
This would be an issue then we need to use decimal operations here to not lose precision

packages/web/app/components/modals/tx-sender/investment-flow/Success.stories.tsx
```diff
@@ -18,8 +21,11 @@ const props = {
       preMoneyValuationEur: 0,
       existingShareCapital: 0,
       equityTokensPerShare: 0,
-      investmentCalculatedValues: {
-        sharePrice: 0,
+      sharePrice: 0,
```
Moejoe90: IMO we should make `equityTokensPerShare` , `sharePrice` with some value and better some complex value to see how these values get calculated in the end vs using `0` which will result in `0` or even infinity since we are maybe deviding by 0

packages/web/app/modules/tx/monitor/sagas.ts
Moejoe90: We need it everytime we decide to change transactions `additional data`.  
That's why need a migration stretegy while i agree its not top priority but situatiosn like this will come up everytime. 
And the more usres we have the bigger the chance is we will have out of date transactions

For sure it shouldn't look like how it is now though

packages/web/app/modules/tx/monitor/sagas.ts
```diff
@@ -174,13 +191,24 @@ export function* updatePendingTxs({
...
     if (e instanceof SchemaMismatchError) {
-      logger.warn(new Error("Found Withdraw Schema Mismatch"));
-      yield apiUserService.deletePendingTx(apiPendingTx.pendingTransaction!.transaction.hash);
+      logger.warn("Found pending transaction Schema Mismatch", e);
+
+      if (apiPendingTx.pendingTransaction) {
+        yield apiUserService.deletePendingTx(apiPendingTx.pendingTransaction.transaction.hash);
+      }
+    } else {
+      // rethrow error
```
Moejoe90: do we really need this comment?

packages/web/app/modules/tx/transactions/investment/types.ts
```diff
@@ -1,19 +1,25 @@
...
-  };
-  equityTokens: string;
-  estimatedReward: string;
-  etherPriceEur: string;
-  gasCostEth: string;
-  investmentEth: string;
-  investmentEur: string;
-  isIcbm: boolean;
-};
+const InvestmentAdditionalDataEtoSchema = YupTS.object({
```
Moejoe90: I like the idea of defiing a schema vs this hardcoded conditions up

packages/web/app/modules/tx/monitor/sagas.ts
desfero: hm, agree, the code is quite self-explanatory :D

packages/web/app/modules/tx/transactions/investment/types.ts
desfero: still we need to solve in some way versioning but honestly, I don't think that will happen anytime soon

packages/web/app/components/modals/tx-sender/investment-flow/InvestmentTransactionDetails.tsx
desfero: done

packages/web/app/types.d.ts
desfero: done

packages/web/app/utils/opaqueTypesUtils.ts
desfero: done

PR #3665, opened 2019-10-23
Fix mobile metamask signing flow

desfero:
**Testing notes**
Please update metamask mobile to the latest beta version as there is also some bug with signing transactions with some previous beta.
There is no need for a full regression testing of wallets as wallet related code is not affected by this PR


packages/web/app/utils/PromiseUtils.ts
```diff
@@ -4,7 +4,7 @@ export const promisify = <T>(fn: Function) =>
   function(...args: any[]): Promise<T> {
     return new Promise<T>((resolve, reject) => {
       fn(...args, (err: any, result: T) => {
-        if (err !== null) {
+        if (err !== null && err !== undefined) {
```
desfero: For some reason, metamask mobile is not in sync with a callback approach in node js where `error` is always `null` or some `Error` object.

Moejoe90: Should we keep a comment on the code for this. As someone else checking this will think we made a mistake

PR #3666, opened 2019-10-23
Refactors TX watchers

Moejoe90:
This refactors the txWatchers to sagas and gets rid of event channel in order to use sagas completely. 

This will give us the chance to use API calls


packages/web/app/modules/tx/event-channel/sagas.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+export function* getTransactionOrThrow(
+  { web3Manager }: TGlobalDependencies,
+  txHash: string,
+): Iterator<any> {
+  const tx: Web3.Transaction = yield web3Manager.getTransactionByHash(txHash);
+  const txReceipt: Web3.TransactionReceipt | null = yield web3Manager.internalWeb3Adapter.getTransactionReceipt(
+    txHash,
+  );
```
drakkein: Maybe we should use `all` here as a part of improvements? 
```const { tx, txReceipt} = all({...})```

File: packages/web/app/modules/tx/event-channel/sagas.ts [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+
+  // Both requests `getTx` and `getTransactionReceipt` can end up in two seperate nodes
+  const isMined = tx && tx.blockNumber && txReceipt && txReceipt.blockNumber;
+
+  if (!isMined) {
+    return null;
+  }
+
+  if (txReceipt!.status === TRANSACTION_STATUS.REVERTED) {
+    if (txReceipt!.gasUsed === tx.gas) {
```
drakkein: Do we need these non null assertions here? Isn't it handled by isMined? Probably ts gonna have problem resolving it, so it's not important.

packages/web/app/modules/tx/event-channel/types.ts
```diff
@@ -15,17 +13,16 @@ export type TEventEmitterChannelEvents =
     }
   | {
       type: EEventEmitterChannelEvents.TX_MINED;
-      tx: Web3.Transaction;
     }
   | {
       type: EEventEmitterChannelEvents.ERROR;
-      error: any;
+      error: Error;
```
drakkein: <3

packages/web/app/modules/tx/monitor/sagas.ts
```diff
@@ -135,11 +135,7 @@ export function* ensumrePendingTransactionSchemaIsValid(
   }
 }
 
-export function* updatePendingTxs({
-  apiUserService,
-  web3Manager,
-  logger,
-}: TGlobalDependencies): any {
+export function* updatePendingTxs({ apiUserService, logger }: TGlobalDependencies): any {
```
drakkein: Iterator<any>, also in line 209

Moejoe90: done

packages/web/app/modules/tx/event-channel/sagas.ts
Moejoe90: The TSC complains here if i remove the non null assertions, I think there is still some work to be done there. So i will leave them for now

AlexeyTuboltsev: btw if you'd write it like this:
```
if (!(tx && tx.blockNumber && txReceipt && txReceipt.blockNumber)) {
    return null;
  } else if (txReceipt.status === TRANSACTION_STATUS.REVERTED) {
    if (txReceipt.gasUsed === tx.gas) {
      // All gas is burned in this case
      throw new OutOfGasError();
    }
    throw new RevertedTransactionError();
  } else {
    return tx;
  }
```
there won't be any problems with TSC (and it makes readability a bit better, but this is a matter of taste)

AlexeyTuboltsev: btw if you'd write it like this:
```
if (!(tx && tx.blockNumber && txReceipt && txReceipt.blockNumber)) {
    return null;
  } else if (txReceipt.status === TRANSACTION_STATUS.REVERTED) {
    if (txReceipt.gasUsed === tx.gas) {
      // All gas is burned in this case
      throw new OutOfGasError();
    }
    throw new RevertedTransactionError();
  } else {
    return tx;
  }
```
there won't be any problems with TSC (and it makes readability a bit better, but this is a matter of taste)

AlexeyTuboltsev: ok, let's leave it like this to preserve the `isMined` constant for readability

PR #3682, opened 2019-10-28
Fixes pledged amount

Moejoe90:


packages/web/app/components/eto/overview/EtoOverviewStatus/CampaigningWidget/RegisterNowWidget.tsx
```diff
@@ -31,7 +31,7 @@ const RegisterNowWidget: React.FunctionComponent<TExternalProps> = ({
               values={{
                 totalAmount: (
                   <Money
-                    value={pledgedAmount.toString()}
+                    value={pledgedAmount ? pledgedAmount.toString() : undefined}
```
desfero: hm, but there is still a problem here as someone may in future remove ternary as by types `pledgedAmount` is always number. Can we fix types for `pledgedAmount` (`TComponentProps` in `connectCampaigningActivatedWidget`, maybe we can just union `IStateProps` with`TExternalProps` as then we have consistency and full type checking)

PR #3684, opened 2019-10-29
Show full target amount

desfero:
**Testing notes**
Fixes https://neufund.slack.com/archives/GAK8XK0CB/p1572259303003400


packages/web/app/components/shared/formatters/MoneyRange.tsx
```diff
@@ -60,9 +61,8 @@ export const MoneyRange: React.FunctionComponent<
         />
       );
     } else if (
-      Object.values(ENumberOutputFormat).includes(outputFormat) &&
```
desfero: I don't see any reason to keep `Object.values...includes` as we then do the same filtering below 🤔?

Issue #3663, opened 2019-10-23
Public sale soft cap changes


andywalton:
### Background
This is related to the #3662 Soft Cap epic, specifically changes needed for the Public sale.

### Designs
See designs in Zeplin here: [https://zpl.io/V495jPz](https://zpl.io/V495jPz)

![003b - Public sale company listing page - soft cap reached](https://user-images.githubusercontent.com/609657/67398532-1e6a3380-f5ab-11e9-87aa-4b842cce3714.png)

### Tasks
Please review each image in Zeplin and note the annotation for a detailed breakdown of the tasks. 

Here's a simple list of tasks:

- [x] **000 - Background colour change for the app**
Please change the grey background of the class `Layout-module__layout___vcQPe` from `#F3F5F7` to `#F9FAFB`

- [x] **001 - Public sale ETO card - soft cap not reached**
- [x] **002 - Public sale ETO card - soft cap reached**

1.  Consolidate the funds raised to show the equivalent in EUR and stop showing it separately as ETH and nEUR.
2. Change the colour of the soft cap vertical line
3. Conditionally change the label if the soft cap is reached
4. Update the footer CTA as follows: 

```
When there are many days: ENDS IN 27 DAYS 
When it gets down to 1 day say: ENDS TOMORROW AT 12:12:12 UTC
When it’s today, it should say. ENDS TODAY AT 12:12:12 UTC
```

- [x] **003a - Public sale company listing page - Soft cap not reached**
- [x] **003b - Public sale company listing page - soft cap reached**
- [x] **003c - Public sale company listing page - soft cap reached (tooltip)**

1. Changes to the INVESTMENT AMOUNT to show the TARGET INVESTMENT AMOUNT and only show the max amount. ie. remove the soft cap amount.
2. Update the campaign metrics graph
3. Conditionally show a Fundraising soft cap message with a call to action that toggles the Fundraising Statistics.

- [x] **004 - Minor copy change to ETO Investment Terms**
Add the copy (Excl. discounts) to the Investment Amount field under ETO Investment Terms.

- [x] **005 - Portfolio view**
Conditionally show that the soft cap has been reached by changing the label of the bar graph. 

### Acceptance criteria
As per the Tasks.

### Technical notes
N/a

### Testing notes
As per the Tasks and Zeplin annotations.

PR #3688, opened 2019-10-30
#3663 Public soft cap changes - part 2

desfero:
**Testing notes**
`View details` inside `Fundraising statistics` tooltip is not implemented as before we need to update reacstrap to the latest. I talked with @andywalton and it's ok for now to go without `view details` there.


packages/web/app/modules/eto/utils.ts
```diff
@@ -178,3 +192,24 @@ export const getEtoSubState = ({
     }
   }
 };
+
+export const getInvestmentCalculatedPercentage = (eto: TEtoSpecsData) =>
```
Moejoe90: I know this was just moved but i think we really need to make these calculations with BigNumbers
So this is non-blocking 
https://github.com/Neufund/platform-frontend/issues/3695

packages/web/app/modules/eto/utils.ts
```diff
@@ -178,3 +192,24 @@ export const getEtoSubState = ({
...
+  (eto.newSharesToIssue / eto.minimumNewSharesToIssue) * 100;
+
+export const getCurrentInvestmentProgressPercentage = (eto: TEtoWithCompanyAndContract) => {
+  const totalTokensInt = eto.contract!.totalInvestment.totalTokensInt;
+
+  return (
+    (parseInt(totalTokensInt, 10) / (eto.minimumNewSharesToIssue * eto.equityTokensPerShare)) * 100
```
Moejoe90: `parseInt` :scream:

Issue #3520, opened 2019-09-24
Add ability to block investing in an ETO


andywalton:
**Background**
With us working towards allowing US Accredited Investors to invest in ETOs on the platform, we need to provide the ability to block ETOs that don't allow US Accredited Investors.

Therefore we require a new state on the company detail view that hides the Invest button and replaces this with a message that tells the Investor that this investment opportunity isn't available for their region etc.

This is the widget where we'll need to place this message:
<img width="1205" alt="Screenshot 2019-09-24 17 17 17" src="https://user-images.githubusercontent.com/609657/65525135-52403380-deef-11e9-8d5d-479c45bd4fa3.png">

**Designs**
The main place to block investing is on the ETO company page. We can do something like the below:
<img width="1260" alt="Screenshot 2019-10-08 00 20 02" src="https://user-images.githubusercontent.com/609657/66747325-87073100-ee84-11e9-9d1a-300a36c29a85.png">

However, the copy and call to actions in the ETO summary widget should change based on two possible scenarios:

1. A US investor is told they can't invest in an ETO offering because the Issuer does not accept Accredited US Investors.

```
Unfortunately, this offer doesn't allow investors from the US to participate. Please browse other investment offers.

GO TO DASHBOARD
```

2. A US investors accreditation has expired, we need to tell them to manage this and reverify their accreditation status. The call to action  call to action button should change to the following:

```
Your US investor accreditation has expired. Please provide your latest accreditation documentation to unlock investing.

MANAGE ACCREDITATION
```

@vizmab - please will you review the copy above.

**Acceptance criteria**
- [ ] An investor is told they can't invest in an ETO offering based on Issuer not accepting Accredited US Investors.
- [ ] The opposite of this is the case where the Issuer allows Accredited US Investors, we should show the relevant CTA available at the stage of the ETO for that investor.
- [ ] We block US accredited investors from investing when their accreditation has expired.

**Technical notes**
Backend will provide the relevant API for this.

**Testing notes**
Same as acceptance, test both the true and false use cases of the above. #

desfero:
Partialy completed in #3694. For now, moving back to the backlog.

Moejoe90:
For now we will close the ticekt as the investing is blocked in case there are new requirements please create a new ticket

PR #3694, opened 2019-10-31
#3520 Block investing for US investors

desfero:
**Testing notes**
It's a partial implementation of #3520 which only blocks investing for all US investors. For now, #3520 is moved back to backlog until we finalize requirements


packages/web/app/components/eto/overview/EtoOverviewStatus/Message.tsx
```diff
@@ -15,14 +15,14 @@ const Message: React.FunctionComponent<IExternalProps & TDataTestId> = ({
...
   title,
   summary,
+  children,
   "data-test-id": dataTestId,
 }) => (
   <section className={styles.widget} data-test-id={dataTestId}>
     {tick}
-    <header className={styles.message}>
-      {title && <h5>{title}</h5>}
-      {summary && <p>{summary}</p>}
```
desfero: this one is not quite valid as we often put some additional `p` tags inside `summary` and that's not a valid HTML (aka `p` in `p`)

Issue #3681, opened 2019-10-28
[ETO widget] - founders quote not loading 


glupiawymowka:
**Description**
ETO in draft/published/coming soon state should show the founders quote. 
Currently, it's showing the forever loading indicator. 
Also, it shows a broken state for the unregistered investor. 

**Env**
dev, stage, most probably also prod

**Steps to reproduce**
1. Login with ISSUER_PREVIEW
2. Open the ETO Page tab

ISSUER view
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/cac04c38-26bf-4d37-abc0-e073b9b35d5a)

INVESTOR view 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/b719b61c-22c9-434d-b799-3c0e37992862)

3. You can also check Planti as the unregistered investor (nothing to be committed here yet!) 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/c88b751d-4465-443e-a368-37aff0acd9f9)

**Actual behavior** 
- loading indicator instead of founders quote
- broken NAN commitments appear

**Expected behavior** 
ISSUER flow : 
- [x] founders quote visible for Draft, Pending and Campaigning (no whitelist) states

INVESTOR flow:
- [ ] `NAN INVESTORS COMMITED -` removed for unregistered investors, for the ETO in Coming soon/Campaigning (without whitelist) states
- [ ]  `Interested in taking part in this offering - Register` for unregistered investors
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/1f306cd8-2082-4d07-b8fc-df053cffe85d)
- [x] Founders quote visible in Coming soon/Campaigning (no whitelist) state, for registered investors

glupiawymowka:
state for unregistered users can be verified after merge

PR #3697, opened 2019-10-31
#3681 Fix founders quote not loading

drakkein:


packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -73,18 +74,24 @@ export const calculateWhitelistingState = ({
   bookbuildingLimitReached,
   investorsCount,
   investmentCalculatedValues,
+  isAuthorized,
 }: TcalculateWhitelistingState) => {
   //TODO this should be put in line with api after API's canEnableBookbuilding after limit is reached gets fixed by Marcin
   if (
-    investorsCount === undefined ||
-    (investmentCalculatedValues === undefined && whitelistingIsActive)
+    (isAuthorized && investorsCount === undefined && whitelistingIsActive) ||
```
Moejoe90: since its almost the same condition why not use 
`if(isAuthorized && whitelistingIsActive && (investorsCount === undefined )||investmentCalculatedValues === undefined)`

PR #3704, opened 2019-11-05
Temp Fix for KYC US form + Cleanup

Moejoe90:


packages/web/app/components/kyc/personal/Start.unsafe.tsx
```diff
@@ -226,15 +243,20 @@ const KYCForm = injectIntlHelpers<IProps & IKycIndividualData & IDispatchProps>(
       <div className="p-4 text-center">
         <Button
           type="submit"
-          disabled={!props.isValid || props.loadingData}
+          disabled={
+            uploadedFilesLoading ||
```
Moejoe90: This is not the right way to do this. 
The upload component should be connected with Formik and has its own validation schema etc. 
This is done to as a quick fix

PR #3724, opened 2019-11-12
#3404 Fix neur fee calculation

drakkein:
Old PR for reference https://github.com/Neufund/platform-frontend/pull/3427

Changed way of performing calculations according to @rudolfix directions


Moejoe90:
will close this and we will rework the solution

PR #3730, opened 2019-11-13
Block KYC by region

desfero:


packages/web/app/modules/kyc/utils.ts
```diff
@@ -48,3 +49,29 @@ export const kycStatusToTranslationMessage = (status: EKycRequestStatus): TMessa
...
   }
 };
+
+export function appendIfExists<T>(array: ReadonlyArray<T>, item: T | undefined): ReadonlyArray<T> {
+  if (!array) array = [];
+  if (item) return [...array, item];
+  return array;
+}
+
+export function updateArrayItem<T extends { id?: string }>(
```
Moejoe90: I wonder if we can get rid of this util

desfero: we probably can replace them with lodash, maybe I will do some cleanup during onfido changes

Issue #3599, opened 2019-10-09
Issuer Flow - Whitelisting (Your ETO Starts Soon)


pstroinski:
**Background**
Issuer should see a proper status of the ETO (Whitelisting) during the whitelist stage.
We should also display two modules:
- whitelisting module (already implemented)
- set date module

We need to adjust the ETO Date module:
Title: ETO DATE
Body: You have successfully set the date for your ETO. Make sure you communicate this to your community. The whitelist subscription will be automatically disabled 14 days before the start of the presale.

Presale starts in:
The date of the ETO can still be changed for the next ___ days, ___ hours.

Button: Change the ETO date

**Designs**
We should adjust the module with the date box.
![should_be_in_this_format_with_different_data](https://user-images.githubusercontent.com/27957897/66492589-f4961480-eab4-11e9-8428-e9cafc5ec39a.png)

**Acceptance criteria**
- [ ] Available module with proper whitelist status (open, closed, disable)
- [ ] Module with the set date option with adjusted copy

Louisswryan:
Hey @drakkein 

Please see below a solution for this:

https://zpl.io/aN58mWn


It isn't so nice but this who environment is being reviewed.

One change that is necessary, is to change the countdown timer to the actually date of presale. It is my understanding that currently it is set to when the issuer may no longer change the date.

I will contact you directly about this.

Cheers!

glupiawymowka:
tested on stage

PR #3733, opened 2019-11-13
#3599 Improve whitelisting in issuer flow

drakkein:


packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -75,12 +76,14 @@ export const calculateWhitelistingState = ({
...
   isAuthorized,
+  isInvestor,
 }: TcalculateWhitelistingState) => {
   //TODO this should be put in line with api after API's canEnableBookbuilding after limit is reached gets fixed by Marcin
   if (
     isAuthorized &&
+    isInvestor &&
     whitelistingIsActive &&
-    (investorsCount === undefined || investmentCalculatedValues === undefined)
+    investmentCalculatedValues === undefined
```
Moejoe90: and what happens if hte user is not an investor?

drakkein: settled on slack

packages/web/intl/locales/en-en.json
```diff
@@ -1116,12 +1116,13 @@
...
-  "settings.book-building-widget.start-book-building": "Enable subscription for your whitelist",
   "settings.book-building-widget.stop-book-building": "Disable whitelisting",
-  "settings.choose-pre-eto-date.book-building-will-stop": "Your ETO smart contracts have been deployed and your nominee signed all legal agreements. After setting the date of the ETO you will be asked to confirm all ETO terms and sign all legal agreements on-chain. Set the start date and time (UTC time) of your presale now. Public sale will automatically start after your presale ends. Unless you disable the subscription for the whitelist manually it will automatically close {minOffsetPeriod, plural, one {# day} other {# days}} before your chosen presale date",
+  "settings.book-building-widget.whitelist-description": "After Whitelisting, investors can sign up and indicate the amount they’d like to invest. You can disable the whitelist any time but please note - all of Neufund’s investors will be notified every time you disable and enable the whitelist. The whitelist will be automatically disabled 7 days before your presale starts.",
```
Moejoe90: These shouldn't be hardcoded

packages/web/intl/locales/en-en.json
```diff
@@ -1116,12 +1116,13 @@
...
+  "settings.choose-pre-eto-date.book-building-will-stop": "You have successfully set the date for your ETO. Make sure you communicate this to your community. The whitelist subscription will be automatically disabled 14 days before the start of the presale.",
```
Moejoe90: This shouldn't be hardcoded at all

Moejoe90:
> * Let me ask around about the days to close. We can set a presale to tomorrow, then why do we hardcode 7 days for the whitelisting? It shows 0.333 days on .io currently (also not so fortunate, but more accurate)

This shouldn't be hardcoded at all. this should be an automatic value coming from the backend so all answers will be correct for some cases and wrong for the other.

Issue #3718, opened 2019-11-11
Issuer Flow - In-Signing (Your ETO Starts Soon)


pstroinski:
**Background**
We should display two modules on the screen, currently, it's only one:
![image.png](https://images.zenhubusercontent.com/5beab1a20cefa317ea424319/854cc632-3ff7-46b0-97f9-fc4fbb14cac2)

Thre should be the one with the ISHA as displayed above but we also need to add:
- [x] countdown to the end the signing phase with a CLEAR indication that if the documents won't be uploaded on time the ETO will fail

**Designs**
Should be done in the same way as the counter in the pre-sale or public sale. Just change the text that's till the end of pre-sale to the end of the signing.

**Acceptance criteria**
- [x] Available module for the signed ISHA upload
- [x] Showing module with a countdown to the end of the signing with the clear indication that if not done on time ETO will fail

andywalton:
![Upload ISHA countdown timer + copy](https://user-images.githubusercontent.com/609657/68871128-40fbf200-06fc-11ea-949c-014661824ff3.png)

@drakkein here is a quick mockup. The designs are in Zeplin here: https://zpl.io/an7WkjY

Please use our existing Presale / Public sale component and update the text as follows:

**Title:**
SIGNED ISHA DEADLINE

**Intro paragraph:**
Please upload your signed Investment and Shareholders Agreement within the next:

`<Countdown timer + deadline date>`

**Warning text:**
Note, if you don’t do this your fundraising campaign will fail and all collected funds will be returned to the investors who invested.

glupiawymowka:
tested on stge

PR #3743, opened 2019-11-16
#3718 Add ISHA sign counter

drakkein:


packages/web/app/modules/eto-flow/selectors.ts
```diff
@@ -356,3 +356,17 @@ export const selectAreAgreementsSignedByNominee = createSelector(
     return undefined;
   },
 );
+
+export const selectIssuerEtoNextStateStartDate = createSelector(
+  selectIssuerEtoWithCompanyAndContract,
+  eto => {
+    if (eto && isOnChain(eto)) {
+      const nextState: EETOStateOnChain | undefined = eto.contract.timedState + 1;
```
Moejoe90: I think we should add a comment on why we always add 1 to timedState

PR #3753, opened 2019-11-19
 routing-based lazy loading

AlexeyTuboltsev:
here's a small helper component for compose() to simplify lazy loading of components


Moejoe90:
Lets have some checks first to see if its worth it to add this for now

PR #3773, opened 2019-11-24
Do not stop watching for transaction on network error

desfero:


packages/web/app/modules/tx/event-channel/sagas.ts
```diff
@@ -59,46 +59,59 @@ export function* getTransactionOrThrow(
   return tx;
 }
 
-// onNewBlock should return true to finish observing
+/**
+ * Watch for a given `txHash` and emit events to the `txChannel`
+ * @note `watchForTx` won't cancel watching on her own.
```
Moejoe90: `watching on its own.`

packages/web/app/modules/tx/event-channel/sagas.ts
```diff
@@ -59,46 +59,59 @@ export function* getTransactionOrThrow(
...
+ */
 export function* watchForTx(
   { web3Manager }: TGlobalDependencies,
   txHash: string,
   txChannel: Channel<TEventEmitterChannelEvents>,
 ): Iterator<any> {
   let lastBlockId = -1;
-  try {
-    while (true) {
+  while (true) {
```
Moejoe90: So we agree here that there are no terminal errors correct?

PR #3778, opened 2019-11-25
fix lazy-loading in e2e tests

AlexeyTuboltsev:


packages/web/app/config/.env.common
```diff
@@ -35,7 +35,7 @@ NF_GOOGLE_TAG_ID=UA-88781119-2
 # Enables NEUR operations in wallet Purchase and Redeem
 NF_NEURO_WITHDRAW_ENABLED=1
 
-NF_SENTRY_DSN=
+NF_SENTRY_DSN=https://1a01753f602c4bbcad39de77a992fc8d@sdsn.neufund.io/7
```
Moejoe90: please remove this we don't need to report the errors to sentry anymore

packages/web/app/modules/nominee-flow/lazy.ts
```diff
@@ -1,2 +1,9 @@
-export const generateHash = (buffer: Buffer): Promise<string> =>
-  import("ipfs-only-hash").then((Hash: IpfsHasher) => Hash.of(buffer));
+export const lazyLoadIpfsOnlyHash = (): Promise<any> => {
+  if (process.env.NF_CYPRESS_RUN === "1") {
+    return new Promise(resolve => {
+      resolve(require("ipfs-only-hash"));
+    });
+  } else {
+    return import(/* webpackChunkName: "ipfs-only-hash" */ "ipfs-only-hash");
```
Moejoe90: why do you have this comment here?

AlexeyTuboltsev: this is a way for webpack to specify a name of the chunk. Otherwise it uses a random string for it.

packages/web/app/config/.env.common
AlexeyTuboltsev: done

packages/web/app/modules/nominee-flow/lazy.ts
AlexeyTuboltsev: comment left

.circleci/config.yml
```diff
@@ -151,23 +151,12 @@ workflows:
   mainflow:
     jobs:
       - install-lint-test
```
Moejoe90: can you bring these back?

AlexeyTuboltsev: done

PR #3795, opened 2019-11-27
Fix nominee race conditions

desfero:
**Technical notes**
We probably need some e2e tests for that edge case (when more than one nominee ETO is on-chain) but I need to move back to KYC flow. Will do a cleanup ticket to include e2e tests.


packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -142,28 +158,40 @@ export function* getDataAndInitNomineeTasks(_: TGlobalDependencies): Iterator<an
...
+    if (isEmpty(nomineeEtos)) {
       yield put(actions.nomineeFlow.storeNomineeTasksStatus(nomineeTasksStatus));
       return;
     } else {
       nomineeTasksStatus[ENomineeTask.LINK_TO_ISSUER] = ENomineeTaskStatus.DONE;
     }
 
-    let etoSpecificNomineeTaskStatusEffects: { [key: string]: Iterator<Effect> } = {};
-    Object.keys(data.nomineeEtos).forEach((key: string) => {
-      etoSpecificNomineeTaskStatusEffects[key] = neuCall(
```
desfero: sorry, was forced to remove mutation inside `forEach`, I'm not able to sleep at night if I don't that we have mutations inside forEach methods :D

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -630,7 +658,7 @@ export function* nomineeFlowSagas(): Iterator<any> {
   yield fork(neuTakeLatest, actions.nomineeFlow.loadNomineeRequests, loadNomineeRequests);
   yield fork(neuTakeLatest, actions.nomineeFlow.createNomineeRequest, createNomineeRequest);
   yield fork(
-    neuTakeLatest,
+    neuTakeEvery,
```
desfero: ok, so this one is interesting because we call `initNomineeEtoSpecificTasks` in parallel for all ETOS we end up canceling previous `nomineeLoadEtoAgreementsStatus` actions but we should not as they all get status for different etos.

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -89,9 +96,18 @@ export function* initNomineeEtoSpecificTasks(
 
     if (isOnChain(eto)) {
       yield put(actions.nomineeFlow.nomineeLoadEtoAgreementsStatus(eto));
-      yield take(actions.nomineeFlow.nomineeSetAgreementsStatus);
+      yield neuTakeOnly(actions.nomineeFlow.nomineeSetAgreementsStatus, {
```
desfero: because `initNomineeEtoSpecificTasks` is called in parallel for all nominee etos we need to only take `nomineeSetAgreementsStatus` for that specific ETO. In general, I'm not a fun of dispatching and action (line above) and then waiting for it to complete by waiting for another action because it violates separation and introduces very strong coupling (imagine someone change the logic inside saga responsible for `nomineeSetAgreementsStatus` and now it doesn't dispatch `nomineeSetAgreementsStatus` or dispatch it when only some condition is met). In general, probably we should always use `neuCall` inside sagas as it's just easier to understand and we don't need to wait for any actions to be dispatched

Moejoe90: I agree with you as long as the sagas are withing the same module. 
I think the moment the sagas are from a seperate module we should start calling them via actions to remove the code coupeling.  what do you think?

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -142,28 +158,46 @@ export function* getDataAndInitNomineeTasks(_: TGlobalDependencies): Iterator<an
 
     yield all([neuCall(loadBankAccountDetails), neuCall(loadNomineeEtos)]);
 
-    const data = yield all({
+    const {
```
Moejoe90: I really don't understand why we would select via `all` its just hiding the logic without much optimization as its only a selector

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -142,28 +158,46 @@ export function* getDataAndInitNomineeTasks(_: TGlobalDependencies): Iterator<an
...
     }
 
-    let etoSpecificNomineeTaskStatusEffects: { [key: string]: Iterator<Effect> } = {};
-    Object.keys(data.nomineeEtos).forEach((key: string) => {
-      etoSpecificNomineeTaskStatusEffects[key] = neuCall(
-        initNomineeEtoSpecificTasks,
-        data.nomineeEtos[key],
-      );
-    });
+    const etoSpecificNomineeTaskStatusEffects = Object.values(nomineeEtos).reduce<{
```
Moejoe90: My issue with this change is that `reduce` is not typescript freindly you can see by having to type everything there. 
It was my request to remove `reduce` actually.

PR #3799, opened 2019-11-27
Changes getReceipt and getTransaction to use the wallet node

Moejoe90:
This also hides the `internalWeb3Adapter` all communication should be done via `Web3Manager` and its in `Web3Manager` we decied which web3 should conduct the specific `web3`


packages/web/app/lib/web3/Web3Manager/Web3Manager.ts
```diff
@@ -40,7 +41,7 @@ try {
 export class Web3Manager extends EventEmitter {
   public personalWallet?: IPersonalWallet;
   public networkId!: EthereumNetworkId;
-  public internalWeb3Adapter!: Web3Adapter;
+  protected internalWeb3Adapter!: Web3Adapter;
```
desfero: <3

packages/web/app/lib/web3/light-wallet/LightWallet.ts
```diff
@@ -168,84 +167,3 @@ export class LightWallet implements IPersonalWallet {
     return !!this.password;
   }
 }
-
-@injectable()
-export class LightWalletConnector {
```
desfero: good, wanted to move connected from lightwallet also

packages/web/app/lib/web3/Web3Manager/Web3Manager.ts
desfero: 🥇

packages/web/app/lib/web3/browser-wallet/BrowserWallet.ts
```diff
@@ -76,6 +76,14 @@ export class BrowserWallet implements IPersonalWallet {
     }
```
desfero: Can we also move `BrowserWalletConnector` as a separate file? Then we will reach full consistency between all providers

PR #3805, opened 2019-11-28
Optimizes EtoLinkStep

Moejoe90:


packages/web/app/test-e2e/eto-registration/EtoLinkNomineeStep.spec.e2e.io.ts
```diff
@@ -1,35 +1,25 @@
...
-  fillRequiredCompanyInformation();
-
-  assertPublishListingPage();
-
-  // fill Eto Terms and Investment Terms
-  fillAndAssert("eto-progress-widget-investment-terms", investmentTermsRequiredForm);
-
-  fillAndAssert("eto-progress-widget-eto-terms", etoTermsRequiredForm);
-
+  pushEtoToAPI(getJwtToken());
```
desfero: 👍

packages/web/app/test-e2e/eto-registration/utils.ts
```diff
@@ -0,0 +1,319 @@
+export const pushEtoToAPI = (jwt: string) => {
+  cy.request({
+    method: "PUT",
+    url: "/api/eto-listing/companies/me",
+    headers: {
+      "Content-Type": "application/json",
+      authorization: `Bearer ${jwt}`,
+    },
+    body: {
+      brand_name: "Davecoin",
```
desfero: Maybe we can allow overwriting data? And by default, we can use our eto fixtures?

packages/web/app/test-e2e/eto-registration/utils.ts
```diff
@@ -0,0 +1,319 @@
+export const pushEtoToAPI = (jwt: string) => {
```
desfero: Why we just don't call `getJwtToken` inside `pushEtoToApi`, it should not be a responsibility of a callee to pass token. You can even use our helper method `makeAuthenticatedCall`

desfero: And what about `EtoRegistrationUtils` file? Maybe this one should go there? Or we can introduce some utils folder here as there are a plenty of them already

packages/web/app/test-e2e/eto-registration/utils.ts
```diff
@@ -0,0 +1,319 @@
...
+  cy.request({
+    method: "PUT",
+    url: "/api/eto-listing/companies/me",
```
desfero: also would be good to keep urls in constants, we even have one already called `COMPANIES_ME_PATH`

packages/web/app/test-e2e/eto-registration/utils.ts
```diff
@@ -0,0 +1,319 @@
...
+          percent: 0.65,
+        },
+      ],
+      zip_code: "12345",
+    },
+  });
+};
+
+export const pushEtoDataToAPI = (jwt: string) => {
+  cy.request({
```
desfero: same comments apply here from the method above

packages/web/app/test-e2e/eto-registration/utils.ts
Moejoe90: done

packages/web/app/test-e2e/eto-registration/utils.ts
Moejoe90: I was thinking about this issue. like who is resposible for providing the JWT. 
In all cases since `makeAuthenticatedCall` then I'll use that

Moejoe90: There will be a full file structure overhaul for these utils maybe also full documentation. 
So i will keep for now

packages/web/app/test-e2e/eto-registration/utils.ts
Moejoe90: done

packages/web/app/test-e2e/eto-registration/utils.ts
Moejoe90: I will also do this as part of the overhaul because our fixtures are out of date i feel at least they werent working

PR #3823, opened 2019-12-07
[All devs should approve] - Updates Redux-Saga and Typescript with minimal change

Moejoe90:
This updates Redux-Sagas + Typescript

Have a look at the breaking changes and new features for `Redux-sagas`
https://github.com/redux-saga/redux-saga/blob/master/CHANGELOG.md#breaking-changes

Also have a look at typescripts latest nice features like `typechaining`
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-6.html

Migrating all sagas in one PR proved to be impossible thus thus PR touches only the bare minimum in order to make the app run.


packages/web/app/modules/sagas.ts
```diff
@@ -84,13 +84,13 @@ function* allSagas(): Iterator<Effect> {
...
 
-function* handleRootError(error: Error): Iterator<Effect> {
+function* handleRootError(error: Error): Generator<any, any, any> {
   const { logger }: TGlobalDependencies = yield getContext("deps");
 
   logger.error(error);
 }
 
-export function* rootSaga(): Iterator<Effect> {
+export function* rootSaga(): /* Iterator<unknown, Saga<any[]>> */ any {
```
drakkein: Why do we have this comment here? Shouldn't we remove it if there is an `any` used now?

Moejoe90: good catch this should be typed as `Generator<any, any, any>`

Moejoe90: fixed here 
https://github.com/Neufund/platform-frontend/pull/3843

packages/web/app/components/settings/backup-seed/BackupSeedVerify.unsafe.tsx
```diff
@@ -112,8 +112,9 @@ class BackupSeedVerify extends React.Component<IBackupSeedVerifyProps, IBackupSe
         clearable={true}
         matchPos="start"
         matchProp="value"
-        ref={(ref: TElementRef<VirtualizedSelect>) =>
-          (this.verificationSelectRefs[wordOnPageNumber] = ref)
+        ref={
+          ((ref: TElementRef<VirtualizedSelect>) =>
```
desfero: This one is really strange, i tried to understand the real reason but failed after a couple of attempts. Maybe we can set `TElementRef<any>` to overcome the error so then there is no need to use force casting?

packages/web/app/components/shared/formatters/FormatNumberRange.tsx
```diff
@@ -61,7 +61,11 @@ export const FormatNumberRange: React.FunctionComponent<INumberRangeProps> = ({
       outputFormat,
     });
 
-    if (Object.values(ESpecialNumber).includes(valueUpto)) {
+    if (
+      typeof valueUpto === "string" &&
+      Object.values(ESpecialNumber).includes(valueUpto as ESpecialNumber)
```
desfero: i would prefer here to write type predicate 
```
const isSpecialNumber = (
  value: TBigNumberVariants | undefined | null | ESpecialNumber,
): value is ESpecialNumber =>
  !!value && Object.values<ESpecialNumber | TBigNumberVariants>(ESpecialNumber).includes(value);
```
and then 
```
 if (
      isSpecialNumber(valueUpto)
      // TODO: Figure out why is this assertion like this
    ) {
```
and then we can also remove force casting from line `73`
```
          {getSpecialNumberTranslation(valueUpto)}
```

packages/web/app/components/wallet-selector/wallet-recover/recovery/SeedRecovery.unsafe.tsx
```diff
@@ -56,7 +56,7 @@ export class WalletLightSeedRecoveryComponent extends React.Component<
         clearable={true}
         matchPos="start"
         matchProp="value"
-        ref={(ref: VirtualizedSelect | null) => (this.verificationSelectRefs[index] = ref)}
+        ref={((ref: VirtualizedSelect | null) => (this.verificationSelectRefs[index] = ref)) as any}
```
desfero: same here, we can apply `ref: TElementRef<any>`

packages/web/scripts/extract-intl-messages.sh
```diff
@@ -10,7 +10,7 @@ if [[ $* == *--check* ]]; then
 fi
 
 echo "Creating messages..."
-node_modules/.bin/babel "./app/**/*.{ts,tsx}" > /dev/null
+node_modules/.bin/babel --plugins @babel/plugin-proposal-optional-chaining "./app/**/*.{ts,tsx}" > /dev/null
```
desfero: so placing the plugin in the `babelrc` config is not an option here?

Moejoe90: I don't want to include it in the general babel settings, as its purely done for the translation script. 

I didn't want to mix the plugins together

Moejoe90:
As pe our discussion I will try to conduct these fixes in the next PR

Issue #2508, opened 2019-02-25
Start using new block feature from Provider Engine


Moejoe90:
**Background**
Provider Engine has a callback function for new blocks. This feature can be used to emit `NEW_BLOCK_MINED` events to the app. All previous calls that depend on timely delay for polling can actually get triggered by `NEW_BLOCK_MINED`. 

Advantages: 
 - This will minimize the number of redundant requests to the eth node
 - All direct polling requests can be batched into one request

Provider Engine is currently outdated. 

Please update Provider Engine to the latest version 
Things to get taken into account is:

Bundle Size if its too big we won't add it

Check with @rudolfix the issue he wanted to solve  
https://github.com/Neufund/platform-frontend/pull/3848
**Designs**
N/A
**Acceptance criteria**
 - All three wallets should Emit `NEW_BLOCK` event while the app is running. 

 - Replace all polling mechanism with new block trigger

Good to have: 
 - All polling calls should be batched

**Technical notes**

**Testing notes**

rudolfix:
please note that we should have only one provider that can generate this event at a time. 
what about unlogged state? should our "read only" provider by used to generate new block events? wallet providers are only in context of logged on user

Moejoe90:
I will update the ticket so that it uses `https://github.com/MetaMask/eth-block-tracker`
With this we only need to wrap a web3 object with the tracker then events will happen automatically

PR #3848, opened 2019-12-10
uses dummy block tracker for wallet engine

rudolfix:
**What it does**
Prevents downloading any blocks by the wallet engine. Default implementation requests blocks with all transactions included which can be quite large. Now no blocks will be downloaded at ll.

**Why we do it**
We now route such requests through our backend and this has lower performance than our load balanced nodes. From user perspective you often wait for many seconds after clicking login button because (somehow) engine waits for first block to be downloaded...


Moejoe90:
Closing as this should be solved in https://github.com/Neufund/platform-frontend/issues/2508

PR #3850, opened 2019-12-10
Wraps sagas around a package + Documentation

Moejoe90:


packages/sagas/.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
```
drakkein: Can we make it as `.js` file config and reexport settings from `shared/prettierrc,json` as it is done in `web` package
```javascript
module.exports = {
  ...require("@neufund/shared/.prettierrc.json"),

  // override config below
};
```

drakkein: Discussed on Slack

packages/web/app/test-e2e/utils/assertions.ts
```diff
@@ -135,15 +135,15 @@ export const assertWaitForExternalPendingTransactionCount = (
   });
 };
 
-export const assertIsUserVerifiedOnBlockchain = (address: string, timeout: number = 5000) => {
+export const assertIsUserVerifiedOnBlockchain = (address: string, timeout: number = 10000) => {
```
desfero: +1

Issue #3744, opened 2019-11-17
Changes to eto progress bar logic and visualisation


Louisswryan:
Hello,

We are making a change to the logic of the eto progress bar. 
This affects how we show softcap and hardcap. 

From now on, the progress bar moves from empty to softcap which is called 'target' and is our new point of completion for the progress bar. 

After this, the amount raised can continue to grow up until the point of hardcap or 'limit'.
In this instance, the percentage funded will simply continue to rise past the 100% mark as is done by seedrs among other platforms. Once at 100%, the progress bar stays the same regardless of hardcap being reached ect...

Zeplin link with designs  below:
https://zpl.io/2ExEPYy

This progress bar features on the eto card, the eto widget and portfolio so each must be accounted for.

Please note the ETO widget contains all of the same content but is arranged in a way to guide the investor towards the call to action. Please ask if there are difficulties surrounding this. I am aware that this is for a quick release so understand if cannot be a priority.

For the portfolio view, we simply apply the new progress bar. Can we please move the campaign countdown to underneath the token name and rather than using the view profile link, simply link the token name to the profile to clear that necessary space.

There are one or two more visual changes here e.g. the change of the lichtenstein offering tab and removal of the filtering subjects. Let's talk this through once this has been assigned and decide what we can do and what is unrealistic or unnecessary.

Thank you!

Louisswryan:
Had a meeting today with @drakkein to decide on priorities for this task. The conclusions are listed below.

 - New progress bar and logic remains the core feature. 

 - ETO widget button changes are not actionable as this would create overlapping work so ETO widget will remain as it currently is (except for the replaced progress bar).

- In the investor profile, the %funded will shrink to 12px font size and the size of the progress bar will be justified to the size of the column that the bar sits within.

 - A further change is necessary in the profile section. The progress bar should be accompanied by the remaining time left in the eto/phase of eto. This will sit beneath the token name. So that it has enough space, it is necessary that we remove the 'view profile' link that sits to the right of the token name. We will increase the token name to 16px font size and make the block that contains the name and token icon the link to the issuer profile

Louisswryan:
Hello,

This update is in response to an issue that was highlighted this morning:

Currently, the target amount in the progress bar is susceptible to change based on discounts that are given out, usually in the presale phase.

This amount has been designed to be static so needs to be made dynamic to reflect this need.

We also need to make the original target amount accessible so will do this with the use of a tooltip.


New rule for tooltip

[...]

PR #3855, opened 2019-12-11
Cleanup for #3744

drakkein:
- Arithmetics related utilities are stored in **packages/web/app/lib/api/eto/EtoUtils.ts**
- State related utilities are stored in **packages/web/app/modules/eto/utils.ts**
- Progress percentage and target cap calculations has been rewritten to use BigNumber


packages/web/app/components/eto/shared/InvestmentAmount.tsx
```diff
@@ -11,23 +13,31 @@ import {
 } from "../../shared/formatters/utils";
 import { Tooltip } from "../../shared/tooltips/Tooltip";
 import { ECustomTooltipTextPosition } from "../../shared/tooltips/TooltipBase";
+import { Entry } from "../public-view/EtoInvestmentTermsWidget";
```
desfero: If we want to use `Entry` here then we should extract it from `EtoInvestmentTermsWidget` as otherwise, you have a circular reference between `EtoInvestmentTermsWidget` and `InvestmentAmount`. Also if we use `Entry` here it's then coupled with the view of `EtoInvestmentTermsWidget` therefore it should be moved outside of `eto/shared` folder to keep consistency.

packages/web/app/components/eto/shared/InvestmentAmount.tsx
```diff
@@ -11,23 +13,31 @@ import {
...
 import { ToBeAnnounced } from "./ToBeAnnouncedTooltip";
 
 type TExternalProps = {
-  etoData: TEtoWithCompanyAndContractReadonly;
+  eto: TEtoWithCompanyAndContractReadonly;
```
desfero: 👍

packages/web/app/components/shared/ProgressBarSimple.stories.tsx
```diff
@@ -4,8 +4,8 @@ import React from "react";
 import { ProgressBarSimple } from "./ProgressBarSimple";
 
 storiesOf("NDS|Atoms/progress/ProgressBarSimple", module)
-  .add("0%", () => <ProgressBarSimple progress={0} />)
-  .add("25%", () => <ProgressBarSimple progress={25} />)
-  .add("50%", () => <ProgressBarSimple progress={50} />)
-  .add("75%", () => <ProgressBarSimple progress={75} />)
-  .add("100%", () => <ProgressBarSimple progress={100} />);
+  .add("0%", () => <ProgressBarSimple progress={"0"} />)
```
desfero: There is no need for `{ }` if we use strings, probably it's worth to add a linter rule to catch that

packages/web/app/components/shared/ProgressBarSimple.tsx
```diff
@@ -7,25 +8,38 @@ import { invariant } from "../../utils/invariant";
 import * as styles from "./ProgressBarSimple.module.scss";
 
 interface IExternalProps {
-  progress: number;
+  progress: string;
+  limitTo?: string;
```
desfero: i'm somehow not 100% about `limitTo` prop. Progress should never exceed `100%` as it's impossible from the visualization point of view therefore if it's possible somehow to be more than 100% then we should probably do some `if` level above so `ProgressBarSimple` is not aware of it.

packages/web/app/modules/eto/utils.ts
```diff
@@ -193,26 +196,18 @@ export const getEtoSubState = ({
     }
   }
 };
+export const getInvestmentCalculatedPercentage = (eto: TEtoWithCompanyAndContractReadonly) => {
+  if (eto && isOnChain(eto)) {
```
desfero: `eto` is always defined here (at least from typings point) so we can simplify just to `if (isOnChain(eto))`

packages/web/app/components/shared/ProgressBarSimple.tsx
drakkein: ok, it is used only once anyway, so it will be good to not overcomplicate the progressbar component

packages/web/app/components/shared/ProgressBarSimple.tsx
```diff
@@ -12,21 +12,11 @@ interface IExternalProps {
   limitTo?: string;
```
desfero: `limitTo` prop should be removed too

packages/web/app/components/shared/Entry.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
```
desfero: Maybe we can add `stories` here given it's moved to the core `shared` folder?

Issue #3854, opened 2019-12-11
Refresh kyc status after submitting KYC personal data


desfero:
**Background**
After we merge backend https://github.com/Neufund/platform-backend/pull/2355 KYC supported instantid provider will also depend on the individual country therefore after updating kyc data we need to refresh kyc status to get the newest requirements for instantid providers.

**Designs**

**Acceptance criteria**
- [x] Given I chose a country with a not supported kyc provider (for e.g. Vatican) When i go to the `Identity verification` page Then notification appears with a proper information 
- [x] Given I chose a country with a not supported onfido kyc provider When I go to the `Identity verification` page Then only idnow is available
- [x] Given I chose a country with a not supported idnow kyc provider When I go to the `Identity verification` page Then only onfido is available

**Technical notes**

**Testing notes**

desfero:
already deploy to production

PR #3856, opened 2019-12-11
#3854 Refresh KYC status

desfero:


yarn.lock
```diff
@@ -15971,6 +15971,16 @@ redux@^4.0.4:
     loose-envify "^1.4.0"
     symbol-observable "^1.2.0"
 
+redux@~3.5.2:
```
desfero: seems this one was missing as `onfido-sdk-ui` depends on it. Strange that `yarn install --frozen-lockfile` didn't fail on CI

packages/web/webpack/webpack.config.dev.js
```diff
@@ -45,7 +45,7 @@ module.exports = merge.smart(configCommon, {
         "*; " + // this should be only enabled for twitter-iframe.html
         "font-src 'self' fonts.gstatic.com; " +
         "media-src 'self' blob:; " +
-        "img-src 'self' blob: data: documents.neufund.io documents.neufund.net www.google-analytics.com stats.g.doubleclick.net " +
+        "img-src 'self' blob: data: documents.neufund.io documents.neufund.net www.google-analytics.com stats.g.doubleclick.net https://lipis.github.io/flag-icon-css/flags " +
```
desfero: https://lipis.github.io/flag-icon-css/flags is used by `onfido-sdk-ui`

packages/web/app/modules/kyc/sagas.ts
```diff
@@ -35,18 +36,33 @@ import {
...
     const kycStatus: TKycStatus = yield apiKycService.getKycStatus();
 
     yield put(actions.kyc.setStatus(kycStatus));
+  } catch (e) {
+    logger.error("Error while getting KYC status", e);
+
+    yield put(actions.kyc.setStatusError(e.message));
+
+    // let the caller know that status is unknown
+    throw e;
```
Moejoe90: the only issue i see here is if an error occurs here we will have two logged errors in sentry

packages/web/app/modules/kyc/sagas.ts
```diff
@@ -146,6 +175,9 @@ function* submitPersonalDataSaga(
 ): Generator<any, any, any> {
   const result: IHttpResponse<IKycIndividualData> = yield apiKycService.putPersonalData(data);
 
+  // update kyc status after submitting personal data as it may affect supported instant id providers
+  yield neuCall(loadKycStatus);
```
Moejoe90: and for example here do we rely on the root saga also?

packages/web/webpack/webpack.config.dev.js
Moejoe90: we probably need a devops ticket for this no?

pstroinski: This is already implemented

packages/web/app/modules/kyc/sagas.ts
desfero: let's keep two logs for now

Issue #1290, opened 2018-09-05
Implement visual regression on full views with Cypress 3


Moejoe90:
Cypress 3 has a good feature of screenshotting full views, this can be the bases for a good visual regression tool that.

`Edit:`
After the initial merge with code checks implement an initial test that uses screenshotting with cypress and code checks

desfero:
@Moejoe90 that make sense and it will really help with changing any kind of styles. 
We need to think about all the problems that can arise. For e.g. using fixtures in tests can break visual regression as wallet amount will not always be the same.

Moejoe90:
This should be easy now after migrating to codechecks 
Mostly will be done by chris

Moejoe90:
@desfero Good that you mentioned this point. as per the meeting we will use component blacking to hide any components that would change due to fixtures or anything

Moejoe90:
This is an example of how how to implement this feature 
https://github.com/codechecks/vis-reg-cypress-example

PR #3857, opened 2019-12-11
#1290 Part 1 Cypress visual regressions

drakkein:
This PR contains initial solution for Cypress visual regression.


.circleci/config.yml
```diff
@@ -96,9 +100,9 @@ jobs:
     working_directory: /home/circleci/app
 
     steps:
+      - checkout
```
Moejoe90: why do we need the checkout here?

packages/web/app/styles/global.scss
```diff
@@ -3,7 +3,6 @@
 html,
 body,
 [id="app"] {
-  height: 100%;
```
Moejoe90: this is changing the whole global style of the app why do you need it here?

.circleci/config.yml
```diff
@@ -137,6 +136,24 @@ jobs:
...
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v25-deps-{{ checksum "package.json" }}
+            - v25-deps-
+
+      - run: yarn --frozen-lockfile
+      - run:
+          command: yarn workspace @neufund/web build:prod:e2e:io
```
Moejoe90: why do you need to build the production build here? 
Is there something special going on

.circleci/config.yml
```diff
@@ -173,6 +191,10 @@ workflows:
       - codechecks-vis-reg:
           requires:
             - build-frontend
+      #temporary only wait for .io tests
+      - codechecks-e2e-vis-reg:
+          requires:
+            - build-frontend
 
   nightly:
```
Moejoe90: Why not add it to the nightly build also?

packages/web/app/test-e2e/vis-reg/Issuer.spec.e2e.visreg.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    cy.awaitedScreenshot('[aria-describedby="transactions-history-heading"]');
+  });
+
+  it("should render profile", () => {
+    goToProfile();
+
+    cy.screenshot();
+  });
+});
```
Moejoe90: What about the Nominee? 
I don't see any regression tests for the user type `Nominee`

packages/web/cypress/support/commands.js
```diff
@@ -125,3 +127,64 @@ Cypress.Commands.add("iframe", selector => {
     });
   });
 });
+
+Cypress.Screenshot.defaults({
```
Moejoe90: Please add Documentation (Add a document into our document section) on how visual regression works and when and why you should add blackout etc. 

This will help new developers and others get quick into the flow.

packages/web/regconfig_e2e.json
```diff
@@ -0,0 +1,11 @@
+{
+  "core": {
+    "actualDir": "cypress/screenshots"
+  },
+  "plugins": {
+    "reg-simple-keygen-plugin": {
```
Moejoe90: The documentation should include an explenation on why we need this plugin for example

.circleci/config.yml
drakkein: We have sometimes differences between dev and prod builds if it comes to styling. I want to catch them, so that's why I'm doing the prod build here

packages/web/app/styles/global.scss
drakkein: Cypress have some issues determining the viewport size when height is set

drakkein: Min-height has been preserved, so there should be no issues

.circleci/config.yml
Moejoe90: But that's what i meant we already make a `prod` build here in `build-frontend`

docs/e2e-visual-regression.md
```diff
@@ -0,0 +1,26 @@
+**Writing new visual regression tests**
+
+All visual regression tests should have *.spec.e2e.visreg.ts* extension in order to be run in vis-reg scope. 
+
+To perform screenshot you have to invoke 
+`cy.screenshot();` or `cy.awaitedScreenshot(selector: string);`. Built-in method will make a screenshot as-is, while *awaitedScreenshot* will ensure that specified selector is accessible before taking a screenshot.
```
Moejoe90: `visible`

PR #3864, opened 2019-12-18
Fix storybook deployment

desfero:
**Technical notes**
Besides fixing deployment also npm script name was changed to `storybook:deploy-ghpages` to preserve consistency


packages/web/.storybook/config.js
```diff
@@ -1,6 +1,6 @@
 import "../app/polyfills.sideEffect";
 
-import "./setup-styles";
+import "./setup-styles.sideEffect";
```
desfero: this line is an exact cause of a difference between local and production storybook build. Because webpack removes all unused `imports` during production build (at the moment it preserves `*.sideEffect*, .scss and .css` imports (check package.json of the web for `sideEffects` key).

packages/web/package.json
```diff
@@ -37,6 +36,8 @@
     "start:remote:serve-on-network": "(cp -n ./app/config/.env.example .env || true) && NODE_ENV=development && NF_SERVE_ON_NETWORK=$(eval network private_ip) webpack-dev-server --config ./webpack/webpack.config.dev-remote.js",
     "start:translate": "(cp -n ./app/config/.env.example .env || true) && NODE_ENV=development && NF_ENABLE_TRANSLATE_OVERLAY=1 webpack-dev-server --config ./webpack/webpack.config.translate.js",
     "storybook": "(cp -n ./app/config/.env.storybook .env || true) && start-storybook -p 9001 -c .storybook -s ./app/assets",
+    "storybook:build": "(cp -n ./app/config/.env.storybook .env || true) && build-storybook -c .storybook -s ./app/assets",
```
desfero: Here we now build storybook with the exact same env variables as the local storybook build. Previously we didn't use our `.env.storybook` file which makes two build a little bit different

packages/web/.storybook/config.js
drakkein: 🥇

PR #3865, opened 2019-12-18
Preserve aspect ratio of ETO main image

desfero:
**Testing notes**
This one removes the annoying jumping of the ETO main image. (please note it's only supported by latest browsers):
**Before**
Dec-18-2019 16-04-54
**After**
Dec-18-2019 16-02-55


packages/web/app/components/eto/overview/EtoOverviewThumbnail/Cover.tsx
```diff
@@ -5,13 +5,19 @@ import { FormattedMessage } from "react-intl-phraseapp";
...
 import * as styles from "./Cover.module.scss";
 
+type TImage = {
+  srcSet: ISrcSet;
+  alt: string;
+};
+
 interface IProps {
-  companyBanner: IResponsiveImage;
+  companyBanner: TImage;
```
desfero: `IResponsiveImage` is a pretty wrong type here as it contains all props from `ResponsiveImage` component. Same with all other related places modified

PR #3873, opened 2020-01-06
Add basic browser wallet test 

desfero:
**Technical notes**
It's pretty hard to test with real metamask with current e2e tests flow but the login flow is so important to tests to be covered therefore I decided to go with private key web3 provider. 

In general, IMO it's pretty okay to abstract what's behind `window.ethereum` as it should always adhere to web3 provider protocol therefore it should be in sync with how metamask works.

If the reviewer thinks it's okay in the next PR I will add also e2e tests for transaction sending with browser wallet.

Also, utils related to fixtures were moved to as separate `fixtures` file.


packages/web/app/modules/tx/transactions/withdraw/sagas.ts
```diff
@@ -45,7 +47,7 @@ export function* generateEthWithdrawTransaction(
   const valueUlpsAsBigN = new BigNumber(valueUlps);
 
   const etherTokenBalance: BigNumber = yield select(selectEtherTokenBalanceAsBigNumber);
-  const from: string = yield select(selectEthereumAddressWithChecksum);
+  const from: string = yield select(selectEthereumAddress);
```
desfero: For some reason, WalletSubprovider from `web3-provider-engine` accepts only a lowercased ethereum address. 
https://github.com/MetaMask/web3-provider-engine/blob/master/subproviders/wallet.js#L16

PR #3876, opened 2020-01-06
Updates lolex + fixes typings, updates cypress

Moejoe90:


Moejoe90:
done as part of `3852`

PR #3878, opened 2020-01-07
Do not refresh JWT after inactivity logout

desfero:


packages/web/app/test-e2e/eto-registration/SetEtoStartDate.spec.e2e.ts
```diff
@@ -1,41 +1,37 @@
...
 import { loginFixtureAccount } from "../utils/userHelpers";
 
 const setStartDay = (startDate: moment.Moment, textToCheck: RegExp) => {
-  cy.visit(appRoutes.dashboard)
-    .get(tid("eto-settings-start-date-open-date-picker"))
-    .click()
-    .get(tid("eto-settings-start-date-input"))
+  goToIssuerDashboard();
+
+  cy.get(tid("eto-settings-start-date-open-date-picker")).click();
```
arjunu: is there a plan to use constants for these ids?

desfero: At the moment we always use strings for `tid`s. In the long term, we want to migrate to use constants but no one yet proposed some consistent way of sharing them between e2e test and the app.

packages/web/app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
```diff
@@ -80,11 +83,110 @@ describe("JWT Refreshing and Escalation", () => {
 
       const diff = jwtExpiryDate.diff(now, "milliseconds");
 
+      let expectedTokenRefreshTimeFromNow = diff - AUTH_TOKEN_REFRESH_THRESHOLD;
+
+      keepSessionActive(expectedTokenRefreshTimeFromNow);
+
+      cy.route("POST", "**/jwt/refresh").as("jwtRefresh");
```
Moejoe90: please abstract this into a constant, we can still leave the alias here but at least the `cy.route` should be abstracted into one place

packages/web/app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
```diff
@@ -80,11 +83,110 @@ describe("JWT Refreshing and Escalation", () => {
...
+      cy.tick(AUTH_INACTIVITY_THRESHOLD / 2);
+
+      assertDashboard();
+
+      cy.tick(AUTH_INACTIVITY_THRESHOLD);
+
+      assertLogin();
+    });
+
+    cy.clock().then(clock => {
```
Moejoe90: this is a bit confusing can you confirm that even after calling a new instance of `cy.clock` you are actually restoring the same clock in `156`?

packages/web/app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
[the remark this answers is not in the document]
pavelneu: addressed, 
moved this line into the function in utils

packages/web/app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
```diff
@@ -80,11 +83,110 @@ describe("JWT Refreshing and Escalation", () => {
...
+      now += AUTH_INACTIVITY_THRESHOLD / 2;
+
+      assertDashboard();
+
+      cy.tick(AUTH_INACTIVITY_THRESHOLD);
+      now += AUTH_INACTIVITY_THRESHOLD;
+
+      assertLogin();
+
+      cy.route({
```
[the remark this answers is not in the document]
pavelneu: addressed

packages/web/app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
pavelneu: addressed,
put into a constant

packages/web/app/test-e2e/general-app/JWTEscalationAndRefreshing.spec.e2e.io.ts
pavelneu: @desfero can you help me answering on this question?

desfero: @Moejoe90 right, after calling `cy.restore`  here we do actually restore clock from the above, Actually, I will do a small change to nest `cy.tick` and `cy.restore` so it's visible who's actually owning a clock at the given moment

PR #3880, opened 2020-01-07
Enforce node version check

DmitriyNoa:
NodeJS version 11.X.X has to be used to ensure that all dependencies are installed properly


package.json
```diff
@@ -3,6 +3,9 @@
   "workspaces": [
     "packages/**"
   ],
+  "engines": {
```
desfero: will it show an error in case I'm on `^12.0.0`? AFAIK it will only show warning unless some flag is set during install. 
Maybe we can also add `.nvmrc` with some pinned version of node so we can just easily type `nvm use` and it will pick everywhere (CI, local dev machines) the same node version. What do you think?

DmitriyNoa: Not exactly. It throws an error: 
`error @: The engine "node" is incompatible with this module. Expected version "^11.0.0". Got "12.13.1"` and stops the installation proccess

desfero: good to know, always thought that it will just show a warning 👍

PR #3884, opened 2020-01-08
Check wallet recovery phrases on "Send Words"

arjunu:
- While recovering wallet, on click of Send Words, if the 24 words recovery phrase is wrong, it will show an error modal.
- the Menmonic is tested using `bitcore-mnemonic` which is now added as a dependency.
invalid-seed


packages/web/app/components/wallet-selector/wallet-recover/recovery/RecoverWallet.stories.tsx
```diff
@@ -18,9 +19,15 @@ storiesOf("RecoveryProcesses", module)
       },
     }),
   )
-  .add("Enter Seed", () => <RecoveryProcessesComponent submitForm={action("recover")} />)
+  .add("Enter Seed", () => (
+    <RecoveryProcessesComponent submitForm={action("recover")} onInvalidSeed={noop} />
```
desfero: just a nitpick, it would be good to use `action('onInvalidSeed`)` in place of lodash `noop` as then in storybook you can see in actions panel logs when the function is called. Same on line `29`

arjunu: done

packages/web/app/lib/web3/light-wallet/LightWalletUtils.ts
```diff
@@ -1,3 +1,4 @@
+import * as Mnemonic from "bitcore-mnemonic";
```
Moejoe90: Do you need here to import `*`? 
we only need to use `isValid`

packages/web/app/lib/web3/light-wallet/LightWalletUtils.ts
```diff
@@ -157,3 +158,5 @@ export const getWalletPrivKey = async (
     throw new LightWrongPasswordSaltError();
   }
 };
+
+export const testWalletSeed = (seed: string): boolean => Mnemonic.isValid(seed);
```
Moejoe90: :+1:

packages/web/app/test-e2e/wallets/WalletRecover.spec.e2e.io.ts
```diff
@@ -19,6 +19,18 @@ import {
 
 describe("Wallet recover", function(): void {
   this.retries(2);
+
+  it("should show error modal for invalid recovery phrases", () => {
+    cy.visit(`${recoverRoutes.seed}`);
+
+    const words = "mutual mutual phone brief hedgehog friend brown actual candy will tank case phone rather program clap scrap dog trouble phrase fit section snack world".split(
```
Moejoe90: Non-blocking 
I suggest renaming `words` to `wrongMnemonics` or `wrongWords` to indicate that these set of words should fail on the check

PR #3888, opened 2020-01-10
3762 optimize getting bookbuilding stats

DmitriyNoa:


packages/web/app/lib/api/eto/EtoApi.ts
```diff
@@ -149,6 +152,22 @@ export class EtoApi {
...
+  public getBookBuildingStatsList(etosIds: string[]): Promise<Dictionary<IBookBuildingStats>> {
+    return this.httpClient
+      .get<TBookBuildingsStatsList>({
+        baseUrl: BASE_PATH,
+        url: `/bookbuilding-stats?eto_ids=${etosIds}`,
+      })
+      .then(response => {
+        const etosStats: Dictionary<IBookBuildingStats> = response.body.bookbuildingStats;
+        // backend returns eto with ID in different casing, so it has to be lowercase
+        for (let key in etosStats) {
```
Moejoe90: As per our discussion the addresses shouldn't IMO be lower cased as they fulfill 
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md 
Which is good to have internally to confirm that this is a valid ethereum address. 

Please Invistigate why do we need to do this if this is because the `eto` end points returns`lower cased` addresses then we will need to ask the `backend` guys to change their endpoints and be consistent. 

Another case could be that we are lower casing this at some other point in the app which is also not good

DmitriyNoa: So the issue was in HttpCelint toSnakeCase function that snake cases the response from the backend and it affected the etoId as they were used as object keys. Changed to use web2-utils method to transform etoId back to valid format.

PR #3889, opened 2020-01-10
#2508 update web3 provider engine

AlexeyTuboltsev:
this PR
- updates web3-provider-engine
- adds a global web3 block tracking logic, events and actions that will serve as triggers for api calls instead of polling. (E.g. for eto data on eto state transitions)


packages/web/app/modules/web3/actions.ts
```diff
@@ -5,6 +5,9 @@ import { TWalletMetadata } from "./types";
 
 export const web3Actions = {
   personalWalletDisconnected: createActionFactory("PERSONAL_WALLET_DISCONNECTED"),
+  newBlockArrived: createActionFactory("NEW_BLOCK_ARRIVED", (blockNumber: string) => ({
```
Moejoe90: what do you think if we add a new `ETH_BLOCK_TRACKER_ERROR` action

Moejoe90: What do you think if we add also `ETH_BLOCK_TRACKER_ERROR` action? it would be usful sometimes for the app to deal with these issues

packages/web/app/modules/web3/sagas.ts
```diff
@@ -71,16 +71,31 @@ export function* loadPreviousWallet({
...
+      type: EWeb3ManagerEvents.NEW_PERSONAL_WALLET_PLUGGED;
+      payload: { isUnlocked: boolean; metaData: TWalletMetadata };
+    }
+  | {
+      type: EWeb3ManagerEvents.NEW_BLOCK_ARRIVED;
+      payload: { blockNumber: string };
+    }
+  | {
+      type: EWeb3ManagerEvents.ETH_BLOCK_TRACKER_ERROR;
+      payload: { error: unknown };
```
Moejoe90: is it really `unknown`? I assumed it will be an `error` object

packages/web/package.json
```diff
@@ -210,6 +211,7 @@
     "copy-text-to-clipboard": "^2.1.0",
     "crypto-random-string": "^3.0.1",
     "detect-browser": "^2.0.0",
+    "eth-block-tracker": "^4.4.3",
```
Moejoe90: please make sure to use exactly the same version used in `web3-provider-engine`

packages/web/typings/eth-block-tracker.d.ts
```diff
@@ -0,0 +1,29 @@
+declare module "eth-block-tracker" {
```
Moejoe90: :+1:

packages/web/typings/web3-provider-engine.d.ts
```diff
@@ -0,0 +1,36 @@
+// Type definitions for web3-provider-engine 14.0
+// Project: https://github.com/MetaMask/provider-engine#readme
+// Definitions by: Leonid Logvinov <https://github.com/LogvinovLeon>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+import { Provider, JSONRPCRequestPayload, JSONRPCResponsePayload } from "ethereum-protocol";
```
Moejoe90: why not use the same version from `definitly-typed` then or is it not on `npm?

yarn.lock
```diff
@@ -9437,6 +9464,44 @@ eth-block-tracker@^3.0.0:
     pify "^2.3.0"
     tape "^4.6.3"
 
+eth-block-tracker@^4.4.2, eth-block-tracker@^4.4.3:
```
Moejoe90: as expected its not the same version based on the yarn lock file at least

packages/web/package.json
Moejoe90: As expected please use version `4.4.2`

packages/web/app/modules/web3/actions.ts
AlexeyTuboltsev: done

packages/web/app/modules/web3/sagas.ts
AlexeyTuboltsev: done

packages/web/typings/web3-provider-engine.d.ts
AlexeyTuboltsev: web3-provider-engine 15.x has a different API, so I had to write my own def.

packages/web/package.json
AlexeyTuboltsev: they both are resolved to 4.4.3 so I'd leave this in place

btw there are other `web3-provider-engine` versions used as deps some packages: 14.1.0 (dep of ledger),
13.3.2 (by eth-tx-summary) and 14.0.6 (truffle-privatekey-provider) 
all of them use different (old) versions of `eth-block-tracker` too

yarn.lock
AlexeyTuboltsev: they both are resolved to 4.4.3

packages/web/package.json
Moejoe90: we need to clean them up can you add a `tech` ticket for this

PR #3894, opened 2020-01-13
Update Quintessence terms of use external link

arjunu:


arjunu:
Wrong base, will re-create PR

PR #3896, opened 2020-01-13
Update typechain

drakkein:


Moejoe90:
We can't implement this now unless we update web3

Issue #3695, opened 2019-10-31
Use bigNumber during calculations in /eto/utils.ts


Moejoe90:
**Description**
https://github.com/Neufund/platform-frontend/pull/3688/files#diff-3f365e3303b7807728cd92a4ec77367eR196-R205

needs to be changed to bigNumbers. 
**Environment**

**Steps To Reproduce**

**Actual behaviour**

**Expected behavior**

**Comments, context, logs**

**Screenshots**

PR #3899, opened 2020-01-14
#3695 Use BigNumber during calculations in /eto/utils.ts

drakkein:


packages/web/app/components/eto/utils.ts
```diff
@@ -85,25 +85,57 @@ type TConvertPercentageToFractionOptions = { passThroughInvalidData: true };
 export const convertPercentageToFraction = (options?: TConvertPercentageToFractionOptions) => (
   data: number | undefined,
 ) => {
-  const parseFn = (number: number) => parseFloat((number / 100).toPrecision(4));
+  const parseFn = (value: number | TBigNumberVariants): number => {
+    // cast to string to avoid errors with longer numbers
+    const valueBn =
+      typeof value === "number" ? new BigNumber(value.toString()) : new BigNumber(value);
+    const fractionBn = valueBn.dividedBy("100").toFixed(4);
```
DmitriyNoa: What do you think about moving precision to the function parameters?

drakkein: Yeah, this is good idea. I was refactoring it 1:1, so I was not even thinking about it.

Moejoe90:
Will close for now and reopen once https://github.com/Neufund/platform-frontend/pull/3807 is merged as it will require further change

Issue #2658, opened 2019-03-28
Remove `Confirm Your Transaction` modal for lightwallet users


desfero:
**Background**
Currently, during transaction flow, light wallet users always see `Confirm Your Transaction` modal with some text to sign a transaction on ledger or metamask. This behavior is not correct because the user already signed a transaction on the modal before (`Accept transaction` modal).

See `gif` below to see exact flow:
![Mar-26-2019 22-54-39](https://user-images.githubusercontent.com/7031578/55161708-baa89500-5198-11e9-9493-3879745e1b4b.gif)

After this issue is done user with a light wallet should not see `Confirm Your Transaction` modal.

**Designs**
none

**Acceptance criteria**
- [x] Given I'm a **light wallet** user when I'm doing through transaction flow (withdraw/payout/invest/...) then I **should not** see `Confirm Your Transaction` modal.
- [x] Given I'm a **metamask/ledger** user when I'm doing through transaction flow (withdraw/payout/invest/...) then I **should** see `Confirm Your Transaction` modal.

**Technical notes**
 After this change a couple of e2e tests will fail, so please fix them.

**Testing notes**
none

Moejoe90:
Since there needs to be something while the data loads a new `loading` state was added

glupiawymowka:
tested both for light and external wallets on stage

PR #3902, opened 2020-01-14
#2658 - Adds new Loading state for txSender

Moejoe90:
Removes txSender Signing state for LightWallet user


packages/web/app/modules/tx/sender/sagas.ts
```diff
@@ -133,7 +136,12 @@ function* txSendProcess(
     // accept transaction on wallet
     yield call(connectWallet);
 
-    yield put(actions.txSender.txSenderWalletPlugged());
+    // TODO: Use is wallet unlocked from WalletMeta instead of asserting the wallet type directly
```
Moejoe90: I didn't want to use any other wallet MetaData at the moment until the wallets are reviewed as in they work correctly and return the correct metadata

PR #3904, opened 2020-01-14
Fix recovery flow starts pooling

arjunu:
### Fix 
- The issue was `lightWalletRecoverWatch` saga dispatches `actions.auth.setUser` which starts Wallet sagas and a bunch of others that watch for that action.  Once wallet recovery is completed, logout action isn't dispatched and hence the started sagas aren't killed.  
- It also `loadKycRequestData` which is not required for recover wallet and hence is removed
- This PR updates `updateUser` saga to accept a third optional parameter `updateStore` which would be passed as `false` in this case which tells the saga not to dispatch `actions.auth.setUser`

### Refactor 
 - In `packages/web/app/modules/wallet-selector/light-wizard/sagas.ts`, replaced nested `throw`s that were being caught locally which makes code hard to read. Instead error method is called followed by a `return` statement
- removed unused `waitUntilUserLogoutIsDone` method


packages/web/app/modules/auth/user/external/sagas.ts
```diff
@@ -14,26 +14,27 @@ export function* loadUser({ apiUserService }: TGlobalDependencies): Generator<an
...
 
 export function* updateUser(
   { apiUserService }: TGlobalDependencies,
   updatedUser: IUserInput,
+  updateStore: boolean = true,
 ): Generator<any, any, any> {
   const user: IUser = yield apiUserService.updateUser(updatedUser);
-  yield put(actions.auth.setUser(user));
+
+  if (updateStore) {
```
Moejoe90: I think we should think of a way to get rid  of this as in if we don't need to update the store we can just call directly
  const user: IUser = yield apiUserService.updateUser(updatedUser);

packages/web/app/modules/wallet-selector/light-wizard/sagas.ts
```diff
@@ -130,42 +135,49 @@ export function* lightWalletRecoverWatch(
...
       } else {
-        if (user.verifiedEmail === email.toLowerCase()) yield neuCall(updateUser, userUpdate);
-        else {
-          throw new EmailAlreadyExists();
+        if (user.verifiedEmail === email.toLowerCase()) {
+          yield neuCall(updateUser, userUpdate, false);
+        } else {
+          yield neuCall(handleLightWalletError, new EmailAlreadyExists());
+          return;
         }
```
Moejoe90: maybe we move `160` and `163` to `167`
and call the api directly 
```
yield apiUserService.updateUser(updatedUser);
```

packages/web/app/modules/auth/user/external/sagas.ts
arjunu: the `updateStore` here is for profile sagas `addNewEmailEffect`, `abortEmailUpdateEffect` & `resendEmailEffect`

packages/web/app/modules/wallet-selector/light-wizard/sagas.ts
arjunu: simplified the if-else

arjunu:
Related tests are failing. Should wait for https://github.com/neufund/platform-frontend/issues/3910

PR #3906, opened 2020-01-15
[HOTFIX] Revert height change

drakkein:


packages/web/app/styles/global.scss
```diff
@@ -3,6 +3,7 @@
 html,
 body,
 [id="app"] {
+  height: 100%;
   min-height: 100%;
```
desfero: In general, I'm almost 90% sure we don't need `min-height` here but let's just for now push the fix

PR #3914, opened 2020-01-15
Fix investment type styles shift

DmitriyNoa:


Moejoe90:
Visual Regression Checked

PR #3927, opened 2020-01-20
Fixes Transfer flow Ether + Token 

Moejoe90:
This fixes the issue by adding the token information to the `pending_transactions` api


packages/web/app/modules/tx/monitor/sagas.ts
```diff
@@ -107,14 +107,18 @@ export function* ensurePendingTransactionSchemaIsValid(
...
         (pendingTransaction.transactionAdditionalData &&
           (pendingTransaction.transactionAdditionalData.to === undefined ||
             pendingTransaction.transactionAdditionalData.amount === undefined ||
             pendingTransaction.transactionAdditionalData.amountEur === undefined ||
             pendingTransaction.transactionAdditionalData.totalEur === undefined ||
-            pendingTransaction.transactionAdditionalData.total === undefined))
+            pendingTransaction.transactionAdditionalData.total === undefined ||
+            pendingTransaction.transactionAdditionalData.tokenDecimals === undefined ||
+            pendingTransaction.transactionAdditionalData.tokenImage === undefined ||
+            pendingTransaction.transactionAdditionalData.tokenSymbol === undefined))
```
drakkein: This is getting messy, but if it's a temporary fix, then lets leave it for now

Moejoe90: We need to create some migration mechanisim. 
There's already a ticket for that but you're totally right it shouldn't stay like this

Issue #3920, opened 2020-01-16
Create the new text field components and add all cases to storybook


Moejoe90:
- [ ] The new components shouldn't replace the old ones. 
- [ ] The new text field components should be in one folder 
- [ ] Use the style guide mentioned in the epic ticket #3775
- [ ] Add all cases to storybook using Atom/Particle Appoach ask @desfero if you need help with that

PR #3931, opened 2020-01-20
#3920 Create nds text input

drakkein:


packages/web/app/styles/mixins/typography.scss
```diff
@@ -151,6 +151,28 @@
   line-height: 1.71;
 }
 
+@mixin nds-input-text {
```
Moejoe90: are these new typography styles? if so are they in the style guide?

packages/web/app/components/shared/inputs/InputError.module.scss
```diff
@@ -0,0 +1,7 @@
+.error-message {
+  @include nds-input-label-small;
+
+  color: $neu-flamingo-red;
+  text-transform: none;
+  margin: 4px 0;
```
Moejoe90: can you make this `rem`

packages/web/app/components/shared/inputs/InputDescription.module.scss
```diff
@@ -0,0 +1,7 @@
+.description {
+  @include nds-input-label-small;
+
+  color: $neu-gull-gray;
+  text-transform: none;
+  margin: 4px 0;
```
Moejoe90: use `px-to-rem` please

packages/web/app/components/shared/inputs/Field.tsx
```diff
@@ -0,0 +1,53 @@
...
+  name: string;
+  label?: string;
+  description?: string;
+  isValid?: boolean;
+  required?: boolean;
+  disabled?: boolean;
+  placeholder?: string;
+};
+//TODO: connect to formik after update
+export const FieldLayout: React.FunctionComponent<TFieldProps &
```
Moejoe90: Please don't wait for the update. 
We need to use this component in its connected Form use `FormInput.tsx` as a guide on how to use `FormikConsumer` etc.

packages/web/app/styles/mixins/typography.scss
drakkein: These styles are missing from NDS Text styles catalog, but are present in inputs styleguide

Moejoe90: Then we need to talk with @andywalton about this

drakkein:
@glupiawymowka Error message not displaying has been fixed, sorry for your lost time.

Issue #3691, opened 2019-10-31
Update bootstrap to the latest


desfero:
**Background**
Right now we are a couple of major versions behind bootstrap. 
The biggest point is that we are not able to use `Link` inside tooltips because it's not a portal (fixed in 7.x).

PR #3953, opened 2020-01-26
#3691 Update bootstrap

drakkein:


packages/web/app/components/shared/tooltips/TooltipBase.tsx
```diff
@@ -22,10 +22,7 @@ interface IState {
   tooltipOpen: boolean;
 }
 
-export class TooltipBase extends React.Component<
-  IProps & TooltipProps<{ hideArrow?: boolean }>,
```
drakkein: TooltipProps are not generic type any more

packages/web/app/styles/overrides.scss
```diff
@@ -49,6 +49,7 @@ textarea {
   color: $neu-greyish-brown;
   font-size: px-to-rem(20px);
   padding: 0 px-to-rem(10px) px-to-rem(4px);
+  height: calc(2.25rem + 2px);
```
drakkein: In the new version this was changed to (1.5**em** + 0.75**rem** + 2px), I'm setting the previous version

packages/web/app/components/shared/tooltips/Tooltip.module.scss
```diff
@@ -19,6 +19,8 @@ $icon-size: 18px;
         display: none;
       }
 
+      display: block;
```
drakkein: Without setting tooltip anchor to be displayed as block (we need to have correct dimensions) tooltip is positioned weirdly.

PR #3960, opened 2020-01-28
HOTFIX - Fixes payouts view logic

Moejoe90:


packages/web/app/components/shared/formatters/MoneyWithLessThan.tsx
```diff
@@ -0,0 +1,52 @@
...
+  //TODO: Refactor Money + Formatters to reuse the components
+  return fixedZeroPricision ===
+    toFixedPrecision({
+      value,
+      inputFormat: ENumberInputFormat.ULPS,
+      decimalPlaces: selectDecimalPlaces(valueType),
+      roundingMode: ERoundingMode.HALF_UP,
+    }) && !new BigNumber(value).isZero() ? (
+    <span className={cn(styles.money, props.className, props.theme)}>
+      &#60; {new BigNumber("10").toPower(-selectDecimalPlaces(valueType)).toString()}
```
drakkein: Yeah, this one definitely needs to be refactored, but as we need this ASAP then it's ok

Moejoe90: Actually IMO this component shouldn't exist. Money should be `smart enough` to do this on its own but for now as a quick fix I will keep it like this and we need to create a refactor for the formatters and Money

PR #3961, opened 2020-01-28
Changes to ETO View

arjunu:


packages/web/app/components/eto/public-view/CoverBanner.tsx
```diff
@@ -46,25 +50,39 @@ const BannerBase: React.FunctionComponent<TDataTestId> = ({
...
+  let messageId;
+
+  if ([EEtoState.PREVIEW, EEtoState.PENDING, EEtoState.LISTED].includes(eto.state)) {
+    messageId =
+      jurisdiction === EJurisdiction.GERMANY
+        ? `eto-overview.cover-banner.jurisdiction.de.to-be`
+        : `eto-overview.cover-banner.jurisdiction.li.to-be`;
+  } else if (
+    eto.state === EEtoState.PROSPECTUS_APPROVED ||
+    (isOnChain(eto) && etoStateOnChain && etoStateOnChain < EETOStateOnChain.Signing)
```
rudolfix: looks good, however you can just use `eto.state === EEtoState.ON_CHAIN` instead of `isOnChain`, second one checks presence of loaded contract which you do not really need here.

`etoStateOnChain` may be zero! in that case this expression will not return `true` which is wrong!
btw. it's guaranteed to have no null/None etoStateOnChain if eto.state === EEtoState.ON_CHAIN

packages/web/app/components/eto/shared/EtoView.tsx
```diff
@@ -104,9 +111,21 @@ const EtoViewCampaignOverview: React.FunctionComponent<IProps> = ({ eto, isUserF
       ? EColumnSpan.TWO_COL
       : EColumnSpan.THREE_COL;
 
+  const showTimeline =
+    eto.state > EEtoState.LISTED &&
```
rudolfix: this will not work! AFAIK you have states represented as strings and alphabetic order is not the same as order of states. maybe there's ordered list somewhere where you can compare indexes?

packages/web/app/components/eto/shared/EtoView.tsx
```diff
@@ -104,9 +111,21 @@ const EtoViewCampaignOverview: React.FunctionComponent<IProps> = ({ eto, isUserF
...
+    !!getETODocument(eto, EEtoDocumentType.APPROVED_INVESTOR_OFFERING_DOCUMENT);
+
+  const isSupportedJurisdiction = [EJurisdiction.GERMANY, EJurisdiction.LIECHTENSTEIN].includes(
+    eto.product.jurisdiction,
+  );
+  const showDisclaimer =
```
rudolfix: 👍

packages/web/app/components/eto/shared/EtoView.tsx
```diff
@@ -104,9 +111,21 @@ const EtoViewCampaignOverview: React.FunctionComponent<IProps> = ({ eto, isUserF
       ? EColumnSpan.TWO_COL
       : EColumnSpan.THREE_COL;
 
+  const showTimeline =
+    eto.state > EEtoState.LISTED &&
+    !!getETODocument(eto, EEtoDocumentType.APPROVED_INVESTOR_OFFERING_DOCUMENT);
```
rudolfix: you do not need to check it. show timeline in LISTED, PROSPECTUS_APPROVED and ON_CHAIN states

packages/web/app/components/eto/public-view/CoverBanner.tsx
Moejoe90: Can you also for readability make `isOnChain(eto) && etoStateOnChain && etoStateOnChain < EETOStateOnChain.Signing`  as constant with a name that idicates the condition

packages/web/app/components/eto/public-view/CoverBanner.tsx
```diff
@@ -46,25 +50,39 @@ const BannerBase: React.FunctionComponent<TDataTestId> = ({
...
-      return null;
+export const InvestorBannerLayout: React.FunctionComponent<IJurisdictionBannerProps &
+  IExternalProps> = ({ eto, jurisdiction, etoStateOnChain }) => {
+  if (jurisdiction !== EJurisdiction.GERMANY && jurisdiction !== EJurisdiction.LIECHTENSTEIN) {
+    return null;
   }
+
+  let messageId;
+
+  if ([EEtoState.PREVIEW, EEtoState.PENDING, EEtoState.LISTED].includes(eto.state)) {
```
Moejoe90: can you make `[EEtoState.PREVIEW, EEtoState.PENDING, EEtoState.LISTED].includes(eto.state)` and constant with a readable name

packages/web/app/components/eto/shared/EtoView.tsx
Moejoe90: @rudolfix the orginial ticket says  
```
The timeline needs to be hidden on the Marketing listing and visible on the states starting from the Investment Offer listing (prospectus/investment offer uploaded)
```
that the document should be uploaded

packages/web/app/components/eto/public-view/CoverBanner.tsx
arjunu: defined as `toBeOffered`

packages/web/app/components/eto/public-view/CoverBanner.tsx
arjunu: now defined as `currentlyOffered`

arjunu: Since we have Typescript strict non null check, I added `EETOStateOnChain.Setup` as a fallback value:
```js
(eto.state === EEtoState.ON_CHAIN &&
      (etoStateOnChain || EETOStateOnChain.Setup) < EETOStateOnChain.Signing)
```

packages/web/app/components/eto/shared/EtoView.tsx
arjunu: updated the condition

packages/web/app/components/eto/shared/EtoView.tsx
arjunu: updated the condition

.circleci/config.yml
```diff
@@ -181,7 +181,6 @@ workflows:
       - codechecks:
           requires:
             - build-frontend
-            - build-docker-test-e2e
```
rudolfix: is it OK?

Moejoe90: yes this is on purpose as part of the plan to remove the dependancy of `e2e` results on running the codechecks. 

So now even if the tests fail on `e2e` level QA could still test the feature via code checks

PR #3965, opened 2020-01-29
Fixes Payout widget

Moejoe90:


Moejoe90:
Will merge without QA as this can't be tested on `io`

PR #3986, opened 2020-02-05
Update wallet readme to mention `pod install` before starting IOS app

desfero:


DmitriyNoa:
Awesome. Now it works. Thanks for adding this to README

Issue #3892, opened 2020-01-12
Check Treeshaking for the new packages


Moejoe90:
**Background**
While creating new packages we need to make sure that tree shaking still works to not load anything we don't use into the bundle.

You can use the `Bundle-Analyzer` to analyse the bundle and check for treeshaking and deadcode elimination for the seperate packages. this includes 

- [x] `@neufund/sagas`

- [x] `@neufund/shared`

**Designs**

**Acceptance criteria**

**Technical notes**

**Testing notes**

PR #3987, opened 2020-02-05
#3892 Fix tree shaking for sagas packages

desfero:
This pr fixes tree shaking for `sagas` package as now `redux-saga-test-plan` package is included in the final production bundle.

Also marks `sagas` and `shared` as `sideEffects` free to three shake imports if they are not used.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "workspaces": [
-    "packages/**"
+    "packages/*"
```
desfero: we don't expect to have packages deeply in the folder structure so can simplify the glob to just pull single level deep

packages/sagas/src/tests.ts
```diff
@@ -0,0 +1,5 @@
+export * from "redux-saga-test-plan";
```
arjunu: why not `sagas/src/modules/tests`?

packages/web/app/modules/sagasUtils.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { cancelled, delay, expectSaga, put } from "@neufund/sagas";
+import { cancelled, delay, put } from "@neufund/sagas";
```
arjunu: why not `@neufund/saga/sagaEffects`?

packages/sagas/src/tests.ts
desfero: just to keep it as close as possible to `index.ts` as it's a top-level export. In general, we should also extract `sagaEffects` from `modules` folder as it's just a utility IMO.

packages/web/app/modules/sagasUtils.spec.ts
desfero: so now we have `@neufund/sagas` which contain utility functions related to saga (wraps `redux-saga` and `typed-sagas` packages and `@neufund/sagas/tests` which only provides test-related imports. Is there any reason to import as `@neufund/sagas/sagaEffects`?

arjunu: Just so that when we add more modules here, tree shaking works for them too.

package.json
arjunu: did you notice any perf improvements after this for running `yarn`?

Issue #3821, opened 2019-12-06
Button fixes to match button states to the latest as per NDS


andywalton:
Hey, we've made a few small tweaks to the button designs.

This tickets task is to match the hover button states for the:
1. `.btn-primary` - Primary button 
2. `.btn-secondary` - Secondary button (formerly called Outline)
3. `.btn-link` - Link button (formerly called Ghost)

The hover state and active states are the same and they use our darker brand yellow with our base gray for button labels and icons:
```
$yellow-darker-1
#C6D30D
```

The text and icons are the same as the button label:
```
$base-gray
#142431
```

Also note, there are four height size modifiers for buttons:
1. Default = 40px height 
2. Extra small = 24px height 
3. Small = 32px height 
4. Large = 48px height

We also have a circular button - this is already correctly developed and it's only used in our Documents section. It however can be used at any of the above sizes and it's hover state and disabled states would be the same as the other buttons listed above. 

**Designs:**
View the below in Zeplin: https://zpl.io/agBnxDA

The buttons should look like the below at each state changes:

![image.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/973e15a5-32c0-4de4-bda5-2b43a979c684)

![image.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/8e16cdaa-db11-4c6a-8925-913d5b77ef29)

![image.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/f5026729-f2ea-436e-aee4-e6acc40f59d9)

---

![image.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/9c9f7201-e333-4503-ba75-cb6534134305)

Icons, a button can either have a leading icon or a trailing icon, or only an icon. The space between the icon and button label should be `8px` and the icon colour should always match the label colour. So when it's disabled, the button label changes to a different gray, so should the icon which should match it.

---

The main fix for this ticket please is to match...
- [ ] the hover states - as described above, currently we have an incorrect button hover states on the platform. For example, the primary button has a drop shadow which should rather be removed.

Once that is done this ticket can be pushed for testing.

The rest is a bonus, if you do anything else, please just mention it in the PR.

Thanks,
Andy

glupiawymowka:
moving back to in progress due to some possible build issues

glupiawymowka:
This ticket was merged without the full test and it came out that it's only partly done. 
[...]

PR #4015, opened 2020-02-10
#3821 Button states CSS fixes

drakkein:
This PR fixes styling for hover and focussed states in our buttons styles. 

Changes could be verified via Storybook.

_Button primary_
image

_Button outline (secondary)_
image

_Button ghost (link)_
image

Also added new extra small size for all variations of buttons
image

And changed styling for circle button
image


packages/design-system/src/components/buttons/Button.module.scss
```diff
@@ -19,21 +21,21 @@
     color: $neu-gull-gray;
   }
 
+  &:focus {
```
desfero: let's not use `:focus` given we have already a polyfill for `focus-visible` included. You can check how focus was managed previously to see the differences

packages/design-system/.storybook/config.js
```diff
@@ -3,6 +3,8 @@ import { addParameters, addDecorator } from "@storybook/react";
 import { withA11y } from "@storybook/addon-a11y";
 import { addReadme } from "storybook-readme";
 
+import "focus-visible";
```
desfero: Given that `focus-visible` is only used inside `design-system` package maybe we should do a `polyfills` file here and import it in the `web` app polyfills? If you think it's reasonable then we also need to add `sideEffects` property in package.json `  "sideEffects": [
    "*.sideEffect.*",
    "*.scss",
    "*.css"
  ],` for the design-system package and check if the polyfill is apllyied properly for the production build

packages/web/webpack/webpack.config.ci.prod.js
```diff
@@ -22,6 +22,7 @@ module.exports = merge.smart(configCommon, {
     },
     runtimeChunk: true,
     minimize: false,
+    sideEffects: false,
```
drakkein: hmmm.. somehow our `focus-visible` polyfill is being removed after optimization

Moejoe90:
Blocked for now until https://github.com/Neufund/platform-frontend/pull/4133/

packages/web/webpack/webpack.config.ci.prod.js
arjunu: removed

Issue #3991, opened 2020-02-06
Add navigation library to mobile app


desfero:
**Background**
Given that we have wireframes now we can progress further with the initial setup for the mobile app.

Please add a navigation library and a couple of mock routes (we know how the app gonna look from the wireframes) just for us to know how we should use it.

**Technical notes**
Also at the moment, react-native community is split by `react navigation` and `react native navigation` so please research which one is better for us.

**Testing notes**

DmitriyNoa:
I heard positive recommendations on the Airbnb navigation library https://github.com/airbnb/native-navigation. Also to consider.

desfero:
After some research and going through the docs `react-nativation` seems the most promising library for us:

My thoughts about the alternatives
`native-navigation` - completely unmaintained from 2018 and even marked as not production-ready.
`react-native-navigation` - a good one as navigation is handled natively so it's quite performant but there is quite a small amount of customization options. 
`react-navigation` - the one below by community and the most popular one with support for a ton of custom customizations.
`react-router-native` - a native version of `react-router`, this one is not suited for the proper native navigation as the idea here is to mimic the web routing in the native app which is not the same.

From npm trends `react-navigation` is clearly the most popular https://www.npmtrends.com/react-navigation-vs-react-native-navigation-vs-native-navigation-vs-react-router-native

PR #4016, opened 2020-02-11
#3991 Add navigation library to mobile

desfero:
**Technical notes**
This PR connects wallet with `react-navigation` library with some example routing. 
Please check the comment in a ticket with brief summary of why `react-navigation` was finally chosen as a library to go.


packages/wallet/package.json
```diff
@@ -13,8 +13,15 @@
   },
   "dependencies": {
     "@neufund/shared": "^1.0.0",
-    "react": "^16.12.0",
+    "@react-native-community/masked-view": "^0.1.6",
+    "@react-navigation/native": "^5.0.1",
+    "@react-navigation/stack": "^5.0.1",
+    "react": "16.9.0",
```
desfero: `react` was moved back to the version compatible with current `react-native` version

DmitriyNoa:
Tested on IOS and Android emulators. Works as expected.

packages/wallet/app/appRoutes.ts
```diff
@@ -0,0 +1,7 @@
+const appRoutes = {
+  Landing: "Landing",
```
DmitriyNoa: Maybe let's keep constants names capitalized as we have in in web package? E.g. APP_ROUTES, LANDING?

packages/wallet/index.js
```diff
@@ -1,9 +1,7 @@
-/**
```
DmitriyNoa: Is it possible to move this to TS?

packages/wallet/app/appRoutes.ts
desfero: It depends where the constants are. I see `appRoutes` on the web is also camelCased, same with `externalRoutes`. Maybe let's, for now, stick with the web convention and then later decide if we want to refactor as it's just simple rename :)

packages/wallet/index.js
desfero: makes sense to keep consistency, fixed

PR #4017, opened 2020-02-11
Setup Docs for Storybook

arjunu:
image

Uses addon https://github.com/tuchk4/storybook-readme. 

Note: Instead of docs in the panel, we can also have the API with the stories (see https://tuchk4.github.io/storybook-readme/?path=/story/button--button). 

Why I like this:
- documentation will be in markdown and lives in a README file 
- it can also be read by directly viewing the file

Other changes:
- Simplified button stories as there were too many combinations and they were all in separate pages. Now they have been grouped together.
- Elevated Button to "Molecules"


packages/design-system/src/components/buttons/Button.stories.tsx
```diff
@@ -1,223 +1,161 @@
...
-      isLoading={true}
-      iconProps={{ alt: "Do something" }}
-    />
-  </>
-);
+const GenerateStory: React.FunctionComponent<TGenerateStoryProps & TButtonProps> = ({
+  title,
+  ...props
+}) => {
+  const Wrapper = props.width === EButtonWidth.BLOCK ? BlockWrapper : InlineBlockWrapper;
```
Moejoe90: are there any benefits in doing this instead of directly using the components for each story?

arjunu: no, this is just to reduce code duplication

Issue #3993, opened 2020-02-06
Add e2e testing library for mobile app


desfero:
**Background**
Let's research how we want to write e2e tests for the mobile app and how to connect them and run on multiple devices during CI build. At the moment in the community, two major frameworks are used `appium` and `detox`, we need to find which one is better for us and also which one provides the ability to run on multiple devices.

**Designs**

**Acceptance criteria**

**Technical notes**

**Testing notes**

PR #4021, opened 2020-02-12
#3993 Add e2e tests library to mobile app

desfero:
**Technical notes**
This PR just add e2e tests configuration, CI connection will come as a separate pr soon. 
Please make sure that the manual setup docs are working for you, maybe there is something missing. if you encounter any problems let me know we can research the reason together.


packages/wallet/e2e/import-wallet/ImportWalletFromPrivateKey.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { device, element, by, expect } from "detox";
```
DmitriyNoa: Here is something I have noticed in our current WEB setup. As all e2e tests are in a separate folder it's often not obvious to find a specific test. It would be great if we have follow more strict rules about where to put and how to name e2e files and tests, maybe as a part of readme.

Or maybe we can keep along with module files., inside the Import wallet module.

What do you think?

packages/wallet/README.md
```diff
@@ -55,3 +55,47 @@ To fix automatically formatting issues use
...
+### IOS setup
+
+#### 1. Add [applesimutils](https://github.com/wix/AppleSimulatorUtils)
+
+Install with [homebrew](http://brew.sh/) the latest version of applesimutil used by detox to
+communicate with the simulator.
+
+```bash
+brew tap wix/brew
+brew install applesimutil
```
DmitriyNoa: Trying to install this, so far - Error: No formulae found in taps.

DmitriyNoa: Ahh - it's a typo - brew install applesimutil**s**, please fix

packages/wallet/README.md
```diff
@@ -55,3 +55,47 @@ To fix automatically formatting issues use
...
+```
+
+#### 2. Make sure that the simulator is available on your machine (was installed by Xcode).
+
+List all available devices with `applesimutils --list` and check whether simulator specified in
+`./package.json` `detox.configuration["ios.sim.debug"]device.type` is on the list. If it's not
```
DmitriyNoa: How about moving this config from package json to env vars later?

packages/wallet/README.md
```diff
@@ -55,3 +55,47 @@ To fix automatically formatting issues use
 ```bash
 yarn format:fix
 ```
+
+#### Run e2e tests on ios or android emulator
+
+Before running e2e tests make sure you have followed **E2E tests setup** guide below.
+
+```bash
+yarn e2e:ios
```
DmitriyNoa: IOS e2e work as expected

packages/wallet/README.md
```diff
@@ -55,3 +55,47 @@ To fix automatically formatting issues use
...
+yarn e2e:android-debug
```
DmitriyNoa: Android e2e debug failed with:
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'wallet'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not download kotlin-compiler-embeddable.jar (org.jetbrains.kotlin:kotlin-compiler-embeddable:1.3.10)
      > Could not get resource 'https://jcenter.bintray.com/org/jetbrains/kotlin/kotlin-compiler-embeddable/1.3.10/kotlin-compiler-embeddable-1.3.10.jar'.
         > Connection reset
```

packages/wallet/README.md
```diff
@@ -55,3 +55,47 @@ To fix automatically formatting issues use
...
+yarn e2e:android-release
```
DmitriyNoa: Android e2e release works as expected

packages/wallet/README.md
DmitriyNoa: Interesting - when running the debug version second time, with an android simulator open (simulator that left after running the release e2e test command) tests are working.  

I closed the simulator and removed the build folder, then running debug command worked as expected. Looks like it fails only the first time

DmitriyNoa: Can you specify somewhere - why do we need 2 e2e scripts for Android and only one for IOS?

packages/wallet/e2e/init.ts
```diff
@@ -0,0 +1,23 @@
+import detox from "detox";
+import adapter from "detox/runners/jest/adapter";
+
+import { detox as config } from "../package.json";
+
+// Set the default timeout
+jest.setTimeout(120000);
```
DmitriyNoa: 12 seconds timeout? What is it waiting for?

packages/wallet/e2e/init.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// Typings for jasmine are not properly defined in jest
+(jasmine as any).getEnv().addReporter(adapter);
+
+beforeAll(async () => {
+  await detox.init(config, { initGlobals: false });
```
DmitriyNoa: Same here, do you know why is it so long to wait here?

packages/wallet/README.md
desfero: Strange, good to know that it's working. Let's see whether that kind of error appears again.

desfero: Hm, I think we should add a second script for ios too to build ios for Release during CI. Will add the second ios script in the PR connecting e2e to CircleCi pipeline. Is that okay for you?

packages/wallet/e2e/import-wallet/ImportWalletFromPrivateKey.spec.ts
desfero: You are completely right here, web e2e test structure is really hard to follow. From the naming point, i think it's gonna be on QA as they will provide the test spec cases we need to cover so I hope they gonna provide a proper naming convention related to the domain e2e tests are testing. @pavelneu already did some changes to the web e2e structure in #4033 so then we can to the point of writing e2e tests we can sync the approach with QA team. What do you think? Is that a reasonable way to go from your point?

DmitriyNoa: Good idea, @pavelneu it might make sense for you to jump into this PR and check it, so we know we are on the right way from the start.

packages/wallet/README.md
desfero: fixed

packages/wallet/README.md
desfero: Done, there is now separate script for ios Debug and Release mode

packages/wallet/README.md
desfero: That's a very good point. I did some quick research and for now, it's not possible, see https://github.com/wix/Detox/issues/1238

packages/wallet/e2e/init.ts
desfero: These are the values suggested by the detox team. On my machine 10s timeout is enough but on slower CPU it may take longer to initialize detox so I think it's just a safe value that seems to fit a wide range of machines

packages/wallet/e2e/init.ts
desfero: same as above

packages/wallet/README.md
```diff
@@ -55,3 +55,47 @@ To fix automatically formatting issues use
 ```bash
 yarn format:fix
 ```
+
+#### Run e2e tests on ios or android emulator
```
DmitriyNoa: I thought maybe let's keep main README as a list/table of contents with links to relevant README, as if we keep writing good docs this may grow significantly.
So in this example - move all e2e related to /e2e/READ.md
What do you think?

packages/wallet/README.md
DmitriyNoa: This looks outdated. Shouldn't it be yarn e2e:ios-debug or ...:android?

packages/wallet/README.md
desfero: It's good point to keep root readme clean. Whole e2e tests setup section was moved to `./tests/e2e/README.md`.

DmitriyNoa:
Tested on clean installation. Following scripts work:
```
yarn e2e:ios-debug
yarn e2e:ios-release
yarn e2e:android-release
```

Only one that fails is:
```
yarn e2e:android-debug
```
Here is the failure:
Screenshot 2020-02-17 at 17 51 58

DmitriyNoa:
After playing around I figured out that this happens if you run one android release test after android debug test, or wise-versa.
If I clean the build manually with commands belove everything work.
```
rm -rf android/build
rm -rf android/.gradle
``` 
Maybe we can add this to run before
```
yarn e2e:build:android-debug
yarn e2e:build:android-release
```
so we force clean it everytime

PR #4022, opened 2020-02-12
[WIP] 3992 mobile wallet storybook

DmitriyNoa:
This PR adds the storybook configuration. The CI pipeline will be added in a separate PR


packages/wallet/storybook/storyLoader.js
```diff
@@ -0,0 +1,19 @@
+// Auto-generated file created by react-native-storybook-loader
```
DmitriyNoa: This is a sample of auto-generated storybbok loader. Later it will not be included in the code and added to .gitigniro

packages/wallet/storybook/index.ts
```diff
@@ -0,0 +1,20 @@
+import { AppRegistry } from 'react-native';
+import { getStorybookUI, configure } from '@storybook/react-native';
+
+import './rn-addons';
+// Load storybook
+
+// as per official Storybook documentation React native doesn't not support dynamic story loading (https://storybook.js.org/docs/basics/writing-stories/#loading-stories) we have to use react-native-storybook-loader to generate stories list vis cli
+
+import { loadStories } from './storyLoader';
```
DmitriyNoa: As per official Storybook documentation React native doesn't support dynamic story loading (https://storybook.js.org/docs/basics/writing-stories/#loading-stories) we have to use react-native-storybook-loader to generate stories list via cli. The prestorybook script will generate the storyLoader file every time before the storybook is launched.

As an alternative solution, we can:
- Store all the stories in a single directory.
- Just import all the stories manually.

PR #4031, opened 2020-02-13
Fixes footer Responsivnes

Moejoe90:
fixes this 
image


andywalton:
@Moejoe90 @arjunu @AlexeyTuboltsev 

I'd add to the above request -> please simply change the CSS so that the height isn't fixed at 64px. Make it have `min-height: 64px;`

For the mobile breakpoint, please also change the padding left and right to be `20px` because it's currently set at `5rem` which is far to much and forces the wrapping of our copyright statement onto a 3rd line when the screen is resized to a width of 320px.

That should sort it (even if it wraps onto a 3rd line):
Screenshot 2020-02-13 14 24 50

PR #4039, opened 2020-02-14
Add missing PEP and income self-declaration field back to KYC individual flow

arjunu:


packages/web/app/components/kyc/personal/Address.tsx
```diff
@@ -67,7 +68,7 @@ const KYCForm: React.FunctionComponent<TProps> = ({
     <>
       <KycStep
         step={3}
-        allSteps={5}
+        allSteps={TOTAL_STEPS_PERSONAL_KYC}
```
Moejoe90: :+1:

packages/web/app/lib/api/kyc/KycApi.interfaces.ts
```diff
@@ -141,8 +141,22 @@ export const KycIdNowIdentificationSchema = YupTS.object({
 export type TKycIdNowIdentification = YupTS.TypeOf<typeof KycIdNowIdentificationSchema>;
 
 export const KycPersonalDataSchemaRequiredWithAdditionalData = KycPersonalDataSchemaRequired.concat(
-  KycAdditionalDataSchema,
-);
+  Yup.object().shape({
+    isPoliticallyExposed: Yup.string().required(),
+  }),
```
Moejoe90: why wouldn't this be one once shape that you can concat? 
Then you don't need to repeat 

```
 Yup.object().shape({
);	    isPoliticallyExposed: Yup.string().required(),
  }),
```

packages/web/app/modules/kyc/financial-disclosure/sagas.ts
```diff
@@ -0,0 +1,28 @@
...
+import { createMessage } from "../../../components/translatedMessages/utils";
+import { TGlobalDependencies } from "../../../di/setupBindings";
+import { actions, TActionFromCreator } from "../../actions";
+import { neuCall } from "../../sagasUtils";
+import { submitPersonalDataSaga } from "../sagas";
+
+export function* submitFinancialDisclosure(
+  { notificationCenter, logger }: TGlobalDependencies,
+  action: TActionFromCreator<typeof actions.kyc.kycSubmitFinancialDisclosure>,
+): Generator<any, any, any> {
```
Moejoe90: please type the generator

packages/web/app/modules/kyc/financial-disclosure/sagas.ts
```diff
@@ -0,0 +1,28 @@
...
+  try {
+    const { data, close } = action.payload;
+    yield neuCall(submitPersonalDataSaga, data);
+
+    if (close) {
```
Moejoe90: we should start slowly getting away from this style of writing sagas. 

The idea should be that once the saga is finished/done `Success` a success event is fired an a listener should react based on that. 

So we have two types of sagas. View sagas and Logic Sagas. 

This could stay like this for now but in the future try to think slowly about moving out from this pattern

arjunu: gotcha

packages/web/app/lib/api/kyc/KycApi.interfaces.ts
arjunu: refactored

arjunu:
Merged with master and made changes to form to reflect new Formik API.

arjunu:
> * [ ]  if the user choses `yes` in the PEP question, this is not being sent to the backoffice.
>   The backoffice shows the default `no` answer
>   Also after moving the request back to `draft` you can login as the user and see that all the answers are saved as they were filled, but the PEP question sets back to `NO`
>   image

Fixed

Issue #4007, opened 2020-02-09
Technical Milestone Mock-up Requirements


rudolfix:
**Background**
Our focus for technical milestone is to come with good internal structure of the app: we need to come up with interfaces/events that components generate, without implementing all the technical details.

#3999 
We start without any encryption at all

#3998 
No encryption, ignore versioning and migrations

#4000 
We can fully implement this, figure out schema of the wallet and use references to key material

#4001 & #4002 
No encryption of keys, we can use LightWallet as web3 provider and cryptography library, we should design decent external interfaces, figure out references to cryptographic material and combine private key storage with cryptographic library

#4003 
we should fully implement notifications

#4004 
we should fully implement this.  There’s react-native lib we can use https://github.com/WalletConnect/walletconnect-monorepo/tree/master/packages/react-native but if we intend to keep it in the final implementation, we need to make sure
* how QR codes are handled
* how we can handle light wallet imports and other extensions with it (like heartbeats, timestamps)
* how it handles notifications
* do a proper code review

#4005 
we should pick up decent QR code lib and fully implement wallet connect QR code parsing

#4006 
we should propose good high level external interface and implement relevant methods

PR #4041, opened 2020-02-16
Connect mobile app with modules

desfero:
**Technical notes**

To progress further with #4007 we need to connect our app with our modules system (sagas and di). 

This pr gives us building blocks we can use to implement a proper abstraction for app storage requirements. 

Also seems I was able to fix `react-native` workspaces linking problems by following and experimenting with the examples from https://github.com/gorhom/react-monorepo.

**New scripts for the mobile app**
`yarn tsc` - compile typescript (at the moment there is something wrong with `react-native` typings will fix in the separate pr)
`yarn start:ios:release and yarn start:android:release` - convenient wrappers to run in release mode


packages/shared-modules/src/extensions.ts
```diff
@@ -0,0 +1,48 @@
+import { IExtension } from "@neufund/sagas";
+import { Container } from "inversify";
+
+import { INeuModule } from "./types";
+
+/**
+ * Load context modules from module libs into global container
+ */
+const getLoadContextExtension = (container: Container): IExtension => ({
+  onModuleAdded: (module: INeuModule<unknown>) => {
```
DmitriyNoa: This and onModuleRemoved repeats many times with almost the same signature. Maybe we can move it out and reuse this code?

packages/shared-modules/src/extensions.ts
```diff
@@ -0,0 +1,48 @@
+import { IExtension } from "@neufund/sagas";
+import { Container } from "inversify";
+
+import { INeuModule } from "./types";
+
+/**
+ * Load context modules from module libs into global container
+ */
+const getLoadContextExtension = (container: Container): IExtension => ({
```
DmitriyNoa: As we have discussed before on the FE discussion meeting on Thursday - let's add some TS docs? What do you think?

packages/shared-modules/src/utils.ts
```diff
@@ -1,5 +1,77 @@
...
 
 const createLibSymbol = <T>(name: string) => Symbol(name) as TLibSymbol<T>;
 
-export { createLibSymbol };
+/**
+ * If react-redux connection function receive `undefined` as a `dispatchToProps` method it then pass `dispatch` method to wrapped component.
+ * Often it's not needed as it leads to problems when wrapped component spread all props to any dom element
+ * @param omitDispatch
+ */
+const getDispatchToProps = (omitDispatch: boolean) => (omitDispatch ? () => ({}) : undefined);
```
DmitriyNoa: Can you explain why do we return anonymous function with empty object?

packages/shared-modules/src/utils.ts
```diff
@@ -1,5 +1,77 @@
...
+    dispatch: (a: GlobalActions | TModuleActionsOrNever<Module>) => void,
+    ownProps: Props,
+  ) => Dispatch;
+  options?: Options<GlobalState, State, Props> & TCustomOptions;
+};
+
+/**
+ * A wrapper around react-redux connect with build in modules support
+ */
+function appConnect<
```
DmitriyNoa: Please provide some TS docs as it has many parameters

packages/wallet/README.md
```diff
@@ -16,12 +16,11 @@ device** (just for android development).
 After environment setup is completed you can start the app in emulator (`yarn start:ios` or
 `yarn start:android`).
 
-_**Note**: For ios before starting the app for the first time `pod`'s should be installed manually._
+_**Note**: For ios before starting the app for the first time and after any changes to the ios
+native modules `pod`'s should be installed manually._
```
DmitriyNoa: Typo after a dot _ ?

packages/wallet/app/di/setupBindings.ts
```diff
@@ -0,0 +1,14 @@
+import { coreModuleApi, TLibSymbolType } from "@neufund/shared-modules";
+import { Container } from "inversify";
```
DmitriyNoa: By the way, what is di stands for, in a folder name?

packages/wallet/app/modules/init/actions.ts
```diff
@@ -0,0 +1,9 @@
+import { createActionFactory } from "@neufund/shared";
+
+export const initActions = {
+  start: createActionFactory("INIT_START"),
```
DmitriyNoa: Let's move all string literals to constants, so we can reuse these later

packages/wallet/app/components/Landing.tsx
```diff
@@ -3,14 +3,34 @@ import { Text, Button, View } from "react-native";
...
+}) => {
   const navigation = useNavigation();
 
+  React.useEffect(() => {
+    init();
+  }, []);
+
   return (
     <View>
       <Text>Landing</Text>
```
DmitriyNoa: I know it's only a stub view, but we need to integrate with some internalization/localization library and see if we can reuse localization files from  web

packages/wallet/app/store/create.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return createStore(
+    {
+      extensions: [
+        getLoadContextExtension(context.container),
+        getContextToDepsExtension(appModule, createGlobalDependencies, context),
+        getSagaExtension(context),
+      ],
+    },
+    setupCoreModule({ backendRootUrl: "fsafa" }),
```
DmitriyNoa: fsafa ? ) In master we already have react-native-config, to pickup env vars, so we can use it to prove the value. Or just add a todo, so we don't forget to clean it

packages/wallet/app/store/types.ts
```diff
@@ -0,0 +1,26 @@
+import { DeepReadonly, Values } from "@neufund/shared";
+import { ActionCreatorsMapObject, Reducer, ReducersMapObject } from "redux";
+
+import { appReducers } from "../modules/reducers";
+
+type TReducersMapToState<T extends ReducersMapObject<any, any>> = T extends ReducersMapObject<
```
DmitriyNoa: Is it possible to replace any ?

packages/wallet/app/store/utils.ts
```diff
@@ -0,0 +1,26 @@
+// helper to use instead of redux connect. It's bound with our app state and it uses dictionary to pass arguments
+import {
+  appConnect as sharedAppConnect,
+  TAppConnectOptions,
+  TModuleSetup,
+} from "@neufund/shared-modules";
+import { InferableComponentEnhancerWithProps } from "react-redux";
+import { TAppGlobalActions, TAppGlobalState } from "./types";
+
+export function appConnect<
```
DmitriyNoa: same here. Lets DOC all utils

packages/shared-modules/src/utils.ts
desfero: hm, it has just one which is an object

packages/wallet/README.md
desfero: fixed

packages/wallet/app/di/setupBindings.ts
desfero: it stands for `dependency injection`. I'm thinking whether we can move the content to `libs` folder as its storing all services but for now, that's at least consistent with the `web` app

packages/wallet/app/components/Landing.tsx
desfero: agree, will add a ticket today to research whether `react-intl` is still an option for react-native project or maybe we need to use something else

packages/wallet/app/store/create.tsx
desfero: good catch, fixed

packages/wallet/app/store/types.ts
desfero: Replaced by `object`

packages/wallet/app/store/utils.ts
desfero: fixed

DmitriyNoa:
Great job. + 100500 for adding TSdocs and extensive comments.

- [x] All e2e tests are passing locally.
- [x] App works.

PR #4046, opened 2020-02-17
Running cypress tests with @tags

pavelneu:
This workaround brings the desired behaviour of running automation tests which marked with specific tags back from mocha.js .

The immediately invoked function expression will overwrite the `it` function of Cypress if tags are provided in `Cypress.env("Tags")`.

```
cypress run --env Tags=@p1,@wallet
```
will run all tests BUT the tests without the `@p1` and `@wallet` tags will be skipped (the other ones with that tag will run usual way)

NB: The tests which are skipped will be marked as Pending in Cypress run


packages/web/cypress/support/filterTestsByTags.js
```diff
@@ -0,0 +1,21 @@
+//The issue/workaround: https://github.com/cypress-io/cypress/issues/1865#issuecomment-585185300
```
Moejoe90: :+1:

arjunu:
@pavelneu can you add a note about this it to `app/packages/web/README.md` under "Running tests"?
Maybe also a table of tags that you are using

PR #4058, opened 2020-02-18
Fix KYC header close icon

arjunu:
Before:
image

After:
image


packages/web/app/components/layouts/header/Header.module.scss
```diff
@@ -136,6 +136,11 @@
   margin-left: 1rem;
 }
 
+.action-button-icon {
+  width: 16px;
```
Moejoe90: can you use `px-to-rem`

Issue #4029, opened 2020-02-12
[Portfolio view] Create new components for storybook


andywalton:
**Background**
This ticket is part of the Portfolio view update epic #4028. The task here is to create the new components require to make the table changes required.

**Designs**
Please review the design examples in Zeplin of the Portfolio view and it's various states: populated with data, loading, empty and error states: https://zpl.io/b6m6OBy

Once you've done that, head over to our Zeplin Neufund Design System style guide and look at the components in the section titled: **Data Tables**

**Newfund Design System -> Data Tables:**
https://app.zeplin.io/styleguide/5dcae025a4cf6cfed74cab61/components?cseid=5e442b9290bcf1b344e3070c

Inside this section there are 4 groups:
<img width="358" alt="Screenshot 2020-02-12 21 29 56" src="https://user-images.githubusercontent.com/609657/74374528-e21f4500-4dde-11ea-87e5-3a6a2819e65f.png">

Each with 3 components inside.

Please create and add these components to Storybook with the relevant required logic to pull data in where necessary which you can borrow from our existing Portfolio view.

Note:
1. We have added in a Total row for the Pending payouts, Reserved assets and Your assets tables.
2. You'll see little down arrows in the table heading labels for Token -> these are for table sorting, please IGNORE this for now and don't implement it. This arrow does not need to be shown.

<img width="277" alt="Screenshot 2020-02-12 21 32 08" src="https://user-images.githubusercontent.com/609657/74374977-c8cac880-4ddf-11ea-8492-2809ec12e0e6.png">

**Acceptance criteria**
- [x] Components have been created and added to Storybook.
- [x] These components are tied to the relevant logic to pull in data for the tables they'll be used with.
- [x] We have new summary total rows.
- [x] Table rows do NOT sort via table heading labels. No sorting please.

**Technical notes**
- [ ] The only tricky part will be to do the summary calculations for each Totals row. This might therefore require backend to return this total unless we calculated it client side.

- [x] Add these new components to our `design-package`

**Testing notes**
- [x] The Storybook components should look as per the designs.

PR #4063, opened 2020-02-18
#4029 Create new tables

drakkein:


packages/design-system/src/components/tables/Table.unsafe.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react";
+import { Column, useTable } from "react-table";
```
Moejoe90: can you lazy load this?

packages/design-system/src/components/tables/Table.unsafe.tsx
```diff
@@ -0,0 +1,75 @@
...
+                ))}
+              </tr>
+            ))}
+          </tfoot>
+        ) : (
+          undefined
+        )}
+      </table>
+    </div>
+  );
```
Moejoe90: The nesting here makes the code readability a bit hard. 

Can you abstract this into a smaller component?

Issue #3942, opened 2020-01-22
Verify Email Popup


Moejoe90:
**Background**
In order to incentivize the user to verify their email we will need to implement a new dialogue box that asks the user to verify their email.

**Designs**
Please view the popup design here:
https://app.zeplin.io/project/5e20533697f09f7ff8d0726d/screen/5e283a812ac9f1555b927758

In short, it is a default modal, with an image, a heading, body copy and a single primary button that 
closes the modal.

Note, copy is subject to change. This will be supplied as a separate ticket, the copy is being reviewed as part of the Welcome email change we will make due to the onboarding changes.

**Acceptance criteria**

- [x] as a user (Investor/Nominee/Issuer) with an unverified email. I will get a dialogue box everytime I login

- [x] as an already logged in user with an unverified email the verify your email dialogue box souldn't show up

**Technical notes**
- The popup should be part of the general `login` saga. This check should happen only once on every login.

- The dialogue box has `close` and a call to action (CTA simply closes the modal). 

- Its ok that the dialogue box doesn't popup again if the user refreshes the page and stays in the dashboard. 

- We should make it possible to display different content Light Wallet users and MetaMask/Ledger users. So based on your wallet type, we might show you a different message.

**Testing notes**
Same as acceptance criteria.

glupiawymowka:
this couldn't be tested on codechecks for the newly introduced issuer's/nominee's email due to the access link. 
Works well for the email change though. 
Tested on mobile safari, firefox and chrome

glupiawymowka:
There is a bug when the user is verifying their email for the first time from the verification link. 
The outcome is that the user sees both "Your email has been verified" together with "please check your inbox" messages

![image](https://user-images.githubusercontent.com/18315589/74826950-c949ee80-530c-11ea-9371-18b266885b32.png)

- [ ] Please also fix the responsiveness on mobile 
![image](https://user-images.githubusercontent.com/18315589/74827058-00200480-530d-11ea-88c8-51e912917535.png)

glupiawymowka:
tested on .io for mobile, chrome, firefox 
both cases of verification flow and missing email verification

PR #4067, opened 2020-02-19
#3942 verify email redirect fix

AlexeyTuboltsev:
this PR 
1. prevents the email reminder from showing up also during the verify email flow
2. fixes a small responsiveness issue
3. adds an e2e test for email reminder


packages/web/app/modules/routing/selectors.ts
```diff
@@ -23,6 +24,16 @@ export const selectRedirectURLFromQueryString = createSelector(selectRouter, (st
   return undefined;
 });
 
+// this is a workaround for #3942 (see QA's comment).
+// TODO refactor after we move to route-based sagas
+export const selectIsVerifyEmailRedirect = createSelector(
```
Moejoe90: perfect case scenario would be to have unit tests for this but this is temporary anyway so we can skip that

packages/web/app/test-e2e/1.long-e2e-tests/2.ChangeEmail.spec.e2e.io.ts
```diff
@@ -89,6 +95,34 @@ describe("Change Email", function(): void {
       loginWithLightWallet(newEmail, DEFAULT_PASSWORD);
 
       assertDashboard();
+      cy.get(tid("unverified-email-reminder-modal")).should("not.exist");
+    });
+
+    it("should not show unverified email reminder when user is verifying their email", () => {
```
pavelneu:
```suggestion
    it("should not show unverified email reminder when user is verifying their email @emailing @p3", () => {
```

Issue #4050, opened 2020-02-17
implement ethereum PK storage and crypto functions disregarding PK storage encryption  and using javascript based cryptography


rudolfix:
**Background**
Implement private key storage & cryptography library as described in #4001. We want to provide `pseudo-secure-enclave` interface where cryptographic operations and cryprto material storage are combined into one.

You need to expose and implement cryptographic functions as specified in requirements tickets, using references to key storage. We need to provide a solid external interface here.

Implement private key storage with async storage disregarding app encryption **but make it ready to be replaced by encrypted storage** - exactly as in #4048 (simple key-value where both are strings!)

Use `ethers` library for ethereum cryptography. It has all you need in form you need. See for example this https://github.com/ethers-io/ethers.js/blob/master/wallet.js

**Acceptance criteria**
* [ ] use `pseudo-secure-enclave` pattern when designing external interface
* [ ] mock private key storage with async storage making it ready for replacement
* [ ] implement all private key touching crypto functions as described #4001
* [ ] expose additional crypto functions as needed (like generate seed from mnemonics, ecverify etc.)

**Technical note**
* [ ] we intend to (almost) fully implement this module in kotlin/swift and merge it with secure storage. we'll update our unit tests when this happens
* [ ] before that happens will replace async storage with keychain based storage as we intend for #4048 that will provide final look & feel in terms of UX before huge effort of porting to native code.
* [ ] private key storage is internal and should not be exposed. no other module should be dependent on it.

**Testing notes**
* [ ] happy path tests for all cryptographic functions using key references. we do not test `ethers` - it's already tested
* [ ] all edge cases and tests for private key storage like: using non existing keys to sign etc.

PR #4084, opened 2020-02-21
#4050 Crypto PK storage and crypto functions 

desfero:
**Technical notes**
An initial implementation of a secure interface with wrappers around ETH operations.

**TODO:**
- [x] - Improve error handling (throw custom errors, catch more errors)
- [x] - Connect eth wrappers with logger
- [x] - Write unit tests
- [x] - Add more tsdoc comments for the external API's
- [x] - Clean up wallet and reference existance checks


packages/wallet/app/lib/eth/EthManager.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(
+    @inject(symbols.config)
+    config: TConfig,
+    @inject(symbols.logger) logger: ILogger,
+  ) {
+    this.config = config;
+    this.logger = logger;
+
+    this.adapter = new EthAdapter(new ethers.providers.JsonRpcProvider(this.config.rpcUrl));
```
DmitriyNoa: Why do you make an instance here manually instead of using injection?

desfero: i will probably migrate it to the factory so we can inject logger but want to avoid injecting `EthAdapter` as an instance to DI as it's an internal detail of EthManager

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -1,12 +1,44 @@
...
 
 import { TGlobalDependencies } from "../../di/setupBindings";
 import { initActions } from "./actions";
 
-export function* initStartSaga({ logger }: TGlobalDependencies): Generator<any, void, any> {
+function* initStartSaga({ logger, ethManager }: TGlobalDependencies): Generator<any, void, any> {
   try {
     // TODO: Provide a proper init flow
 
+    const balance = yield ethManager.getBalance(
```
desfero: Will keep all `ethManager` calls here as an example until we move them to appropriate modules

packages/wallet/app/lib/eth/EthAdapter.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+class EthAdapter {
+  constructor(private provider: providers.Provider) {}
+
+  resolveName(addressOrName: EthereumAddress | EthereumName): Promise<EthereumAddress> {
+    return this.provider.resolveName(addressOrName).then(toEthereumAddress);
+  }
+
+  async getChainId(): Promise<number> {
+    const network = await this.provider.getNetwork();
```
rudolfix: so we plan to use `ethers` also as a gateway to ethereum node?

packages/wallet/app/lib/eth/EthManager.ts
```diff
@@ -0,0 +1,168 @@
...
+  private readonly appStorage: AppStorage;
+
+  private wallet: EthWallet | undefined;
+  private readonly walletCreator: EthWalletCreator;
+
+  constructor(
+    @inject(symbols.config)
+    config: TConfig,
+    @inject(symbols.logger) logger: ILogger,
+    @inject(symbols.appStorage) appStorage: AppStorage,
```
rudolfix: 👍

packages/wallet/app/lib/eth/EthManager.ts
```diff
@@ -0,0 +1,168 @@
...
+    this.walletCreator = new EthWalletCreator();
+
+    this.logger.info(`EthManager initialized. Rpc URL: ${this.config.rpcUrl}`);
+  }
+
+  getBalance(address: EthereumAddress): Promise<utils.BigNumber> {
+    return this.adapter.getBalance(address);
+  }
+
+  unsafeGetWalletMetadata = async (): Promise<TSupportedWallets> => {
```
rudolfix: why this is unsafe? getting wallet metadata is perfectly fine. exporting key/secret is unsafe

packages/wallet/app/lib/eth/EthManager.ts
```diff
@@ -0,0 +1,168 @@
...
+import { symbols } from "../../di/symbols";
+import { TConfig } from "../../di/types";
+import { AppStorage } from "../persistance/AppStorage";
+import { toSecureReference } from "../persistance/SecureStorage";
+import { EthAdapter } from "./EthAdapter";
+import { EthWallet } from "./EthWallet";
+import { EthWalletCreator } from "./EthWalletCreator";
+import { TransactionRequestRequired, TSupportedWallets } from "./types";
+
+const WALLET_REFERENCE = "walletReference";
```
rudolfix: I think we have some mixup here. This is just a storage key for a wallet metadata.
the reference is used to work with private keys and seeds

so `WALLET_STORAGE_KEY` is a proper name

packages/wallet/app/lib/persistance/SecureStorage.ts
```diff
@@ -0,0 +1,42 @@
...
+ * @note Current implementation is just an interface. Soon we will replace it by react-native-keychain.
+ *       In long term we hide all secure storage under native code
+ *       to have a control whether key is stored inside device hardware storage
+ *
+ * @internal SecureStorage access should be restricted to just when really needed
+ *           that's why it should be created manually when needed and never connected
+ *           to DI container to prevent harmful access and data leakage
+ */
+@injectable()
+class SecureStorage {
```
rudolfix: 👍 this is exactly what we want

packages/wallet/app/lib/eth/EthWalletCreator.ts
```diff
@@ -0,0 +1,75 @@
+import { EthereumPrivateKey, EthereumHDMnemonic, toEthereumHDPath } from "@neufund/shared";
```
rudolfix: OK this is not a wallet creator. This is our crypto library. It exposes a standard set of crypto operations where instead of private keys we have references. 

Move the metadata operations on wallet somewhere else!

the basic crypto operations are as follows

* signMessage(key reference, message bytes): hash bytes -  sign message represented as bytes with private key with given reference 
* signTransaction(key reference, transaction json): hash bytes - serialize and sign transaction with with private key with given reference 
* import seed(bytes): reference to seed
* import private key(bytes): reference to private key
* import key from seed(reference to seed, derivation path): reference to key
* create random seed/key(): reference to key/seed
* priv key to pub key(reference to key): public key bytes
* export cryptographic material(key reference): key/seed bytes - export secret data with given reference

packages/wallet/app/lib/eth/EthManager.ts
```diff
@@ -0,0 +1,168 @@
...
+  async signMessage(message: string): Promise<string> {
+    const hasExistingWalletMetadata = await this.hasExistingWallet();
+    if (!hasExistingWalletMetadata || !this.wallet) {
+      throw new Error("No wallet connected");
+    }
+
+    return this.wallet.signMessage(message);
+  }
+
+  async signTransaction(transaction: TransactionRequestRequired): Promise<string> {
```
rudolfix: ok so it's almost good but also very wrong.
please decouple wallet metadata from the secret storage!

so what you do: you take wallet metadata from the app storage
you look for private key reference into metadata
you pass this reference to the crypo library as `signTransaction(transaction, key reference)
crypto library has key storage built in that retrieves the key with given reference and signs the tx and returns signed tx

this is how secure enclave API works!

packages/wallet/app/lib/eth/EthAdapter.ts
desfero: Yes, do you see any downsides?

packages/wallet/app/lib/eth/EthWalletFactory.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  async createFromExisting(): Promise<EthWallet> {
+    const walletMetadataJson = await this.appStorage.getValue(WALLET_METADATA_KEY);
+
+    if (walletMetadataJson === null) {
+      throw new Error("Wallet not available");
+    }
+
+    const wallet: TWalletMetadata = JSON.parse(walletMetadataJson);
```
desfero: gonna be migrated to the more typesafe version when app storage is fully done

packages/wallet/app/lib/eth/EthWalletFactory.ts
```diff
@@ -0,0 +1,111 @@
...
+    const seedReference = await this.ethSecureEnclave.addSecret(seed);
+
+    const privateKeyReference = await this.ethSecureEnclave.deriveKey(
+      seedReference,
+      DEFAULT_HD_PATH,
+    );
+
+    const hdWallet: THDWalletMetadata = {
+      type: EWalletType.HD_WALLET,
+      address: "TODO",
```
rudolfix: yeah - we need another method in secure enclave to get
1. public key by reference
2. address by reference - then you can populate it

packages/wallet/app/lib/eth/EthWalletFactory.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    return new EthWallet(wallet);
+  }
+
+  async createFromPrivateKey(privateKey: EthereumPrivateKey): Promise<EthWallet> {
+    const privateKeyReference = await this.ethSecureEnclave.addSecret(privateKey);
+
+    const privateKeyWalletMetadata: TPrivateKeyWalletMetadata = {
+      type: EWalletType.PRIVATE_KEY_WALLET,
+      address: toEthereumAddress(utils.computeAddress(privateKey)),
```
rudolfix: this can go through secure enclave as well

packages/wallet/tsconfig.unit.json
```diff
@@ -6,6 +6,11 @@
     "noUnusedParameters": false,
     "types": ["jest"]
   },
-  "include": ["./app/**/*.spec.*"],
+  "include": [
+    "./typings/**/*",
```
DmitriyNoa: Just nitpicking we can change this to    "./**/*.d.ts", so we can keep typing along with modules, and only store global types in the root "typings" folder

packages/wallet/app/modules/eth/lib/EthAdapter.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /**
+   * Resolves an ENS name to an ETH address
+   *
+   * @param addressOrName
+   *
+   * @returns An ETH address
+   */
+  resolveName(addressOrName: EthereumAddress | EthereumName): Promise<EthereumAddress> {
```
DmitriyNoa: On a WEB a noticed an issue that we have 2 types of ethereum addresses with checksum and without, that differs only in having capital cases:
Non-checksummed version: 0x12ae66cdc592e10b60f9097a7b0d3c59fce29876
Checksummed version: 0x12AE66CDc592e10B60f9097a7b0D3C59fce29876
Maybe we can have a stricter check to use checksum version as it seems to be a preferable option, on the web we have  EthereumAddressWithChecksum type for that. What do you think?

DmitriyNoa: Same probably everywhere we use an address

packages/wallet/app/modules/eth/lib/EthAdapter.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.logger.info(`Resolving a name for ${addressOrName}`);
+
+    return this.provider.resolveName(addressOrName).then(toEthereumAddress);
+  }
+
+  /**
+   * Resolves current chain id
+   */
+  async getChainId(): Promise<number> {
+    this.logger.info("Getting current chain id");
```
DmitriyNoa: Why do we need information loggers?

packages/wallet/app/modules/eth/lib/EthAdapter.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  /**
+   * Returns the current transaction count counting from the latest block tag.
+   *
+   * @param addressOrName - An ethereum address or ENS name
+   */
+  getTransactionCountFromLatest(addressOrName: EthereumAddress | EthereumName): Promise<number> {
+    this.logger.info(`Getting a transaction count from latest for ${addressOrName}`);
+
+    return this.provider.getTransactionCount(addressOrName, "latest");
```
DmitriyNoa: Can we avoid using string literals?

packages/wallet/app/modules/eth/lib/EthManager.ts
```diff
@@ -0,0 +1,248 @@
...
+import { providers, utils } from "ethers";
+import { inject, injectable } from "inversify";
+
+import { privateSymbols } from "./symbols";
+import { EthAdapter } from "./EthAdapter";
+import { EthWallet } from "./EthWallet";
+import { EthWalletFactory } from "./EthWalletFactory";
+import { TTransactionRequestRequired } from "./types";
+
+class EthManagerError extends Error {
```
DmitriyNoa: It would be good if we can implement our own hight order error e.g. WalletError or something and derive all other error from it, so we can have common signature everywhere in the app

packages/wallet/app/modules/eth/lib/EthManager.ts
```diff
@@ -0,0 +1,248 @@
...
+class EthManager {
+  private readonly adapter: EthAdapter;
+
+  private readonly logger: ILogger;
+
+  private readonly ethWalletFactory: EthWalletFactory;
+  private wallet: EthWallet | undefined;
+
+  constructor(
+    @inject(privateSymbols.rpcUrl)
```
DmitriyNoa: Nitpicking it could be on the same line

packages/wallet/app/modules/eth/lib/EthManager.ts
```diff
@@ -0,0 +1,248 @@
...
+
+    this.logger.info(`EthManager initialized. Rpc URL: ${rpcUrl}`);
+  }
+
+  /**
+   * Returns the current ETH balance of a given address.
+   *
+   * @param address - An ethereum address
+   */
+  getBalance(address: EthereumAddress): Promise<utils.BigNumber> {
```
DmitriyNoa: Just a question, this is going to be returned in wei?

packages/wallet/app/modules/eth/lib/EthManager.ts
rudolfix: we should have derivation tree for the exceptions, having base one for the whole app and then for each module. at the end you derive concrete exceptions from the module one.

or any other structure that let's you filter exceptions coming from particular place. I think I mentioned this somewhere

packages/wallet/app/modules/eth/lib/EthAdapter.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const { chainId } = await this.provider.getNetwork();
+
+    return chainId;
+  }
+
+  /**
+   * Returns the current ETH balance of a given address.
+   *
+   * @param addressOrName - An ethereum address or ENS name
```
rudolfix: `addressOrName`?

packages/wallet/app/modules/eth/lib/EthAdapter.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    return this.provider.getBalance(addressOrName);
+  }
+
+  /**
+   * Returns the current transaction count counting from the latest block tag.
+   *
+   * @param addressOrName - An ethereum address or ENS name
+   */
+  getTransactionCountFromLatest(addressOrName: EthereumAddress | EthereumName): Promise<number> {
```
rudolfix: do we need this? we do not need anything besides what we implement for light/browse/ledger wallet

packages/wallet/app/modules/eth/lib/EthAdapter.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    return this.provider.getTransactionCount(addressOrName, "latest");
+  }
+
+  /**
+   * Returns the current transaction count counting from the pending block tag.
+   *
+   * @param addressOrName - An ethereum address or ENS name
+   */
+  getTransactionCountFromPending(addressOrName: EthereumAddress | EthereumName): Promise<number> {
```
rudolfix: same

packages/wallet/app/modules/eth/lib/EthManager.ts
rudolfix: isn't `EthAdapter` too thin? we could just have provider directly here and remove `EthAdapter`/`EthAdapterFactory`. less is more

packages/wallet/app/modules/eth/lib/EthManager.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  /**
+   * Exports wallet private key to the UI
+   *
+   * @note This operation is unsafe as private key is exposed to the UI
+   *        therefore put without any encryption to the RAM memory
+   *
+   * @throws NoWalletPluggedError - When no wallet was plugged yet
+   */
+  unsafeExportWalletPrivateKey = async (): Promise<EthereumPrivateKey> => {
```
rudolfix: we should have export mnemonics or this is generic export secret?

packages/wallet/app/modules/eth/lib/EthManager.ts
DmitriyNoa: I guess with this wrapper we can easily swap a provider e.g. from ethers to web3 or something else if needed in the future

packages/wallet/tsconfig.unit.json
desfero: Fixed

packages/wallet/app/modules/eth/lib/EthAdapter.ts
desfero: I had raised the same question some time ago to always use `EthereumAddressWithChecksum` but we end up with sticky to `EthereumAddress` as we do not force checksummed address on the UI level.

packages/wallet/app/modules/eth/lib/EthAdapter.ts
desfero: Talked on slack about the reason to use a lot of loggers

packages/wallet/app/modules/eth/lib/EthAdapter.ts
desfero: Moved to enum :)

packages/wallet/app/modules/eth/lib/EthManager.ts
desfero: fixed

packages/wallet/app/modules/eth/lib/EthManager.ts
desfero: interesting that it was not caught by preettier

packages/wallet/app/modules/eth/lib/EthManager.ts
desfero: right, that's the exact reason to have adapter so we can hide whether it's ethers or web3 or anything else from top-level `EthManager`.

And yes, `getBalance` will return `wei` :)

packages/wallet/app/modules/eth/lib/EthAdapter.ts
desfero: yes, it's possible to pass either address or ens name and it will be automatically resolved to an address

packages/wallet/app/modules/eth/lib/EthAdapter.ts
desfero: we do use transaction count to populate `nonce`

packages/wallet/app/modules/eth/lib/EthAdapter.ts
DmitriyNoa: As I understand on UI we are going to use either masked address or even Some user names. So maybe we can move to EthereumAddressWithChecksum internally later.

PR #3918, opened 2020-01-16
[All devs approval required] Introduce module format

desfero:
At the moment it's not possible to share modules between different apps because they are coupled with our global typings and CI container. This PR introduces a module format that will allow modularizing our modules 🚀 and at the same time will keep a proper type of safety everywhere. 

The additional very important benefit of modularization is the lazy loading of modules on demand (for .e.g load all business logic related to bank transfer flow only when a user starts that flow).

Aa couple of files were not marked by git as moved, these files will have a comment on top mentioning that.

Other notable changes:

- **AppDispatch** was renamed to **AppGlobalDispatch** to clearly state that it's a global web app dispatch.
- **IAppState** was renamed to **TAppGlobalState** to clearly state that it's a global app state.
-`wsrun` packages was added to the root to run build commands with proper order (at first all packages not dependent on other `@neufund/*` packages and then all others) (see for details https://github.com/yarnpkg/yarn/issues/7150).
- given that marketing email unsubscription flow is now a separate lazy-loaded module I wrote some e2e tests 

In the long term, we will move all shared modules from `web/src/modules` to `modules` package.

PR #4087, opened 2020-02-21
Fix broken storybook production build

desfero:
**Technical notes**
This pr reverts some changes introduced in #3918 (mainly moving `createStore` from index to `store.ts`)
The reason is the circular dependency between store.ts and `DocumentConfidentialityAgreementModal.ts` (store.ts imports all sagas and DocumentConfidentialityAgreementModal is imported in ETO saga). We haven't caught it on the prod build of the web app because it hasn't concatenated sagas in the `store.ts` module as they were references in more than one place.

In the long term maybe we want to migrate to the structure we have in the mobile app where store.ts is a folder with `create, types, utils` files as it will avoid circular dependencies in some cases.

Also, there is something wrong with storybook webpack/babel configuration as dead code is not eliminated (maybe babel replaces es modules by commonjs so then webpack is not able to analyze unused exports)


arjunu:
@desfero thanks!

PR #4088, opened 2020-02-21
4048 Application storage

DmitriyNoa:
This PR contains interfaces and a draft implementation. The final implementation and unit test coverage will be added when interfaces are finalized.


packages/wallet/app/modules/storage/classes/StorageSchema.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Class representing a schema item
+ * @class  StorageSchema
+ */
+export class StorageSchema {
+  /**
+   * @param {string} version - version of a schema
+   * @param {string} id - schema id
+   * @param schema - YupTS object schema
+   */
+  constructor(public version: string, public id: string, public schema: YupTS.Schema<unknown>) {}
```
rudolfix: IMO it would be better if version is an integer so we have order of versions and we can migrate from version 1 to 2 to 3 to the newest one

We can live with string though, IMO `number` will be more informative

packages/wallet/app/modules/storage/types/IAppStorageProvider.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * An interface that a storage provider has to implement.
+ * Storage provider is an abstraction class that wraps a storage library e.g. AsyncStorage or SQLite,
+ * and provides common interface to interact with the storage and incapsulates specific storage implementation.
+ *
+ * @interface  IAppStorageProvider
+ */
+export interface IAppStorageProvider {
```
rudolfix: OK we can have it! I think you can already name it ISecureStorageProvider and also make it common with @desfero code, where it's being mocked up separately

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+   * @param {string} key - Key to access the value.
+   * @param {string} value - Value to save.
+   */
+  async setItem(key: string, value: IStorageItem) {
+    // get the schema by id
+    const schema = this.schemaStorage.getSchema(value.schemaId);
+
+    if (!schema) {
+      this.logger.error(`Schema is required`);
+      return;
```
rudolfix: I assume the final version will raise on errors :)

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+    try {
+      data = JSON.stringify(value.data);
+    } catch (e) {
+      this.logger.error("JSON stringify error");
+      return;
+    }
+
+    // save data to the storage provider
+    try {
+      await this.provider.setItem(`${this.name}:${key}`, data);
```
rudolfix: I assume that final version will store data + metadata... right now only serialized data is stored

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+    } catch (e) {
+      this.logger.error("Storage error", e);
+    }
+  }
+
+  /**
+   * Get item from storage.
+   * @param {string} key - Key to access the value.
+   */
+  async getItem(key: string) {
```
rudolfix: shouldn't it return `StorageItem`? or we need a separate method that would return item metadata as well

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+    try {
+      const value = await this.provider.getItem(`${this.name}:${key}`);
```
rudolfix: you should deserialize storage item here and run it through `migrate` command of the `StorageSchema`

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+      let deserialized = value ? JSON.parse(value) : {};
+      return deserialized;
+    } catch (e) {
+      this.logger.error("Json parse error", e);
+    }
+  }
+
+  async migrate() {
```
rudolfix: migrate is per schema

packages/wallet/app/modules/storage/types/IStorageItem.ts
```diff
@@ -0,0 +1,22 @@
...
+   */
+  data: unknown;
+  /**
+   * @property {number} lastUpdated - timestamp of the last update.
+   */
+  lastUpdated: number;
+  /**
+   * @property {string} appVersion - application version.
+   */
+  appVersion: string;
```
rudolfix: you need to add `schemaVersion` here otherwise you'll not know what version of the data you are reading

packages/wallet/app/modules/storage/types/IStorageItem.ts
```diff
@@ -0,0 +1,22 @@
...
+   */
+  schemaId: string;
+  /**
+   * @property {unknown} data - serializable data.
+   */
+  data: unknown;
+  /**
+   * @property {number} lastUpdated - timestamp of the last update.
+   */
+  lastUpdated: number;
```
rudolfix: would be cool to have creation date as well

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,126 @@
...
+  getStorageName = (): string => {
+    return this.name;
+  };
+
+  /**
+   * Save item to storage.
+   * @param {string} key - Key to access the value.
+   * @param {string} value - Value to save.
+   */
+  async setItem(key: string, value: IStorageItem) {
```
rudolfix: maybe a signature `setItem(key: string, schemaId: string, value: object): IStorageItem` is better?
`IStorageItem` contains only data that is set internally here so I think there's no reason why external components should pass it.

however it would be nice to return it

packages/wallet/app/modules/storage/classes/AppStorage.ts
DmitriyNoa: Sure. It's just a mock

packages/wallet/app/modules/storage/types/IStorageItem.ts
DmitriyNoa: I created a separate type StorageMetadata. it will hold all the additional data along with the schema version, it will be stored together with the main item data Please check and tell what do you think.

packages/wallet/app/modules/storage/classes/AppStorage.ts
DmitriyNoa: Added metadata to store

packages/wallet/app/modules/storage/classes/StorageSchema.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * @param {string} data - validate data with YUP
+   * returns boolean promise
+   */
+  async validate(data: unknown): Promise<boolean> {
```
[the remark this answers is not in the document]
DmitriyNoa: Added basic migrate signature

packages/wallet/app/modules/storage/types/IAppStorageProvider.ts
DmitriyNoa: This is already mocked with packages/wallet/app/modules/storage/classes/AsyncStorageProvider.ts

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,163 @@
...
+  getStorageName = (): string => {
+    return this.name;
+  };
+
+  /**
+   * Save item to storage.
+   * @param {string} key - Key to access the value.
+   * @param {string} value - Value to save.
+   */
+  async setItem(key: string, schemaId: string, value: unknown): Promise<IStorageItem> {
```
desfero: I see that we have two problems here:

1. setItem/getItem is not typesafe during compile time as both `schemaId` can be mistyped or invalid value passed. It's also hard to catch errors when value schema was changed because typescript won't show any kind of errors related to type mismatch.

2. migrations should be manually provided during `getItem` which may be quite error-prone given we can have a lot of `getItem` calls.

I would suggest that we move `AppStorage` as an abstract class and for each instance of a key (for .e.g wallet storage we create the wrapper on top of `AppStorage` with key and schema (with versioning and migrations up and down)
This will saves us from passing `schemaId` manually and will make `value` properly typed.

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,163 @@
...
+    // get the schema by id
+    const schema = this.schemaStorage.getSchema(schemaId);
+
+    if (!schema) {
+      throw new Error(`Schema not found`);
+    }
+
+    // Create metadata
+    const metadata = new StorageMetaData(Date.now(), Date.now(), schemaId, schema.version);
```
desfero: i would say that we need to read the previous wallet here and get `create` from there as now `create` and `lastUpdated` is always equal.

DmitriyNoa: Why do we need a wallet here? Storage should store any king of data, not only related to a wallet. or what do you mean?

desfero: ah, sorry the PK storage PR put me into a mood of `wallet, wallet everywhere`. You are completely right it should store any kind of data and here i just mean to read the previous value, if there is a previous value took `created` metadata from there.

packages/wallet/app/modules/storage/types/AsyncStorage.d.ts
```diff
@@ -0,0 +1 @@
+//declare module '@react-native-community/async-storage/jest/async-storage-mock';
```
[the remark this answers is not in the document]
desfero: ok, have seen you comment in PK storage pr and it's a good idea to just store typings where they really belong. Just a question then, why it's commented?

packages/wallet/package.json
```diff
@@ -72,21 +77,34 @@
...
+    "eslint-plugin-import": "^2.20.1",
     "jest": "^25.1.0",
     "metro-react-native-babel-preset": "^0.56.0",
     "prettier": "^1.19.1",
     "react-devtools": "^3.6.3",
+    "react-native-clean-project": "^3.3.0",
+    "react-native-typescript-transformer": "^1.2.13",
     "react-test-renderer": "16.9.0",
+    "ts-jest": "^25.2.1",
+    "typedoc": "^0.16.10",
```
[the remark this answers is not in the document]
DmitriyNoa: Yes, will add it

packages/wallet/app/modules/storage/classes/AppStorage.ts
DmitriyNoa: This is a good point. Updated with a dummy "PersonStorage" implementation. Please check

packages/wallet/app/modules/storage/classes/StorageMetaData.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe("StorageMetadata", () => {
+  it("should create an instance of StorageMetadata", () => {
+    const storageMetaData: StorageMetaData = new StorageMetaData(
+      Date.now(),
+      Date.now(),
+      "TestID",
+      1,
+    );
+    expect(storageMetaData instanceof StorageMetaData).toBeTruthy();
```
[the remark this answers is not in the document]
DmitriyNoa: Basically it was a stub that I put for a testing time.

packages/wallet/app/di/setupBindings.ts
```diff
@@ -1,14 +1,27 @@
...
 
 /**
  * We use plain object for injecting deps into sagas
  */
+export function setupBindings(): ContainerModule {
+  return new ContainerModule(bind => {
+    bind<TLibSymbolType<typeof symbols.config>>(symbols.config).toConstantValue({
+      applicationStorageNamespace: "mainStorage",
+    });
+    storageBindings(bind);
```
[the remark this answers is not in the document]
DmitriyNoa: Moved

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,163 @@
...
+   * Save item to storage.
+   * @param {string} key - Key to access the value.
+   * @param {string} value - Value to save.
+   */
+  async setItem(key: string, schemaId: string, value: unknown): Promise<IStorageItem> {
+    // get the schema by id
+    const schema = this.schemaStorage.getSchema(schemaId);
+
+    if (!schema) {
+      throw new Error(`Schema not found`);
```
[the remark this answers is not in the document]
DmitriyNoa: Added

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,163 @@
...
+    const metadata = new StorageMetaData(Date.now(), Date.now(), schemaId, schema.version);
+
+    // create storage item
+    const storageItem = new StorageItem(value, metadata);
+
+    // validate data schema and throw and error if it's invalid.
+    try {
+      await schema.validate(storageItem.data);
+    } catch (error) {
+      throw new SchemaMismatchError(error.errors);
```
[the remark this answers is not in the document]
DmitriyNoa: Good idea. Added

packages/wallet/app/modules/storage/classes/AppStorage.ts
```diff
@@ -0,0 +1,163 @@
...
+import { IStorageItem } from "../types/IStorageItem";
+import { TConfig } from "../../../di/types";
+import { StorageMetaData } from "./StorageMetaData";
+
+/**
+ * A class representing an application storage
+ * @class  AppStorage
+ */
+@injectable()
+class AppStorage {
```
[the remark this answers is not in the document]
DmitriyNoa: Added

packages/wallet/package.json
```diff
@@ -96,6 +114,7 @@
       "**/react-native-*",
       "**/react-native-*/**",
       "**/detox",
+      "**/yup",
```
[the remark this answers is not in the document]
DmitriyNoa: We use an old version of Yup on the web which doesn't have some new methods and types. So we need to keep it separate from the WEB until we update second.

packages/wallet/app/modules/storage/types/types.d.ts
```diff
@@ -0,0 +1 @@
+declare module "react-intl-phraseapp";
```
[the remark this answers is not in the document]
DmitriyNoa: Removed

packages/wallet/app/modules/storage/types/KeyValue.ts
```diff
@@ -0,0 +1 @@
+export type KeyValue = [string, string | null];
```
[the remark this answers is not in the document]
DmitriyNoa: Removed

packages/wallet/app/modules/storage/types/IStorageItem.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * An interface that represents a storage data item
```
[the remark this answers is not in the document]
DmitriyNoa: "I'm just curious whether we can get rid of duplicating names and hope that tsdoc parser can read the name from a real code" - yes, we can need to investigate it. I also feel strange adding names in ts docs along with actual methods.

packages/wallet/app/modules/storage/types/AsyncStorage.d.ts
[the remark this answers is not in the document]
DmitriyNoa: Leftover from testing. Uncommented

packages/wallet/jest.config.js
```diff
@@ -1,4 +1,18 @@
...
+    ...tsjPreset.transform,
+    "\\.js$": "<rootDir>/node_modules/react-native/jest/preprocessor.js",
+  },
+  globals: {
+    "ts-jest": {
+      babelConfig: true,
+      tsConfig: 'tsconfig.unit.json'
+    },
+  },
+  setupFiles: ["./jest-setup-file.ts"],
```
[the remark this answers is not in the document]
DmitriyNoa: Cleaned to a minimal working one

packages/wallet/app/modules/storage/classes/StorageMetaData.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export class StorageMetaData implements IStorageMetaData {
+  public appVersion: number;
+  constructor(
+    public created: number,
+    public lastUpdated: number,
+    public schemaId: string,
+    public schemaVersion: number,
+  ) {
+    this.appVersion = 1; // replace with real app version getter
```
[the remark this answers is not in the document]
DmitriyNoa: This is something to talk about. Do we match the version in package.json with the actual app version in AppStore? E.g. we might have different versions later on App and Android stores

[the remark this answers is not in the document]
desfero: ok, agree that's a topic we need to research as we have package.json/ios/android versions :D

Issue #3981, opened 2020-02-03
Lightwallet Login


Moejoe90:
**Background**
This is a design change to the already available lightwallet login. In theory there should be no logic change to the login process.

**Designs**

https://app.zeplin.io/project/5e20533697f09f7ff8d0726d/screen/5e205e0a34ae967ece08c961

**Acceptance criteria**

- [ ] Given I'm a user trying to login. I will only reach the login page via an access link

- [ ] The form should also work in storybook as a story. 

- [ ] Add Login without link to storybook
https://app.zeplin.io/project/5e20533697f09f7ff8d0726d/screen/5e205e0befd9b2aac03e1831

- [ ] Given I'm a user who doesn't have access link I should see 
https://app.zeplin.io/project/5e20533697f09f7ff8d0726d/screen/5e205e0befd9b2aac03e1831

**Technical notes**
+ The wallet selector should be shared between `register` and `login`

+ In theory there shouldn't be any difference in logic between the old login and this. Ther will be a new link in the future but that can be done as a seperate ticket

**Testing notes**

PR #4094, opened 2020-02-24
#3981 LightWallet login redesign

drakkein:
This PR redesigns Light wallet login page

image

image

image

image

image


packages/web/app/components/appRoutes.ts
```diff
@@ -11,6 +11,7 @@ export const appRoutes = {
 
   registerNominee: "/nominee/register",
   login: "/login",
+  loginBrowser: "login/browser",
```
AlexeyTuboltsev: 👍 
I too think this internal login/signup router should be put at the top level completely so that sagas could start off of it

packages/web/app/components/shared/inputs/TextField.tsx
```diff
@@ -51,20 +52,27 @@ export const TextField: React.FunctionComponent<TFieldProps &
...
   autoFocus,
   className,
+  manualError,
   ...props
 }) => {
   const [unmaskPassword, setUnmaskPassword] = React.useState(false);
+  const [showManualError, setShowManualError] = React.useState(!!manualError);
   // right now there's only one type of adornment, password mask icon
   const hasAdornment = type === "password";
   const overridingType = hasAdornment && unmaskPassword ? "text" : type;
```
AlexeyTuboltsev: I would do this as a separate component. 
we have many text input types:
- a simple one, 
- a multiline field (aka textArea)
- an RTE one, 
- an eth input, 
- a numeric input (that checks if the string can be parsed to number)
- a password input

maybe smth else
this all looks like different components with different internal logic to me. Would be cool to make them all implement one interface. 
Anyway there's no need to put more props and internal logic into one, they are separate.

packages/web/app/components/wallet-selector/light/Login/LoginLightWallet.tsx
```diff
@@ -9,24 +9,31 @@ import { appConnect } from "../../../../store";
...
 
-    {email ? <LoginWithEmailLightWallet email={email} /> : <MissingEmailLightWallet />}
-  </>
-);
+export const LoginLightWalletComponent: React.FunctionComponent<TComponentProps> = ({
+  email,
+  goToPasswordRecovery,
+}) =>
+  email ? (
+    <LoginWithEmailLightWallet email={email} goToPasswordRecovery={goToPasswordRecovery} />
```
AlexeyTuboltsev: I'd move this to a `branch`

packages/web/app/components/wallet-selector/light/Login/LoginWithEmailLightWallet.unsafe.tsx
```diff
@@ -40,30 +40,46 @@ interface IStateProps {
...
 const LoginLightWalletForm: React.FunctionComponent<TProps & FormikProps<IFormValues>> = props => (
   <FormDeprecated>
-    <FormField
+    <TextField
       type="password"
-      placeholder="Password"
+      placeholder={props.intl.formatIntlMessage("wallet-selector.neuwallet.login.placeholder")}
       name={PASSWORD}
       data-test-id="light-wallet-login-with-email-password-field"
+      manualError={props.errorMsg && getMessageTranslation(props.errorMsg)}
```
Moejoe90: we really need to think about a general solution for having these async validations. 

We can't just add a manual validation to the field there should be a way for these errors to show. 

In all cases I think a general solution is outside the scope of this PR but I would ask you to create a ticket that touches this topic. 

Add general Async Validation Mechanisim should be the title or something like that

packages/web/app/components/wallet-selector/light/Login/LoginLightWallet.tsx
```diff
@@ -9,24 +9,31 @@ import { appConnect } from "../../../../store";
...
 
-export const LoginLightWallet = appConnect<IStateProps>({
+export const LoginLightWallet = appConnect<IStateProps, TDispatchProps>({
   stateToProps: s => ({
     email: selectLightWalletEmailFromQueryString(s) || selectPreviousLightWalletEmail(s.web3),
   }),
+  dispatchToProps: dispatch => ({
+    goToPasswordRecovery: () => dispatch(actions.routing.goToPasswordRecovery()),
+  }),
 })(LoginLightWalletComponent);
```
AlexeyTuboltsev: MissingEmailLightWallet is very simple. I'd move it to a `branch`
and in the main `appconnect` I would query all the stuff that `LoginWithEmailLightWallet` needs and then put `LoginLightWalletForm` into the main branch. In other words, make one component with one appconnect instead of many nested ones.

packages/web/app/components/wallet-selector/light/Login/LoginWithEmailLightWallet.unsafe.tsx
```diff
@@ -40,30 +40,46 @@ interface IStateProps {
...
+  goToPasswordRecovery: () => void;
 }
 
-type TProps = IOwnProps & IStateProps & IDispatchProps;
+type TProps = IExternalProps & IStateProps & IDispatchProps & IIntlProps;
 
 const LoginLightWalletForm: React.FunctionComponent<TProps & FormikProps<IFormValues>> = props => (
   <FormDeprecated>
-    <FormField
+    <TextField
```
AlexeyTuboltsev: is this textField always active? Also during `props.isLoading` ?

packages/web/app/components/wallet-selector/light/Login/LoginWithEmailLightWallet.unsafe.tsx
```diff
@@ -72,40 +88,23 @@ const LoginEnhancedLightWalletForm = withFormik<TProps, IFormValues>({
...
-    </Row>
-  </>
+export const LoginWithEmailLightWalletComponent: React.FunctionComponent<TProps> = props => (
+  <section className={styles.section}>
+    <p data-test-id={"modals.wallet-selector.login-light-wallet"}>
+      <FormattedMessage
+        id="wallet-selector.neuwallet.login.prompt"
+        values={{
+          email: <u data-test-id="light-wallet-login-with-email-email-field">{props.email}</u>,
+          lineBreak: <br />,
```
AlexeyTuboltsev: why not simply a `FormattedHTMLMessage`?

packages/web/app/modules/routing/actions.ts
```diff
@@ -40,6 +40,8 @@ export const routingActions = {
   // login
   goToLogin: (state: TLoginRouterState) => push(appRoutes.login, state),
 
+  goToPasswordRecovery: () => push(appRoutes.restore),
```
AlexeyTuboltsev: 👍

packages/web/app/components/appRoutes.ts
drakkein: In this particular situation, this is not being used anywhere. But indeed I do share the same opinion. @Moejoe90 let's make a ticket to do it

packages/web/app/components/shared/inputs/TextField.tsx
drakkein: @AlexeyTuboltsev @Moejoe90 maybe we should leave it as it is for now, since it's working. I've got general concept how we can abstract it described in this issue https://github.com/Neufund/platform-frontend/issues/4099. It should be possible to be done in the way from Tech notes, but we need to spend some time testing it out.

packages/web/app/components/wallet-selector/light/Login/LoginLightWallet.tsx
drakkein: Yeah, indeed this is good idea

packages/web/app/components/wallet-selector/light/Login/LoginWithEmailLightWallet.unsafe.tsx
drakkein: Let's disable it! When `isLoading` is true.

packages/web/app/components/wallet-selector/light/Login/LoginWithEmailLightWallet.unsafe.tsx
drakkein: i need to add `data-test-id="light-wallet-login-with-email-email-field"` to only email part, this is much cleaner than setting it inside translations file and render via `FormattedHTMLMessage`

PR #4098, opened 2020-02-25
Move Text Field & Checkbox to design system package

arjunu:
Main changes:
- Included new Checkbox component
- moved Input components to DS pkg
- moved utils used by Input to DS pkg
- moved yupUtils to DS pkg

Note:
- yupUtils.spec still lives inside Web package cause Design system doesn't have CI step for unit tests yet
- utils.spec.ts still tests the functions moved out of it to Design system for the same reason
- added TODOs for both the above points


arjunu:
Reopened https://github.com/Neufund/platform-frontend/pull/4104 with a new base branch

Issue #4018, opened 2020-02-11
Metamask/web3 wallet auto detection


Moejoe90:
**Background**
This should add some mechanism that allows us to detect if matamask is available and store that information in store. 

Using an acitvation Action the auto detection saga should engage and check for the availability of metamask. and change the store accordingly. 

All other components can use this via selectors.

**Acceptance criteria**
* [ ] autodetect should happen when entering default registration or login route
* [x] if web3 wallet is detected as above, the route should be changed to web browser route
* [x] in case of registration - that's it. we let the user to fill the e-mail and then begin the connect process
* [ ] in case of login - we launch the connect process right away
* [ ] EIP-1102 (see below) changes deprecated `ethereum.enable` please add a ticket to watch this release  of Metamask: https://github.com/MetaMask/metamask-extension/issues/7970

**Technical notes**
Web3 compatible wallets inject an web3 instance into `window` object as per https://eips.ethereum.org/EIPS/eip-1102.
We already have the logic implemented in the first lines of `BrowserWalletConnector::connect` method. You need to extract injection detecting code.

**Testing notes**
Autodetect must run on the following wallets.
desktop
- metamask (you should also be able to login on dev, stage and prod networks)
- gnosis wallet (login should work on production network)
- opera beta with wallet (same as above)

mobile wallets (those have embedded DAPP browsers)
- mobile metamask (you should also be able to login on dev, stage and prod networks)
- opera mobile beta (login only on production)
- imToken (login only on production)

Please note that most of the wallets do not support test networks but still autodetect must work! The login process will be interrupted on dev and stage by invalid chain id.
That does not apply to mobile metamask where you can select the network.

PR #4101, opened 2020-02-25
User SignUp Epic

arjunu:
#p3

Redirect on Browser wallet discovery partially works full implementaion will be done in the future once the relevant blockers are resolved

One link activation is not done on this PR


packages/web/app/components/wallet-selector/WalletSelectorLogin/WalletLedger/WalletLedgerChooser/WalletLedgerChooserBase.module.scss
```diff
@@ -0,0 +1,41 @@
...
+  width: 100%;
+  margin: px-to-rem(20px) 0 px-to-rem(24px);
+  font-family: $font-family-roboto;
+  font-size: px-to-rem(16px);
+}
+
+.title {
+  font-family: $font-family-montserrat;
+  font-size: px-to-rem(32px);
+  font-weight: 600;
```
Moejoe90: Don't we have this in the typography

packages/web/app/test-e2e/routing/RedirectUserWithBrowserWallet.spec.e2e.io.ts
```diff
@@ -0,0 +1,30 @@
+import { assertUserInBrowserWalletLoginPage } from "../utils/assertions";
+import {
+  goToLoginLightWalletWithMockedWeb3,
+  goToLoginWithMockedWeb3,
+  goToRegisterLightWalletWithMockedWeb3,
+  goToRegisterWithMockedWeb3,
+} from "../utils/navigation";
+
+describe("Redirect user with browser wallet", () => {
```
Moejoe90: needs to be defined with a tag

packages/web/app/test-e2e/routing/RegisterRedirectIfNoVerifiedEmail.spec.e2e.io.ts
```diff
@@ -0,0 +1,19 @@
+import { appRoutes } from "../../components/appRoutes";
+import { kycRoutes } from "../../components/kyc/routes";
+import {
+  DEFAULT_PASSWORD,
+  generateRandomEmailAddress,
+  registerWithLightWalletIssuer,
+} from "../utils/index";
+
+describe("Lock some routes if registration is not complete", () => {
```
Moejoe90: needs to be defined with a tag

packages/web/app/test-e2e/obsolete/generateSeedAndAddress.spec.node.ts
```diff
@@ -1,10 +1,10 @@
 import { expect } from "chai";
 
 import { generateSeedAndAddress } from "./generateSeedAndAddress";
-import { DEFAULT_HD_PATH } from "../utils/userHelpers";
+import { DEFAULT_HD_PATH } from "../utils";
 
 describe("generateSeedAndAddress", () => {
-  it("generates seed and password", async () => {
+  it.skip("generates seed and password", async () => {
```
Moejoe90: check this

packages/web/app/test-e2e/investor/bank/redeem/RedeemNEUR.spec.e2e.io.ts
```diff
@@ -34,6 +34,8 @@ describe("Redeem", function(): void {
 
       // start redeem flow
       cy.get(tid("wallet-balance.neur.redeem-button")).click();
+      // The test is very flaky on fast machines
+      cy.wait(500);
```
pavelneu: @Moejoe90 
What if we add here
`cy.get(tid("bank-transfer.reedem-init.continue")).should("be.visible");`
instead of
`cy.wait(500);`
?

packages/web/app/test-e2e/issuer/eto/nominee/LinkNomineeStep.spec.e2e.io.ts
```diff
@@ -121,7 +121,7 @@ describe("Eto Forms link nominee", () => {
     },
   );
 
-  it(
+  it.only(
```
pavelneu: @Moejoe90 
why? won't it skip other tests here?

packages/web/app/test-e2e/other/login/lightwallet/InvestorLogin.spec.e2e.io.ts
```diff
@@ -20,23 +18,48 @@ import {
...
     cy.clearLocalStorage();
+
+    email = generateRandomEmailAddress();
+    password = "strongpassword";
   });
 
-  it("should register user with light-wallet and send email #login #p1", () => {
-    const email = generateRandomEmailAddress();
-    const password = "strongpassword";
+  it("can't register without accepting Terms of Use", () => {
```
pavelneu: @arjunu 
```suggestion
  it("should not register without accepting Terms of Use #login #p3", () => {
```

packages/web/app/test-e2e/other/logout/LogoutActions.spec.e2e.io.ts
```diff
@@ -43,7 +43,7 @@ const setKeyFromStorageWithEvents = (Window: Window, key: string, newValue: stri
   );
 };
 
-describe("Login/Logout", () => {
+describe("login/Logout", () => {
```
pavelneu: @arjunu 
auto-replacing?

packages/web/app/test-e2e/other/recovery/WalletRecovery.spec.e2e.io.ts
```diff
@@ -1,34 +1,32 @@
...
   tid,
   typeLightwalletRecoveryPhrase,
 } from "../../utils/index";
 
 describe("Wallet recovery", function(): void {
-  this.retries(2);
-
-  it("should show error modal for invalid recovery phrases #backup #p1", () => {
-    cy.visit(`${recoverRoutes.seed}`);
+  it("should show error modal for invalid recovery phrases @backup @p1", () => {
```
pavelneu: @Moejoe90 
looks like an issue after merge
the line above with `#` should be here instead of `@`

packages/web/app/test-e2e/other/recovery/WalletRecovery.spec.e2e.io.ts
```diff
@@ -90,18 +84,45 @@ describe("Wallet recovery", function(): void {
...
+        logoutViaAccountMenu();
+
+        cy.visit(appRoutes.login);
         lightWalletTypeLoginInfo(email, password);
         assertDashboard();
       });
     });
   });
+
+  it("should recover existing user with verified email from saved phrases and change email #recover #p3 ", function(): void {
```
pavelneu: @Moejoe90 
checked the whole repository
`#recover` is only one test here
would you want to leave `#recover` instead of `#backup`?

packages/web/app/test-e2e/routing/RedirectUserWithBrowserWallet.spec.e2e.io.ts
```diff
@@ -0,0 +1,31 @@
+import { assertUserInBrowserWalletLoginPage } from "../utils/assertions";
+import {
+  goToLoginLightWalletWithMockedWeb3,
+  goToLoginWithMockedWeb3,
+  goToRegisterLightWalletWithMockedWeb3,
+  goToRegisterWithMockedWeb3,
+} from "../utils/navigation";
+
+describe.skip("Redirect user with browser wallet #redirect #p3", () => {
```
pavelneu: @Moejoe90 
you missed a little with test suites
it should be in `it`

packages/web/app/test-e2e/investor/bank/redeem/RedeemNEUR.spec.e2e.io.ts
Moejoe90: its better if we wait the whole issue is cypress on those CI machines see the element so quick that I think it breakes the flow. Lets keep it like this and see if that's the reason this test is flaky sometimes

packages/web/app/test-e2e/issuer/eto/nominee/LinkNomineeStep.spec.e2e.io.ts
Moejoe90: this is a leftover I'll remove it

packages/web/app/test-e2e/other/login/lightwallet/InvestorLogin.spec.e2e.io.ts
Moejoe90: Added as part of a review fix commit

packages/web/app/test-e2e/other/logout/LogoutActions.spec.e2e.io.ts
Moejoe90: reverting this

packages/web/app/test-e2e/other/recovery/WalletRecovery.spec.e2e.io.ts
Moejoe90: good catch done

packages/web/app/test-e2e/other/recovery/WalletRecovery.spec.e2e.io.ts
Moejoe90: you're the tests guy, :D 
Changing to backup

packages/web/app/test-e2e/other/logout/LogoutActions.spec.e2e.io.ts
arjunu: No idea how this happened

packages/web/app/test-e2e/routing/RedirectUserWithBrowserWallet.spec.e2e.io.ts
Moejoe90: The whole suite is disabled here. It should get enabled back once
https://github.com/Neufund/platform-frontend/issues/4018
or what do you mean here exactly?

andywalton:
Hi, so I have a few small things I think we should fix/improve for this PR - I'll post a separate comment per issue:

The first is on our Log in view for a returning "light wallet" user. The text is weirdly formatted and includes an extra unnecessary line break. Please remove this line break and let the text run correctly.

There is also random `wallet-selector` text in the paragraph.

Please also double check the font styles, ie.

```$body = font size 16px, line height 24px, font Roboto Regular```
To log in to your Neufund account, please click the unique login link we emailed you when you signed up.

```$body-small = font size 14px, line height 24px, font Roboto Regular```
Forgotten your password? Recover your account 

001-LW-login-returning-user-text-layout-issue

**Comment by andywalton** [comment]
## Login MetaMask/Ledger Nano

1. Please improve the alignment of the error messages when something goes wrong during MetaMask and/or Ledger Nano connection attempts.

For example, it currently displays:

MetaMask example:
002-Log in with MetaMask-error-alignment

Ledger Nano example:
002b-Log in with Ledger-error-alignment

An example of how it should look:
Screenshot 2020-03-13 15 47 34

Please update the error icon to align next to the text. Check that error text styling is:
```$body-small = font size 14px, line height 24px, font Roboto Regular```

A nice to have, if you can do this now, if not let me know would be to change our error red colour, $base-red to `#FF5B5B;`. If you don't do this then please let me know.

Please use the screen design in Zeplin as your guide: https://zpl.io/aMNAQ4d

**Comment by andywalton** [comment]
Change the welcome modal copy for MetaMask/Ledger Nano signed up users:

004-Change this copy

Ie.

> Please check your inbox
> 
> We've sent a welcome email containing an email verification link to example@example.com. To continue with the account setup, please confirm your email address.

**Comment by andywalton** [comment]
Sign up with email (Light Wallet)

The below image highlights the issues and changes.

Screenshot 2020-03-13 10 07 07

NB, default input borders should be 1px, they should be #CFD7DD and should only change to 2px borders when they're in a focussed/active or error state.

**Comment by andywalton** [comment]
Call to actions for Sign up and Log in.

Please make the CT to actions on the Sign up and Log in views use our `primary-btn` styling - ie. our brand yellow button. The below screenshot shows how we're incorrectly using a secondary button for the "Get started".

Screenshot 2020-03-13 10 15 07

**Comment by glupiawymowka** [comment]
### **Log in fixes**
- [ ] Please remove "wallet-selector" from the copy in the lightwallet-login without link flow.
image

- [ ] Please fix the typo in the MetaMask-login (should be `access to your wallet`) and add a dot at the end of the sentence.
image

- [ ] Ledger-login: Given the Ledger's session ended, and the user tries to log in again without reactivating it, the wallet addresses table loads for a while
ledger reload

- [ ] Missing `Learn more` link for inactive/not found Ledger state
image


### **Sign up**
- [ ] `Please enter a valid email` should be shown also in this case. Currently, we don't disable the `get started` button, but the user can't continue getting a popup from Chrome
image
image

File: packages/web/app/test-e2e/other/login/lightwallet/InvestorLogin.spec.e2e.io.ts (hunk 7 above)
Comment (pavelneu) [reply packages/web/app/test-e2e/other/login/lightwallet/InvestorLogin.spec.e2e.io.ts]: @Moejoe90 
there is still missed hashes (test suites)

it("should not register without accepting Terms of Use `#login #p3`", () => {

File: packages/web/app/test-e2e/routing/RedirectUserWithBrowserWallet.spec.e2e.io.ts (hunk 11 above)
Comment (pavelneu) [reply packages/web/app/test-e2e/routing/RedirectUserWithBrowserWallet.spec.e2e.io.ts]: please move the hashes (test suites) from `describe` to `it` (to one line below)

**Comment by andywalton** [comment]
**Account recovery**

The input fields for the account recovery words, please fix the top and bottom right border radius for these fields. Something is breaking the rounded border radius by adding in some square white box.

Screenshot 2020-03-13 19 16 12

Screenshot 2020-03-13 19 19 55

**Comment by rudolfix** [comment]
Signup/Log-In review:

1. I can register with existing account. and that is fine! because with user owned wallets we just pretend there's any difference between sign up and and login ;> but if we do it then let's do it well
• if you register with existing account and this does not have a verified e-mail, we start verification of e-mail entered on registration form (CURRENT BEHAVIOR)
• if you register existing account with verified e-mail - nothing happens! we perform a normal login and do not initialize verification

@andywalton pls check if this behavior is OK

2. in registration - can't we run validator that e-mail is already used when user types or when input loses focus? we have behavior from the 90ties now
image

3. for nano ledger - please bring back some minimal manual when we cannot detect ledger! error is not enough.
please connect your ledger and run Ethereum app - would be the minimum
I think we are overestimated what our users are able to do

4. after pressing select on the account address in Ledger login - we do not display anything and user needs to guess he needs to confirm on the ledger. why can't we show signer? that would correspond to what metamask shows by itself

5. when registering via Metamask and my network Main Ethereum (no a dev network) - instead of showing error message whole process hangs - as per screenshot. @glupiawymowka  reported it as well

6. metamask mobile - I cannot login/register. I create a custom network with right network id but app still shows that chain id does not match. we need to debug this. it may be a problem with Metamask

7. @andywalton are you sure showing a piece of KYC data is a good idea? if we need a nickname let's ask for display name in settings
image

**Comment by rudolfix** [comment]
recovery review

**summary**
1. you lack basic e2e tests and you do not verify if recovery e-mail is sent with proper data
2. we need to take a step back and do a detailed code review

0. a wish
can we make space or any other whitespace character to work like TAB and move focus to next word? then I can paste the whole list of words "orange iron recycle unusual cannon theory myth echo dizzy prefer arrange ugly fatigue sell rain burden meadow tiny tone spy glance agent catalog clock" PLEASE!

1. error when sending
```
https://proxy.neufund.io:2053/https://platform.neufund.io/api/user/transaction/tx_ethereum_json_rpc
{id: 1, jsonrpc: "2.0", params: [], method: "eth_blockNumber", skipCache: true}
id: 1
jsonrpc: "2.0"
params: []
method: "eth_blockNumber"
skipCache: true}
```
what is this `skipCache`? not a part of standard. we'll return error 400 from the backend.

it may be a web3 problem that requires backend fix

2. case: recovery was done over existing account without verified e-mail
```
GET /user/me
{
  "backup_codes_verified": false,
  "language": "en",
  "latest_accepted_tos_ipfs": "QmRU84rH63u4BPRpreSMDyq4tbXh4yZ5XsnkcGB7rcmyKP",
  "type": "investor",
  "unverified_email": "rudolfix+lw@possij.pl",
  "user_id": "0x0eA457c653bE10E4bC31b46F0dfD96e418Ee3ecd",
  "wallet_subtype": "unknown",
  "wallet_type": "light"
}
```
and then PUT failed because for some reason SALT was not provided:
```
PUT /user/me

backup_codes_verified: false
language: "en"
latest_accepted_tos_ipfs: "QmRU84rH63u4BPRpreSMDyq4tbXh4yZ5XsnkcGB7rcmyKP"
type: "investor"
unverified_email: "rudolfix+lw@possij.pl"
user_id: "0x0eA457c653bE10E4bC31b46F0dfD96e418Ee3ecd"
wallet_subtype: "unknown"
wallet_type: "light"
new_email: "rudolfix+lw@possij.pl"
```
backend returns:
`New e-mail provided but not salt for lightwallet for user id 0x0eA457c653bE10E4bC31b46F0dfD96e418Ee3ecd`

you should ALWAYS send salt when doing PUT user/me after recovery!
also
if unverified_email (  "unverified_email": "rudolfix+lw@possij.pl",) is same as e-mail passed from the form, you should skip `new_email` in PUT request (still SALT needs to be sent!)

**test it with e2e (you can restore same account twice)**

3. after the error above if I press CANCEL recovery I go to light wallet login which does not load. IMO this is serious - some state is not cleared. stays like that until page reload.
BECAUSE JWT IS ALREADY IN LOCAL STORAGE and on refresh I became logged in!
please do code review. this is really dangerous
fix it before (2) so error is easy to reporduce

4. case: when restoring over account that already has VERIFIED EMAIL salt is not sent as well. also when VERIFIED EMAIL is same as the one on the form please do not send it as `new_email`. just use salt
**also why wallet type is `unknown`?**

fixture:
obvious social rare jewel liquid fix giant sea exchange dilemma brush harsh pistol grief obvious promote visit web pluck imitate wire deliver neutral embrace

```
GET /user/me
{
  "backup_codes_verified": true,
  "language": "en",
  "latest_accepted_tos_ipfs": "",
  "type": "investor",
  "user_id": "0x2824a394B5a3386b3140Bc426E40e24217f0a42f",
  "verified_email": "0x2824a@neufund.org",
  "wallet_subtype": "unknown",
  "wallet_type": "unknown"
}

PUT /user/me
backup_codes_verified: true
language: "en"
latest_accepted_tos_ipfs: ""
type: "investor"
user_id: "0x2824a394B5a3386b3140Bc426E40e24217f0a42f"
verified_email: "0x2824a@neufund.org"
wallet_subtype: "unknown"
wallet_type: "unknown"
new_email: "0x2824a@neufund.org"
```

**please add e2e test case for this**

5. when user has VERIFIED EMAIL and I already have e-mail being verified but I enter e-mail in the form, we have all the problems above
```
GET
{
  "backup_codes_verified": true,
  "language": "en",
  "latest_accepted_tos_ipfs": "QmRU84rH63u4BPRpreSMDyq4tbXh4yZ5XsnkcGB7rcmyKP",
  "type": "investor",
  "unverified_email": "rudolfix+rrrfff@possij.pl",
  "user_id": "0x2824a394B5a3386b3140Bc426E40e24217f0a42f",
  "verified_email": "0x2824a@neufund.org",
  "wallet_subtype": "unknown",
  "wallet_type": "unknown"
}

PUT
backup_codes_verified: true
language: "en"
latest_accepted_tos_ipfs: "QmRU84rH63u4BPRpreSMDyq4tbXh4yZ5XsnkcGB7rcmyKP"
type: "investor"
unverified_email: "rudolfix+rrrfff@possij.pl"
user_id: "0x2824a394B5a3386b3140Bc426E40e24217f0a42f"
verified_email: "0x2824a@neufund.org"
wallet_subtype: "unknown"
wallet_type: "unknown"
new_email: "bbbebab@neufund.org"
```

**Comment by glupiawymowka** [comment]
- [ ] there is a strange spread of the other wallet buttons for login|+no access link with light wallet, while it's correct for other screens
image

## **Sign up with MetaMask**
- [ ] Both for MetaMask Sign up & Log in, there is a missing error state in case the plugin is off/missing. We keep the loading indicator forever while per design it should show:
image
Not having the `Try again` button in this case prevents the user from getting a new MM request which means that even if they turn the plugin on, the loading indicator stays and the manual page refresh is needed to proceed. 
metamask off

**Comment by Moejoe90** [comment]
One issue fixed not in code checks yes it the `optional` fields in RegisterLightWallet

**Comment by Moejoe90** [comment]
@glupiawymowka Regarding point number 2 we can't implement this because when you enable metamask after the page is loaded. metamask doesn't get injected into the page and hence that is why the application doesn't discover this. 

Unfortunatly you will need to refresh the page first so that metamask could get injected to the window object

**Comment by Moejoe90** [comment]
Final points from Marcin

- [ ] When user account exists we display the screen below: both in case of verified/unverified e-mail 
This is where the user should see different messages depending on wether its an already existing user or a new one 


- [ ] 4. I’ve found activation link infinite loop as below
link puts you in infinite metamask login loop (probably also ledger/lightwallet)
1. change e-mail. receive activation link below
https://localhost:9090/email-verify?code=4ff59df2-1978-4570-b36d3b563e2cad4a&email=rudolfix%40possij.pl&user_type=investor&wallet_type=browser&wallet_subtype
=metamask
2. activate e-mail
3. change e-mail again
4. you can log out or keep logged in
5. you click on the expired activation link
6. you land into endless login loop
7. the problem does not exist if there's no unverified e-mail. then expired link does not have any effect

_Ask marcin for details if lost_

**Comment by rudolfix** [comment]
to the above comment from @Moejoe90 
1. backup codes > true for light wallet was fixed, thanks!
3. issue below still needs to be fixed
```
There are several cases when `new_email` is not set correctly. I’ve checked them all and it seems that the best and only correct solution is to always set new_email to what you’ve got from the recovery form. I mean really always. Backend will figure out when to send which e-mail template correctly. 
```

PR #4102, opened 2020-02-25
Fixes Issuer Form Issues 

Moejoe90:


packages/web/app/components/eto/registration/pages/EtoMedia.tsx
```diff
@@ -140,10 +140,10 @@ const addTitleIfUrlNotEmpty = (titleValue: string = "") => (data: {
   url?: string;
   title?: string;
 }) => {
-  if (data.url !== undefined) {
+  if (data.url !== (undefined && "")) {
```
arjunu: did you mean
`data.url !== undefined && data.url!==""` ?

because `""!== (undefined && "")` is `true`

arjunu:
```suggestion
  if (!data.url) {
```

packages/web/app/modules/eto-flow/sagas.ts
```diff
@@ -126,24 +127,17 @@ export function* saveEto(
   action: TActionFromCreator<typeof etoFlowActions.saveEtoStart>,
 ): Generator<any, any, any> {
   try {
-    const currentEto: TEtoSpecsData = yield select(selectIssuerEto);
+    const currentEto = yield* call(apiEtoService.getMyEto);
 
     // Eto is only allowed to be modified during PREVIEW state
-    if (currentEto.state !== EEtoState.PREVIEW) {
+    if (currentEto.state && currentEto.state !== EEtoState.PREVIEW) {
```
arjunu: checking `if (currentEto.state)` is defined is redundant

Moejoe90: its not actually this was an error coming from tslint because the returned type up is `DeepPartial<type>` so everything is optional

arjunu: oh ok

glupiawymowka:
what I was able to check for now : 
 - [x] deleted value doesn't appear once you re-enter the form 
- [x] deleted value doesn't appear in the `PUT` request
- [x] the form doesn't throw errors once a voluntary value is set back to `undefined`
- [x] it's possible to : 
  - change Pitch deck link
  - change or remove ytb link
  - change or remove slideshare link 

what still needs to be tested once the ETO preview is back: 
 - [ ] how all the mentioned above fixes reflect on the ETO page
 - [ ] if the slideshare or video is removed together with the headers 
 - [ ] if the changes are reflected also in the investor's view

PR #4107, opened 2020-02-26
Tagged Issuer flow tests

pavelneu:
Tagged all tests for Issuer flow, removed duplicates and moved all to issuer folder

Before:
86 Tests

After:
P1+P2 = 14
P3 = 60

Instead of running 86 Tests after each PR we will run only 14


packages/web/app/test-e2e/issuer/eto/forms/SubmitEtoE2E.spec.e2e.io.ts
```diff
@@ -84,4 +92,20 @@ describe("Eto Forms", () => {
       submitProposal();
     });
   });
+
+  it("should fill all fields in eto @eto @p3", () => {
```
Moejoe90: I really suggest reverting the test. It was written on purpose that way because we were getting out of memory problems. 

The test is too long and cypress was getting out of memory it doesn't make sense to bring it back as one test

PR #4116, opened 2020-02-27
#4018 Add redirect on MetaMask being detected on login routes

drakkein:


packages/web/app/modules/wallet-selector/selectors.ts
```diff
@@ -14,6 +15,12 @@ export const selectUrlUserType = (router: RouterState): EUserType => {
   }
 };
 
+export const selectRouterState = (state: TAppGlobalState) => state.router;
```
AlexeyTuboltsev: IMO the only one piece of code that should be concerned with routes should be the router (saga router or the old one). This is a kinda dangerous approach to circumvent the router. 
No action needed though, I'll change it in the next pr as I'm working on register/login routes anyway

packages/web/app/modules/web3/actions.ts
```diff
@@ -23,4 +23,8 @@ export const web3Actions = {
     createAction("WEB3_LOAD_SEED", { seed, privateKey }),
   clearWalletPrivateDataFromState: () => createSimpleAction("WEB3_CLEAR_SEED"),
   fetchWalletPrivateDataFromWallet: () => createSimpleAction("WEB3_FETCH_SEED"),
+  detectWeb3: createActionFactory("WEB3_WALLET_DETECT"),
```
AlexeyTuboltsev: this one is not used. 
and it shouldn't (IMO). It's better to call this saga directly, not to rely on side effects

packages/web/app/lib/web3/browser-wallet/BrowserWalletConnector.ts
```diff
@@ -16,22 +16,36 @@ import {
...
       const injectedWeb3 = (window as any).web3;
       if (typeof injectedWeb3 === "undefined") {
-        throw new BrowserWalletMissingError();
+        return { injectedWeb3Provider: undefined, isLatestMetaMask: false };
       }
-      injectedWeb3Provider = injectedWeb3.currentProvider;
+      const injectedWeb3Provider = injectedWeb3.currentProvider;
+      return { injectedWeb3Provider, isLatestMetaMask: false };
+    }
+  };
```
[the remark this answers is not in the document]
drakkein: It has been changes as you suggested, somehow GitHub is not recognizing change :thinking:

Moejoe90:
This is already merged with at/#3948_metamask_signup

PR #4124, opened 2020-02-27
Fix can't remove media link in the issuer form 

arjunu:


arjunu:
Reopened here https://github.com/Neufund/platform-frontend/pull/4126 cause the PR wasn't updating

PR #4126, opened 2020-02-27
Fix can't remove media link

arjunu:


pavelneu:
Checked for issuer on io and codechecks.
When a media link is disabled, the eto page has no this icon anymore.

PR #4142, opened 2020-03-04
Fix ETO layout alignment issues

arjunu:
### 1.  Fixed pitch pushed to the right when there are no social channels

before:
image

after:
image

## 2. Fixed overflowing images

before:
image

after:
image


packages/web/app/components/shared/Field.module.scss
```diff
@@ -81,4 +81,12 @@
       }
     }
   }
+
+  &::after {
```
drakkein: we can use `clearfix` mixin which is present in our bootstrap or even add `.clearfix` class to Field, but this is good too

arjunu: updated

PR #4147, opened 2020-03-04
Fix ETO not showing presale progress for whitelisted user

arjunu:
- make `watchEtoView` take `previewCode` and retake ETO from state for very poll
- replace single pledge call with loading all pledges
- load all pledges when loading ETOs
- uncache ETO in selectors `selectEto` & `selectEtoWithCompanyAndContract`
- fixed book building not loading for investor because of wrong condition
- remove duplicate ETO watch
- fix ETO not updating during transition


packages/web/app/modules/eto/sagas.ts
```diff
@@ -287,6 +287,7 @@ function* watchEtosSetAction(
   _: TGlobalDependencies,
   action: TActionFromCreator<typeof actions.eto.setEtos>,
 ): any {
+  console.log("watchEtosSetAction");
```
Moejoe90: Hey can you remove this?

packages/web/app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -137,30 +141,34 @@ export function* loadBookBuildingListStats(
...
       selectIsUserFullyVerified,
     );
 
-    if (userType !== EUserType.INVESTOR || !isVerified) return;
+    if (userType !== EUserType.INVESTOR || !isVerified) {
+      return;
+    }
 
-    yield put(actions.bookBuilding.loadBookBuildingStats(etoId));
+    const response: IHttpResponse<IPledge[]> = yield apiEtoPledgeService.getAllMyPledges();
```
Moejoe90:
```suggestion
    const response = yield* call(() => apiEtoPledgeService.getAllMyPledges());
```

packages/web/app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -137,30 +141,34 @@ export function* loadBookBuildingListStats(
...
+    const pledges: IPledges = {};
+    response.body.forEach((item: IPledge) => {
+      pledges[item.etoId!] = item;
+    });
```
Moejoe90: I think this should be handeled in `getAllMyPledges` as when I use this method I shouldn't think about how the response body looks like

packages/web/app/test-e2e/investor/dashboard/utils.ts
```diff
@@ -0,0 +1,8 @@
+import { MOCK_API_URL } from "../../config";
+
+export const goToNextETOState = (etoId: string) =>
```
Moejoe90: could you add documentation here

packages/web/app/test-e2e/investor/dashboard/utils.ts
```diff
@@ -0,0 +1,8 @@
...
+  cy.request({
+    url: MOCK_API_URL + `e2e-tests/eto/${etoId}/time-travel/`,
+    method: "POST",
+    body: { delta: 10 },
```
Moejoe90: Why is the detla 10? is blocks?

packages/web/app/test-e2e/investor/dashboard/ETOList.spec.e2e.ts
```diff
@@ -0,0 +1,53 @@
...
+  it("loads countdown to pre sale for logged in, whitelisted user", () => {
+    const etoId = etoFixtureAddressByName("ETOInSetupState");
+    createAndLoginNewUser({
+      type: "investor",
+      kyc: "business",
+    });
+
+    cy.visit(etoPublicViewByIdLinkLegacy(etoId));
+
+    cy.wait(5000); //let the store get the ETO data, otherwise this place is flaky
```
Moejoe90: this is an `anti` pattern. 

The trick is to `assert` than an element exists once the data is fully loaded. so then you can wait until the component is there based on `tid` once it `exists` then you know the data is added to the store correctly

packages/web/app/components/dashboard/eto-list/EtoList.tsx
```diff
@@ -80,6 +80,7 @@ const EtoList = compose<IStateProps & TExternalProps, TExternalProps>(
   onEnterAction({
     actionCreator: dispatch => {
       dispatch(actions.eto.loadEtos());
+      dispatch(actions.bookBuilding.loadAllPledges());
```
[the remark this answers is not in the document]
arjunu: moved

packages/web/app/components/portfolio/Portfolio.tsx
```diff
@@ -30,6 +30,7 @@ export const Portfolio = compose<TPortfolioLayoutProps, {}>(
   onEnterAction({
     actionCreator: dispatch => {
       dispatch(actions.eto.loadEtos());
+      dispatch(actions.bookBuilding.loadAllPledges());
```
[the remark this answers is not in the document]
arjunu: moved

packages/web/app/lib/api/eto/EtoPledgeApi.interfaces.unsafe.ts
```diff
@@ -15,6 +15,11 @@ export interface IPledge {
   amountEur: number;
   currency: ECurrency.EUR_TOKEN;
   consentToRevealEmail: boolean;
+  etoId?: string;
```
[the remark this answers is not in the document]
arjunu: because when we save a new pledge it doesn't have this property.

packages/web/app/test-e2e/investor/dashboard/utils.ts
arjunu: it's the time in seconds before next state transition. Added a doc comment

packages/web/app/test-e2e/investor/dashboard/utils.ts
arjunu: added

packages/web/app/modules/bookbuilding-flow/sagas.ts
arjunu: moved

packages/web/app/modules/eto/sagas.ts
```diff
@@ -287,6 +283,7 @@ function* watchEtosSetAction(
   _: TGlobalDependencies,
   action: TActionFromCreator<typeof actions.eto.setEtos>,
 ): any {
+  yield fork(watchPledges);
```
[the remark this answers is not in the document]
arjunu: moved to bookbuilding-flow

packages/web/app/test-e2e/investor/dashboard/ETOList.spec.e2e.ts
arjunu: updated

packages/web/app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -137,30 +141,34 @@ export function* loadBookBuildingListStats(
...
+  notificationCenter,
+  logger,
+}: TGlobalDependencies): Generator<any, void, any> {
   try {
-    const { etoId } = action.payload;
-
     const userType: ReturnType<typeof selectUserType> = yield select(selectUserType);
     const isVerified: ReturnType<typeof selectIsUserFullyVerified> = yield select(
       selectIsUserFullyVerified,
     );
```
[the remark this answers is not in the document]
arjunu: changed

packages/web/app/modules/bookbuilding-flow/sagas.ts
arjunu: changed

packages/web/app/modules/eto-view/shared/sagas.ts
```diff
@@ -164,7 +152,7 @@ export function* etoFlowBackwardsCompat(
 ): Generator<any, void, any> {
   // save various pieces of data for backwards compatibility with other flows, e.g. investment
   yield call(saveEto, eto);
-  yield call(saveUsersPledge, eto);
+  yield put(actions.bookBuilding.loadAllPledges());
```
Moejoe90: same comment here regarding yielding directly

packages/web/app/test-e2e/investor/dashboard/ETOList.spec.e2e.ts
```diff
@@ -0,0 +1,72 @@
...
+import { tid } from "../../utils/selectors";
+import {
+  createAndLoginNewUser,
+  loginFixtureAccount,
+  logout,
+  makeAuthenticatedCall,
+} from "../../utils/userHelpers";
+
+describe("ETO list", () => {
+  it("loads countdown to public sale for not logged in user", () => {
```
Moejoe90: you need to add a priority letter here in tests please check the other e2e tests for examples
`#p1` being highest priority and `#p3` being optional

packages/web/package.json
```diff
@@ -40,7 +40,7 @@
     "storybook:build": "(cp -n ./app/config/.env.storybook .env || true) && build-storybook -c .storybook -s ./app/assets",
     "storybook:deploy-ghpages": "storybook-to-ghpages --script=storybook:build --out=storybook-static",
     "storybook:screenshots": "storycap --serverTimeout 200000 --serverCmd \"yarn storybook\" --parallel 8 http://localhost:9001",
-    "test:browser": "NF_MOCHA_RUN=1 TS_NODE_PROJECT=tsconfig.spec.json TS_NODE_IGNORE='node_modules\\/(?!@neufund)' mocha --require ts-node/register/transpile-only --require jsdom-global/register --require ./test/setupBrowserTests.ts './test/setupTestsHooks.ts' './app/**/*.spec.ts?(x)'",
+    "test:browser": "NF_MOCHA_RUN=1 TS_NODE_PROJECT=tsconfig.spec.json TS_NODE_IGNORE='node_modules\\/(?!@neufund)' mocha --exit --require ts-node/register/transpile-only --require jsdom-global/register --require ./test/setupBrowserTests.ts './test/setupTestsHooks.ts' './app/**/*.spec.ts?(x)'",
```
Moejoe90: why did you add `exit` usually this is a bad sign as this means there are tests that are not exiting like an unresolved promise or something please have a look we need to remove `--exit`

packages/web/app/lib/api/eto/EtoPledgeApi.ts
```diff
@@ -19,18 +20,23 @@ export class EtoPledgeApi {
...
 
     if (response.statusCode === 404) {
-      throw new EtoPledgeNotFound();
+      throw new EtoPledgesNotFound();
     }
 
-    return response;
+    const pledges: IPledges = {};
+    response.body.forEach((item: IPledge) => {
+      pledges[item.etoId!] = item;
```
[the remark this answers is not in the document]
arjunu: I created the type, etoId is optional because when we create a new pledge for calling save, there's no ID

packages/web/app/modules/eto/sagas.ts
```diff
@@ -371,6 +367,8 @@ export function* watchEto(_: TGlobalDependencies, previewCode: string): any {
 }
 
 export function* loadEtos({ apiEtoService, logger, notificationCenter }: TGlobalDependencies): any {
+  yield put(actions.bookBuilding.loadAllPledges());
```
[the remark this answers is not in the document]
arjunu: - pledges and etos load parallelly
- we wanted to use actions to communicate between saga modules

packages/web/app/modules/eto-view/shared/sagas.ts
arjunu: so they load parallely

packages/web/app/lib/api/eto/EtoPledgeApi.ts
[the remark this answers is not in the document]
arjunu: The type did not have this property, I added it see https://github.com/Neufund/platform-frontend/pull/4147/files#diff-154d2168fbddc8ab492c30b24440b8bdR18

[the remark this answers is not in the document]
rudolfix: you should probably have a base type then derived write type and read type. but living with optional etoId is also possible.

packages/web/app/modules/eto/sagas.ts
```diff
@@ -165,7 +165,7 @@ function* loadAdditionalEtoData(
   );
 
   if (shouldLoadPledgeData(eto.state, onChainState)) {
-    yield put(actions.bookBuilding.loadPledge(eto.etoId));
```
rudolfix, line 168: you should not reload all pledges when in the context of particular ETO. you should just load one single pledge associated with that eto.

also you put a condition in `shouldLoadPledgeData` that pertains only to this very eto, but then you reload all pledges which logically should not/should be excluded

please also that by doing that you will periodically poll for all pledges for all etos on the dashboard. this should be done only once with separate watcher

rudolfix: what IMO you should do: 
1. remove loading pledges from additional eto data. it does not belong here
2. for dashboard (or the whole app) use periodic watcher
3. in context of particular eto being display (public profile) restore single pledge loading and use it

packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -48,10 +48,8 @@ export const shouldLoadPledgeData = (
...
   onChainState?: EETOStateOnChain,
 ): boolean =>
-  !!(
-    [EEtoState.LISTED, EEtoState.PROSPECTUS_APPROVED, EEtoState.ON_CHAIN].indexOf(etoState) >= 0 &&
-    (onChainState === undefined || onChainState < EETOStateOnChain.Claim)
-  );
+  [EEtoState.LISTED, EEtoState.PROSPECTUS_APPROVED, EEtoState.ON_CHAIN].indexOf(etoState) >= 0 &&
+  (onChainState === undefined || onChainState < EETOStateOnChain.Claim);
 
 export const shouldLoadBookbuildingStats = (onChainState: EETOStateOnChain | undefined): boolean =>
```
rudolfix, line 54: this condition should be same as load pledges below. we should be able to load bookbuilding before on chain state as well.

packages/web/app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -177,7 +182,8 @@ export function* bookBuildingFlowSagas(): Generator<any, any, any> {
     actions.bookBuilding.loadBookBuildingListStats,
     loadBookBuildingListStats,
   );
-  yield fork(neuTakeEvery, actions.bookBuilding.loadPledge, loadMyPledge);
+  yield fork(neuTakeEvery, actions.bookBuilding.loadAllPledges, loadAllMyPledges);
```
rudolfix, line 185: this comment applies to `watchBookBuildingStats`, in the body in the saga we do not use `watchBookBuildingStats`. why? do we use it on starting the watch? but I think this saga starts it unconditionally (sorry, something I lack react skills :))

packages/web/app/modules/eto/sagas.ts
arjunu: I thought we were moving away from single load pledge.
- I restored it and used it for ETO public view and watch ETO
- Load all pledges watch runs for the whole app
- removed loading pledges from additional eto data

packages/web/app/modules/bookbuilding-flow/sagas.ts
arjunu: `watchBookBuildingStats` is started conditionally if `bookBuildingEnabled` is `true` else we just load bookBuildingStats just once

packages/web/app/modules/bookbuilding-flow/utils.ts
arjunu: this is no longer needed, removed

packages/web/app/lib/api/eto/EtoPledgeApi.ts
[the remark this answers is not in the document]
arjunu: kept this unchanged for now

packages/web/app/modules/eto/sagas.ts
rudolfix: > I restored it and used it for ETO public view and watch ETO

If we have a global watch for pledges we should not watch single pledge on eto. let's decouple those two

packages/web/app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -136,24 +136,22 @@ export function* loadBookBuildingListStats(
...
-    const userType: ReturnType<typeof selectUserType> = yield select(selectUserType);
-    const isVerified: ReturnType<typeof selectIsUserFullyVerified> = yield select(
-      selectIsUserFullyVerified,
-    );
-
-    if (userType !== EUserType.INVESTOR || !isVerified) return;
-
-    yield put(actions.bookBuilding.loadBookBuildingStats(etoId));
+    const canLoadPledgesResult = yield* call(canLoadPledges);
+    if (canLoadPledgesResult) {
```
rudolfix, line 147: this is exactly the case where `shouldLoad` for pledge and bookbuilding is identical if we load those for particular eto. here of course we have additional condition if user is fully verified and is an investor. so yeah we should check it but next we should call `shouldLoad` and also apply it.

packages/web/app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -164,6 +162,35 @@ export function* loadMyPledge(
...
 
+export function* loadAllMyPledges({
+  apiEtoPledgeService,
+  notificationCenter,
+  logger,
+}: TGlobalDependencies): Generator<any, void, any> {
+  try {
+    const canLoadPledgesResult = yield* call(canLoadPledges);
+
+    if (canLoadPledgesResult) {
```
rudolfix, line 173: here it's all right! to load all pledges we just need to make sure if the user is investor and fully verified

packages/web/app/modules/eto/selectors.ts
```diff
@@ -129,9 +109,19 @@ export const selectInvestorEtoWithCompanyAndContract = (
...
   previewCode: string,
 ): TEtoWithCompanyAndContractReadonly | undefined => {
-  const eto = selectEto(state, previewCode);
+  const eto = selectEtoState(state).etos[previewCode];
+
   if (eto) {
-    return selectEtoWithCompanyAndContractInternal(state, eto);
+    const contract = selectEtoContract(state, eto.previewCode);
+    const company = nonNullable(selectCompany(state, eto.companyId));
+    const subState = selectEtoSubState(state, eto);
```
rudolfix, line 117: this one is still cached

packages/web/app/modules/bookbuilding-flow/sagas.ts
arjunu: I have added this condition from the place where `actions.bookBuilding.loadPledgeForEto` is dispatched.
in `packages/web/app/modules/eto-view/shared/sagas.ts`
So it looks like

```
if (shouldLoadBookbuildingStats(onChainState)) {
    yield put(actions.bookBuilding.loadPledgeForEto(eto.etoId));
    yield put(actions.bookBuilding.loadBookBuildingStats(eto.etoId));
 }
```

packages/web/app/modules/eto/sagas.ts
```diff
@@ -273,104 +266,9 @@ export function* loadEtoContract(
   return { ...eto, contract };
 }
 
-function* watchEtoSetAction(
```
arjunu, line 276: moved to `packages/web/app/modules/eto/sagas/watchEtosSetActionSaga.ts`

packages/web/app/modules/eto/selectors.ts
arjunu: uncached

packages/web/app/modules/eto/sagas.ts
arjunu: so 
- dashboard: all pledges are loaded along with all etos
- eto page: single pledge loaded  
- watch pledges: runs on both pages but only after delay. So first load comes from either all pledges or single pledge, 2nd load onwards comes from watch.

packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -44,15 +44,6 @@ export const isPledgeAboveMinimum = (minPledge: number): TestOptions => ({
   },
 });
 
-export const shouldLoadPledgeData = (
```
rudolfix, line 47: this is a good sign that it's removed!

packages/web/app/modules/bookbuilding-flow/utils.ts
rudolfix: ohhh this one is needed still and used! `shouldLoadPledgeData` was removed

btw. this is the right condition for loading bookbuilding stats
```
 !!(
    [EEtoState.LISTED, EEtoState.PROSPECTUS_APPROVED, EEtoState.ON_CHAIN].indexOf(etoState) >= 0 &&
    (onChainState === undefined || onChainState < EETOStateOnChain.Claim)
  );
```

arjunu:
> - whitelisted ETO card is still in “presale". State nor even the investment done from the same account is updated
> - In the network there are multiple calls for a single ETO

These two issues were related to the watch code and have been fixed.

> whitelisted widget reloads but I still can see “waiting for the new block” below the invest button
This issue is with `EndTimeWidget` where on finish logic is not implemented. Will do this as a separate ticket.

packages/web/app/modules/eto/selectors.ts
desfero: what's the reason to uncache here? It's very important to return the same reference when there was no change in internal state as otherwise we gonna have performance impact on rerendering the whole app every time we update the store given that we will return the new eto object on each store update (even when the store update didn't touch that particular eto state) That's the main reason reselect exists and we should be always careful with selectors that return in some way mapped non primitive types.

packages/web/app/components/dashboard/my-wallet/MyWalletWidget.spec.tsx
```diff
@@ -7,7 +7,8 @@ import * as React from "react";
 import { wrapWithBasicProviders } from "../../../../test/integrationTestUtils.unsafe";
 import { MyWalletWidgetComponentLayout } from "./MyWalletWidget";
 
-describe("<MyWalletWidget />", () => {
+// TODO re-write without running sagas
+describe.skip("<MyWalletWidget />", () => {
```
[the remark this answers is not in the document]
arjunu: Unskipped

packages/web/app/modules/bookbuilding-flow/sagas.ts
```diff
@@ -136,24 +138,26 @@ export function* loadBookBuildingListStats(
...
-    if (userType !== EUserType.INVESTOR || !isVerified) return;
-
-    yield put(actions.bookBuilding.loadBookBuildingStats(etoId));
-
-    const pledgeResponse: IHttpResponse<IPledge> = yield apiEtoPledgeService.getMyPledge(etoId);
-    yield put(actions.bookBuilding.setPledge(etoId, pledgeResponse.body));
+    const canLoadPledgesResult = yield* call(canLoadPledges);
+
+    if (canLoadPledgesResult) {
+      const state = yield* select();
```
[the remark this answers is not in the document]
arjunu: changed

packages/web/app/modules/eto/selectors.ts
arjunu: previously we had a loop and it would call the cached selector outside the loop and the eto reference would remain same for all iterations. We no longer have the loop but I kept the uncached selector because I didn't notice any performance change on the dashboard. 

Anyways now I tested with the uncached selector and all good's so reverted this change.

packages/web/app/lib/api/eto/EtoPledgeApi.ts
```diff
@@ -32,6 +34,25 @@ export class EtoPledgeApi {
...
+      allowedStatusCodes: [404],
+    });
+
+    if (response.statusCode === 404) {
+      throw new EtoPledgesNotFound();
+    }
+
+    const pledges: IPledges = {};
+    response.body.forEach((item: IPledge) => {
+      pledges[item.etoId!] = item;
```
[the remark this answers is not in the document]
arjunu: kept this unchanged

packages/web/test/integrationTestUtils.unsafe.tsx
```diff
@@ -230,42 +229,6 @@ export async function waitUntilDoesntThrow(
   }
 }
 
-interface ICreateProviderContext {
-  container?: Container;
-  store?: Store<any>;
-  history?: History;
-}
-
-export function wrapWithProviders(
```
desfero, line 239: <3

PR #4148, opened 2020-03-05
Fix ETO view merge misses

Moejoe90:


packages/web/app/components/eto/overview/EtoOverviewStatus/utils.tsx
```diff
@@ -0,0 +1,56 @@
...
+import * as React from "react";
+import { FormattedMessage } from "react-intl-phraseapp";
+
+import { EOfferingDocumentType } from "./../../../../lib/api/eto/EtoProductsApi.interfaces";
+
+export enum EDocumentType {
+  PROSPECTUS_APPROVED,
+  PROSPECTUS,
+  INVESTMENT_MEMORANDUM,
+}
```
AlexeyTuboltsev: I'd move types to a `types.ts` somewhere in a central place

Moejoe90: I would keep here now because its only used in `utils` as a general idea you're right but this is done to remove the coupeling between translations and the util function

arjunu: if it's only used by this component, I too would prefer to keep it here.

Issue #4141, opened 2020-03-03
Missing assets while logged starting from lading page


drakkein:
**Description**
If you login in starting from landing page (wait for Investment opportunities to load) you will see that you have no assets in your portfolio for few seconds.

**Environment**
Probably all

**Steps To Reproduce**
Go to landing page, click login and login as any account which has some tokens in portfolio. 

**Actual behaviour**
`You have no assets in your portfolio yet.` message is being displayed. 

**Expected behaviour**
Show loading indicator until all data is ready. 

**Comments, context, logs**
We are assuming that ETO's are loading while `displayOrder` from etos state is set as undefined (`packages/web/app/modules/eto/sagas.ts:430`). If we start from landing page and wait for ETO's to load we have populated state, which never gonna be erased, but we do not have token data loaded. In that case we are no showing loading indicator anymore, but as we do not have detailed data, we also show that user have no assets (`selectMyAssetsWithTokenData` selector). 

**Screenshots**
![image](https://user-images.githubusercontent.com/10601291/75820059-f9c56a00-5d9b-11ea-9968-b052a16cea68.png)

glupiawymowka:
tested on stage

PR #4149, opened 2020-03-05
#4141 Add loading state to tokens

drakkein:
This PR adds additional property to state to have source of truth for Tokens loader


packages/web/app/modules/eto/sagas.ts
```diff
@@ -622,14 +619,20 @@ function* loadToken(
...
 
-  yield all(myAssets.map((eto: TEtoWithCompanyAndContractReadonly) => neuCall(loadToken, eto)));
+  yield all(etos.map((eto: TEtoWithCompanyAndContractReadonly) => neuCall(loadToken, eto)));
+
+  yield put(actions.eto.setTokensLoadingDone());
+}
+
+function* loadTokensData(): any {
+  const myAssets = yield select(selectMyAssets);
+
```
Moejoe90: at this stage there is no way for us to guarentee that `myAssets` exist and is updated
I just checked where `loadTokensData` is used lets remove `loadTokensData` and use `loadEtos` instead since `loadEtos` should load the token data correctly

packages/web/app/components/dashboard/my-portfolio/PortfolioStats.tsx
```diff
@@ -162,9 +165,9 @@ const PortfolioStats = compose<TPortfolioStatsProps, {}>(
   }),
   withContainer(PortfolioStatsLayoutContainer),
   branch<TStateProps>(state => state.hasError, renderComponent(PortfolioStatsErrorLayout)),
-  branch<TStateProps>(state => !state.myAssets, renderComponent(LoadingIndicator)),
+  branch<TStateProps>(state => state.isLoading, renderComponent(LoadingIndicator)),
   branch<TStateProps>(
-    state => !!state.myAssets && state.myAssets.length === 0,
+    state => !state.myAssets || state.myAssets.length === 0,
```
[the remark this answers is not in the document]
drakkein: After transpiling we will get something like ```var _state$myAssets;

!!((_state$myAssets = state.myAssets) === null || _state$myAssets === void 0 ? void 0 : _state$myAssets.length) === 0;```

so in case of myAssets being undefined we have conditions `undefined === 0` which will return `false` and will throw us `Error detail: Non nullable values expected, received undefined` furhter.

packages/web/app/modules/eto/sagas.ts
drakkein: Removed as requested

packages/web/app/test-e2e/investor/portfolio/PortfolioStats.spec.e2e.io.ts
```diff
@@ -0,0 +1,23 @@
+import { appRoutes } from "../../../components/appRoutes";
+import { lightWalletTypePasswordAndLogin } from "../../utils/index";
+import { tid } from "../../utils/selectors";
+import { DEFAULT_PASSWORD, loginFixtureAccount } from "../../utils/userHelpers";
+
+describe("Portfolio Stats", () => {
+  it("should show loading without no assets message when logged in starting from landing @dashboard @portfolio-stats @p3", () => {
```
Moejoe90: don't forget the tags are with `#` not `@` symbol

Issue #4153, opened 2020-03-05
Missing logged user check after ETO_RELOAD_ETO_DATA


drakkein:
**Description**
After reloading ETO data we lack of condition to check if user is logged in in order to load investor ticket. 

**Environment**
PROD (all)

**Steps To Reproduce**
Go to ETO page, ex. https://platform.neufund.org/eto/view/LI/1eb004fd-c44d-4bed-9e76-0e0858649587, wait, check sentry.
This error is not critical and do not break sagas chain, so we are not displaying any error page. 

**Actual behaviour**
Saga is not checking if user is logged in 

**Expected behaviour**
Saga should check if user is logged in and fetch investor ticket according to it's state

**Comments, context, logs**
https://sentry.neusrv.net/sentry/production/issues/12176/?referrer=slack

**Screenshots**

glupiawymowka:
tested on stage

PR #4155, opened 2020-03-05
#4153 Fix ETO reload view data saga

drakkein:
This PR adds check for logged user in order to perform loading investor ticket.


packages/web/app/modules/eto-view/shared/sagas.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { expectSaga } from "@neufund/sagas/tests";
```
Moejoe90: :+1:

Issue #4004, opened 2020-02-08
[WIP] WalletConnect - Transaction Signer


rudolfix:
**Background**
User is able to use current desktop app and just sign transactions and messages with the Neufund app.
There’s standard protocol: Wallet Connect that implements that (https://docs.walletconnect.org/tech-spec). 
2FA signer would be session based. User would log in with QR code and that would create temporary session during which app is accepting messages from the desktop. Session would be broken on logout or after some inactivity.

**Desktop app**
In principle we should be able to implement new web3 wallet-connect provider on the desktop app (there is reference implementation: @banciur knows) and use it without any special modifications for the desktop app. On the desktop we'll also support connecting any other wc-compatible wallet which are many now. 

We must minimize commands sent to connected wc-wallet. in fact the only commands we may send are: `accounts`, `eth_sign` and `eth_signTransaction` so we'll need to reimplement reference web3 provider

We should break wallet connect session on logout from mobile app.

**Mobile App**
After establishing a wc-link, mobile app receives standard Ethereum JSON-RPC messages on the websocket.
* [ ] there are special messages `wc_session` that handle wc session establishment, upgrade and unlink.
* [ ] there are `eth_` and other standard JSON-RPC commands. those you just forward to web3 as implemented in #4002. mind that `eth_sign` and `eth_signTransaction` will require usage of private key and must be able to display signer dialogs to obtain user confirmation.
* [ ] when websocket is not connected, firebase notifications will be sent on which app must be reactivated and websocket connected for json rpc commands delivery

**Notifications**
* WalletConnect protocol uses **subscribe** method to connect any notification server. We can still use this mechanism - it’s however a way to easily DOS our bridge server so we need to whitelist notification servers
* In case of our Mobile app we’ll need this method to be called with **user id, device id and topic id**. That will start forwarding notifications to the client by backend component. See #4003 for details on linking uid, device id to firebase registration token. 
* As there’s just one active notification and we should expire notifications after one minute we should overwrite them in UI (https://stackoverflow.com/questions/39782438/can-fcm-notification-on-android-overwrite-previous-one).
* We should expire them on the receiving end - there are standard methods.

**Security Considerations**
1. Hijacking the session by reading session key from QR code displayed on the desktops and connecting before the genuine user. Here the consequences are not severe - desktop will connect to someone else's wallet so not much harm can be done
[...]

Issue #4005, opened 2020-02-09
Handling QR Codes


rudolfix:
**Background**
App needs to understand several QR code types and it should be contextless. So user scans QR code from top level icon and desired action happens. Standard library should handle QR codes end to end: so QR code is read, parsed and depending on QR code type, right handler is called. This is to have control over reading and parsing QR codes which are attack vector on our app.

Deep linking requirements will be described separately and they are not a part of first milestone.

**Security Considerations**
We should be wary of malicious QR codes that will for example import fake light wallet or establish a session with malicious pseudo desktop that will issue fake transactions. That would be a kind of phishing attack. User would scan malicious QR code that will initiate a wallet connect session and then user would receive malicious transactions to sign.

QR code scanning and parsing is very sensitive as this is uncontrolled input into the app and may  be used to exploit bugs (people will use fuzzers and spend hours to mutate our QR codes until they find a crack. https://github.com/MalQR/MalQR.github.io)

**Acceptance criteria**
* [x] we choose a decent QR code library. AFAIK both iOS/Android have QR code scanning and parsing as part of the standard library. There are react wrappers on those.
* [x] we scan and parse QR code data, ignoring unknown QR code types. we should further limit parsed QR code type: for example if user wants to send ETH but parsed QR code is wallet connect, we should ignore it
* [x] we support `wc:` uri of wallet connect
* [ ] (not a part of MVP) we support `ethereum:` for ETH transfers: https://github.com/ethereum/EIPs/issues/67 (we have it on the frontend)
* [ ] (not a part of MVP) optionally we may support ERC20 token transfer extension https://eips.ethereum.org/EIPS/eip-681

**Testing notes**
We should make sure QR code scanning gracefully handle unknown QR codes. For now, we only support wallet connect so there is not that much to check.
- [x] **Given** the user is on `Scan QR code` screen **when** he tries to scan unknown QR code **then** an `Alert` appears with information that QR code is not supported.
- [x]  **Given** the user is on `Neufund Web` screen (Profile -> Neufund Web) **when** he taps on Scan Qr code tries to scan unknown QR code that's not a wallet connect QR code **then** an `Alert` appears with information that wallet connect QR code is required.

Issue #4002, opened 2020-02-06
Expose Cryptographic Function via web3 wallet interface


rudolfix:
**Background**
Expose cryptographic functions from #4001 as part of web3 wallet. Seems it makes sense for code reuse.
* `eth_sign`
* `eth_sendTypedTransaction`

We'll use last one to pass additional metadata on transaction - exactly the same data that we use to display tx details on the desktop.

and

* `eth_sendTransaction` - this we enable for backward compatibility only for app connected to development network/in development mode

**Message/Transaction validation**
Internal web3 provider will perform additional checks on the data passed to sign.
* in case of message signing only the documents with supported schema and known permissions can be signed.
* in case of tx signing only the known txs can be signed: for which metadata is present and it matches to decoded tx data

proper implementation of this will require a "hardcoded" registry of possible transactions with associated metadata schemas and tx validation rules. sound like what we anyway need to do

PR #4178, opened 2020-03-11
#4004 wallet connect transaction signer

desfero:
This pr adds basic handling of wallet connect requests.

What's not included:
- [ ] Persistent storage for wallet connect session
- [ ] Handling of `eth_sendTypedTransaction`
- [ ] Proper user interface
- [ ] Timeouts (session request timeout, inactivity timeout)
- [ ] SSL and bridge pinning 
- [ ] Selectors state is typed as any. We will move it as a cleanup ticket as this needs some sync with the web team.
- [ ] I would like to add some module-level integration tests (where for .e.g wallet connect module is just a black box and testing is just done through dispatching actions and asserting proper state in selectors) This is something we wanted to add in the web team for years already but never found some time to do so. Given that we now have a proper module approach of sagas we can try to write such kinds of integrations tests. Will move it also as a clean up ticket.

TODO:

- [x] Clean up interfaces
- [x] Add unit tests

DEMO:
ezgif-6-4d6b5db33e4c


packages/wallet/app/modules/wallet-connect/lib/WalletConnectManager.ts
```diff
@@ -0,0 +1,276 @@
...
+
+        const peerData = payload.params[0];
+
+        const peer: TWalletConnectPeer = {
+          id: peerData.peerId,
+          meta: peerData.peerMeta,
+        };
+
+        const approveSession = (chainId: number, address: EthereumAddress) => {
+          this.walletConnect.approveSession({
```
desfero: When we have frontend wallet connect implementation ready some rework gonna be needed here to avoid two signing modals (session signer and `eth_sign` with challenge signer).

packages/shared-modules/src/types.ts
```diff
@@ -1,11 +1,16 @@
 import { ISagaModule } from "@neufund/sagas";
 import { Opaque, OpaqueType, ReturnTypeStrict, Values } from "@neufund/shared";
 import { ContainerModule } from "inversify";
-import { ActionCreatorsMapObject } from "redux";
+import { ActionCreatorsMapObject, ReducersMapObject } from "redux";
 
 interface INeuModule<S> extends ISagaModule<S> {
   libs?: ContainerModule[];
```
DmitriyNoa: Should all the props be optional? It seems a bit strange

packages/shared-modules/src/utils.ts
```diff
@@ -82,4 +92,12 @@ function appConnect<
   );
 }
 
-export { createLibSymbol, appConnect, TAppConnectOptions };
+function* neuGetBindings<B extends Dictionary<TLibSymbol<any>>>(
```
DmitriyNoa: Is it possible to avoid any here ?

packages/wallet/app/App.tsx
```diff
@@ -1,7 +1,14 @@
...
 
 import { AppRouter } from "./AppRouter";
+import { SignerModal } from "./components/signer/SignerModal";
 
-const App: React.FunctionComponent = () => <AppRouter />;
+const App: React.FunctionComponent = () => (
+  <>
+    <AppRouter />
+
+    <SignerModal />
```
DmitriyNoa: I guess we can rename this component simple for Modal as it will be used across multiple places. Maybe we can rename it later. Up to you

packages/wallet/app/components/shared/qr-code-scanner/QRCodeScanner.tsx
```diff
@@ -0,0 +1,22 @@
+import React from "react";
+import { useWindowDimensions } from "react-native";
+import RNQRCodeScanner from "react-native-qrcode-scanner";
+
+type TExternalProps = {
+  onRead: (uri: string) => void;
```
DmitriyNoa: We would need to add some basic validation to check that URL is valid. Or it should be provided in onReadCallback ?

packages/wallet/app/components/signer/SignerModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+  data,
+  approve,
+  reject,
+}) => (
+  <Modal animationType="slide" transparent={false} visible={state !== ESignerUIState.IDLE}>
+    <View style={{ paddingVertical: 40, paddingHorizontal: 20, flex: 1, justifyContent: "center" }}>
+      <Text>
+        Signing state: {state}
+        {"\n"}
+        Signing data: {JSON.stringify(data, undefined, 2)}
```
DmitriyNoa: Looks like react-intl is available for react native. Let's start using it instead of simple strings?

packages/wallet/app/components/signer/SignerModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const SignerModalLayout: React.FunctionComponent<TStateProps & TDispatchProps> = ({
+  state,
+  data,
+  approve,
+  reject,
+}) => (
+  <Modal animationType="slide" transparent={false} visible={state !== ESignerUIState.IDLE}>
+    <View style={{ paddingVertical: 40, paddingHorizontal: 20, flex: 1, justifyContent: "center" }}>
```
DmitriyNoa: Can you move styles outside of render? So it's not recreated on every rerender

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -35,10 +35,10 @@ function* initStartSaga({ logger, ethManager }: TGlobalDependencies): Generator<
 
     yield ethManager.sendTransaction({
       to: toEthereumAddress("0x7824e49353BD72E20B61717cf82a06a4EEE209e8"),
-      gasLimit: utils.bigNumberify("21000"),
-      gasPrice: utils.bigNumberify("20000000000"),
+      gasLimit: "0x21000",
```
DmitriyNoa: Why did you switch to strings?

packages/wallet/app/modules/notification-ui/actions.ts
```diff
@@ -0,0 +1,10 @@
+import { createActionFactory } from "@neufund/shared";
```
DmitriyNoa: Let's think about naming. We now have Notifications here and Notifications for push notifications. It might be confusing for a developer outside.

packages/wallet/app/modules/notification-ui/sagas.ts
```diff
@@ -0,0 +1,23 @@
...
+import { TActionFromCreator } from "../../store/types";
+import { notificationUIActions } from "./actions";
+
+function* showInfo(
+  _: TGlobalDependencies,
+  action:
+    | TActionFromCreator<typeof notificationUIActions, typeof notificationUIActions.showInfo>
+    | TActionFromCreator<typeof notificationUIActions, typeof notificationUIActions.showError>,
+): Generator<any, void, any> {
+  Alert.alert(action.payload.message);
```
DmitriyNoa: Does this really should be a native alert ?

DmitriyNoa: Can we avoid "any"s here ?

packages/wallet/app/modules/wallet-connect/lib/WalletConnectManager.ts
```diff
@@ -0,0 +1,276 @@
...
+          if (!WalletConnectEthSignJSONRPCSchema.isValidSync(payload)) {
+            this.handleCallSigningError(payload.id, new InvalidJSONRPCPayloadError(payload.method));
+
+            return;
+          }
+
+          this.handleCallSigningRequest(
+            EWalletConnectManagerEvents.SIGN_MESSAGE,
+            payload.id,
+            payload.params[1],
```
DmitriyNoa: Maybe I didn't find it. Can you add a type of payload.params (or maybe as a doc)? It's difficult to understand what payload params array is

packages/wallet/app/modules/wallet-connect/sagas.ts
```diff
@@ -0,0 +1,208 @@
...
+import { walletConnectActions } from "./actions";
+import { privateSymbols } from "./lib/symbols";
+import { TWalletConnectManagerEmit } from "./lib/types";
+import { InvalidWalletConnectUriError, isValidWalletConnectUri } from "./lib/utils";
+import { EWalletConnectManagerEvents, WalletConnectManager } from "./lib/WalletConnectManager";
+import { navigate } from "../../routeUtils";
+
+function* connectToURI(
+  _: TGlobalDependencies,
+  action: TActionFromCreator<
```
DmitriyNoa: Let's try to avoid "any"s in generators as much as possible

packages/wallet/app/modules/wallet-connect/sagas.ts
```diff
@@ -0,0 +1,208 @@
...
+        assertNever(moduleAction.type, "Invalid action");
+        break;
+    }
+  }
+}
+
+function* connectWalletConnectEvents(wcManager: WalletConnectManager): Generator<any, void, any> {
+  const channel = reduxify<TWalletConnectManagerEmit>(wcManager);
+
+  while (true) {
```
DmitriyNoa: Could you please explain why do we need this potentially infinite loop?

DmitriyNoa: It might lead to some issues or  memory leaks later

packages/wallet/ios/wallet.xcodeproj/project.pbxproj
```diff
@@ -289,8 +289,12 @@
 				TargetAttributes = {
 					00E356ED1AD99517003FC87E = {
 						CreatedOnToolsVersion = 6.2;
+						DevelopmentTeam = X83WK44Z7M;
```
DmitriyNoa: What is this development team id? Maybe shouldn't be pushed here?

packages/wallet/app/modules/wallet-connect/lib/WalletConnectManager.ts
rudolfix: I have a solution! mind that you already have a valid JWT in the mobile app. so if user approves connection and then you are asked to sign "generic" permission (we may drop `tos` permission btw.) you can just return JWT you have. and it will be shared by mobile and desktop which IMO is fine.

packages/wallet/app/components/QRCode.tsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+const QRCodeLayout: React.FunctionComponent<TDispatchProps> = ({ processURI }) => (
+  <QRCodeScanner onRead={processURI} />
+);
+
+const QRCode = compose<TDispatchProps, {}>(
+  appConnect<{}, TDispatchProps>({
+    dispatchToProps: dispatch => ({
+      processURI: uri => dispatch(walletConnectActions.connectToPeer(uri)),
```
rudolfix: this will be dispatched to some other component that will dispatch further based on uri? what happens if uri does not start with `wc:`? 
also please note that sometimes we can start QR code scanning with specific intent: like we only accept wc session uri or only ETH Send `ethereum:` URI.
as per https://github.com/Neufund/platform-frontend/issues/4005

do you want to touch this now?

packages/wallet/app/modules/eth/lib/types.ts
```diff
@@ -6,24 +6,37 @@ export enum EBlockTag {
...
-  gasLimit: TBigNumberVariants;
+  gasLimit: string;
   // TODO: hide under `EthManager` so there is not need to pass it from saga
-  gasPrice: TBigNumberVariants;
+  gasPrice: string;
+  data?: string;
+  value?: string;
+};
+
+export type TUnsignedTransaction = {
```
rudolfix: we do not have Ethereum types? in `ethers`?

packages/wallet/app/modules/signer-ui/sagas.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const payload = action.payload;
+
+  try {
+    // wait until signing request get's approved by the user
+    yield take(signerUIActions.approved);
+
+    switch (payload.signerType) {
+      case ESignerType.WC_SESSION_REQUEST: {
+        const address = yield ethManager.getPluggedWalletAddress();
```
rudolfix: it's good and clean but temporary solution. what we should do is:
- wrap ethManager into our wallet system (like LightWallet) and use it as any other wallet. see https://github.com/Neufund/platform-frontend/issues/4002
- move our tx flow into shared component and instead of calling ethManager directly, we should launch tx saga from confirmation dialog like here https://github.com/Neufund/platform-frontend/issues/4193
- signer and tx modals above should be automatically generated from tx typed data

BUT we should definitely merge this!
Good enough for MVP and when we have typed tx data we can generate dialogs as above. and at the very end merge with existing tx saga

packages/wallet/app/modules/wallet-connect/lib/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+  const result = parseQueryString(queryString);
+
+  return {
+    key: result.key ?? "",
+    bridge: result.bridge ? decodeURIComponent(result.bridge) : "",
+  };
+};
+
+const parseWalletConnectUri = (uri: string): IParseURIResult => {
+  const pathStart: number = uri.indexOf(":") + 1;
```
rudolfix: this is very unsafe! please use a decent URI parser library! and extract protocol host query parameters etc from it!
for example here you assume that `:` exists in this uri and you add 1 to `indexOf` which will result in 0 if protocol is not there!

as this is quite important part, could we have a util class where we store all supported uri schemas and associated parsers? so we do not have this kind of sensitive places spread all over code base

packages/wallet/app/modules/wallet-connect/lib/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const protocol: string = uri.substring(0, pathStart - 1);
+
+  if (protocol !== WC_PROTOCOL) {
+    throw new InvalidWalletConnectUriError();
+  }
+
+  const path: string = uri.substring(pathStart, pathEnd);
+
+  const requiredParams: IRequiredParamsResult = parseRequiredParams(path);
```
rudolfix: yeah, move it to a separate place where we'll also add `ethereum` uri parser and anything we need to support in the future

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -1,16 +1,16 @@
...
 import { initActions } from "./actions";
 
-function* initStartSaga({ logger, ethManager }: TGlobalDependencies): Generator<any, void, any> {
+function* initStartSaga({ logger, ethManager }: TGlobalDependencies): Generator<unknown, void> {
   try {
     // TODO: Provide a proper init flow
 
-    const balance = yield ethManager.getBalance(
-      toEthereumAddress("0x429123b08DF32b0006fd1F3b0Ef893A8993802f3"),
+    const balance = yield* call(() =>
```
rudolfix: this saga is full of test code?

PR #4183, opened 2020-03-12
Changes Web Package Webpack settings

Moejoe90:
**Testing notes**
#p3


packages/web/webpack/webpack.config.prod.js
```diff
@@ -24,4 +34,101 @@ module.exports = merge.smart(configCommon, {
...
+  output: {
+    filename: "[chunkhash].[name].min.js",
+  },
+  plugins: [
+    new CleanWebpackPlugin(),
+    new MiniCssExtractPlugin({
+      filename: "[contenthash].[name].css",
+    }),
+  ],
+  module: {
```
arjunu: module.rules can be shared

packages/web/webpack/webpack.config.ci.prod.js
```diff
@@ -1,18 +1,16 @@
...
-const configCommon = require("./webpack.config.common");
+const path = require("path");
 const paths = require("./paths");
 
-// CircleCI is starting to run out of memory due to minimization processes like Terser
-//
+const configCommon = require("./webpack.config.common");
 
+/**
+ * @webpack a special webpack configuration that should run only on CI the minimization process is disabled
```
desfero, line 11: can we add some information why do we need to disable the minifier on ci? I still don't think we should do that but don't' have time to research the reason behind build slowdown

packages/web/webpack/webpack.config.prod.js
```diff
@@ -2,10 +2,20 @@ const TerserPlugin = require("terser-webpack-plugin");
...
+const { CleanWebpackPlugin } = require("clean-webpack-plugin");
 
-const configCommon = require("./webpack.config.ci.prod");
+const path = require("path");
+const paths = require("./paths");
 
+const configCommon = require("./webpack.config.common");
+
+/**
+ * @webpack a special webpack configuration that should used on when deploying production code
```
desfero, line 14: What's the reason of `@webpack` ? Is it custom modifier for tsdocs?

packages/web/webpack/webpack.config.prod.js
```diff
@@ -2,10 +2,20 @@ const TerserPlugin = require("terser-webpack-plugin");
 const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
 const webpack = require("webpack");
 const merge = require("webpack-merge");
+const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+const { CleanWebpackPlugin } = require("clean-webpack-plugin");
 
-const configCommon = require("./webpack.config.ci.prod");
+const path = require("path");
```
desfero, line 8: `path` can be moved to thrid-party dependencies section

packages/web/webpack/webpack.config.ci.prod.js
Moejoe90: Its not even a build slowdown, its full out of memory issues. 
I can add another line but really there is not much to explain

packages/web/webpack/webpack.config.prod.js
Moejoe90: Its a webpack configration why not have it a custom modifier? is there any reason against that

packages/web/webpack/webpack.config.common.module.js
```diff
@@ -0,0 +1,88 @@
+const path = require("path");
+const paths = require("./paths");
+
+export const commonModuleConfig = {
```
Moejoe90, line 4: I didn't want to use `merge.smart` As i'm not sure how it works under the hood. that will require proper research and time

packages/design-system/src/components/buttons/Button.tsx
```diff
@@ -102,7 +102,7 @@ const Button = React.forwardRef<HTMLButtonElement, TButtonProps>(
   ) => {
     const withIconOnly = children === undefined;
 
-    if (process.env.NODE_ENV === "development" || process.env.NF_CYPRESS_RUN === "1") {
```
desfero, line 105: NF_CYPRESS_RUN is specific env variable for the web and given that it's not useful in the design-system build given it's always is unset

packages/web/app/config/getConfig.ts
```diff
@@ -20,32 +20,38 @@ export interface IConfig {
...
 
-export function getConfig(env: NodeJS.ProcessEnv): IConfig {
-  verifyFeatureFlags(env);
+export function getConfig(): IConfig {
+  verifyFeatureFlags();
 
   return {
     ethereumNetwork: {
-      rpcUrl: getRequiredEnv(env, "NF_RPC_PROVIDER"),
+      rpcUrl: getRequiredEnv(process.env.NF_RPC_PROVIDER, "NF_RPC_PROVIDER"),
```
desfero, line 28: `env, "NF_RPC_PROVIDER"` with the following approach webpack is not able to provide a static value for the env variable and it ends up with injecting whole env object as a string

packages/design-system/webpack/webpack.config.js
```diff
@@ -43,7 +44,15 @@ const webpackConfig = (env, argv = {}) => {
     ],
     optimization: {
       minimizer: [
-        new TerserPlugin(),
+        new TerserPlugin({
+          terserOptions: {
+            output: {
+              // do not preserve any kind of comments in the final bundle
```
arjunu, line 50: so it's on by default?

packages/design-system/src/components/buttons/Button.tsx
arjunu: we need a way to set `process.env.NF_CYPRESS_RUN` for `test:e2e:dev`

packages/web/webpack/webpack.config.prod.js
```diff
@@ -1,11 +1,20 @@
...
-const configCommon = require("./webpack.config.ci.prod");
+const configCommon = require("./webpack.config.common");
+const paths = require("./paths");
 
+/**
+ * @webpack a special webpack configuration that should used on when deploying production code
+ */
 module.exports = merge.smart(configCommon, {
+  devtool: false,
+  mode: "production",
```
arjunu, line 17: this already comes from `NODE_ENV=production webpack`

.circleci/config.yml
```diff
@@ -3,12 +3,12 @@ version: 2.1
 executors:
   docker:
     docker:
-      - image: circleci/node:11.14.0
+      - image: circleci/node:12.16.1
```
arjunu, line 6: They have 12.16.2 

`https://github.com/CircleCI-Public/circleci-dockerfiles/tree/master/node/images/12.16.2-stretch`

packages/web/webpack/webpack.config.common.js
```diff
@@ -24,9 +24,16 @@ module.exports = {
       template: paths.appHtml,
       favicon: paths.favicon,
     }),
-    new webpack.DefinePlugin({
-      "process.env": applicationEnv,
-    }),
+    // TODO: Find a way to warn if not defined env variable is used in the code
+    new webpack.DefinePlugin(
+      Object.entries(appEnv).reduce(
```
arjunu, line 29: didn't get what this is for

packages/web/webpack/webpack.config.prod.js
desfero: Not exactly true, https://webpack.js.org/configuration/mode/
Please remember that setting NODE_ENV doesn't automatically set mode.
The default is `production` but here I want to emphasize and make it consistent with `webpack.config.dev.js`

packages/web/webpack/webpack.config.common.js
desfero: done, now there is a comment

packages/design-system/webpack/webpack.config.js
desfero: nope :(

PR #4185, opened 2020-03-12
ETO.Pitch.Impact

arjunu:


packages/web/app/components/eto/registration/pages/EtoPitch/EtoRegistrationPitch.tsx
```diff
@@ -123,6 +123,13 @@ const EtoRegistrationPitchComponent = ({
         name="marketTraction"
       />
 
+      <FormTextArea
```
Moejoe90: so here is the place where we should have 
`process.env.NF_ISSUER_IMPACT_COMPANY_ENTRY === "1" `

packages/web/app/test-e2e/issuer/eto/fixtures.ts
```diff
@@ -348,11 +348,14 @@ export const productVisionForm: TFormFixture = {
     type: "rich-text",
     value: "Benefits for our investors",
   },
-
   marketTraction: {
     type: "rich-text",
     value: "Something about the market traction",
   },
+  impact: {
```
Moejoe90: This will break unless the backend merge their changes :)

packages/web/app/components/eto/eto-full-view/shared/EtoAccordionElements.tsx
```diff
@@ -179,6 +181,14 @@ const EtoAccordionElements: React.FunctionComponent<{
                 />
               ) : null}
 
+              {process.env.NF_ISSUER_IMPACT_COMPANY_ENTRY === "1" && impact ? (
```
Moejoe90: we don't need `process.env.NF_ISSUER_IMPACT_COMPANY_ENTRY === "1" &&` here as if it exists then its already ok to show

packages/web/app/components/eto/registration/pages/EtoPitch/EtoRegistrationPitch.tsx
arjunu: added

packages/web/app/components/eto/eto-full-view/shared/EtoAccordionElements.tsx
arjunu: removed

Issue #4197, opened 2020-03-16
Move JWT logic from web to shared modules


desfero:
**Background**
JWT token is required to implement notifications and tx history. To move it to shared-module we need to do some refactoring to extract implementation of the JWT storage and sync the `Web3Manager` and `EthManager` interfaces so we can just call a single method to sign JWT challenge.

**Designs**

**Acceptance criteria**
- [ ] AuthHttpClient is available for web app

**Technical notes**

**Testing notes**

PR #4211, opened 2020-03-23
#4197 move jwt logic to shared

desfero:
**Technical notes**
Additional changes:
- `selectAddressWIthChecksum` was removed and now `selectAddress` will always return checksummed address to reduce the ambiguity of wallet address format.
- a better module system for a web app where we properly declare module dependencies 
- in a couple of places more restrictive typing system

**Testing notes**
@p3


packages/shared-modules/src/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,170 @@
...
+import { coreModuleApi } from "../../core/module";
+import { AuthModuleError } from "../errors";
+import { ICreateJwtEndpointResponse } from "../lib/SignatureAuthApi";
+import { symbols } from "../lib/symbols";
+import { jwtActions } from "./actions";
+import { JwtNotAvailable } from "./errors";
+import { selectJwt } from "./selectors";
+import { EJwtPermissions } from "./types";
+
+type TGlobalDependencies = unknown;
```
desfero: this one need to be solved soon, for now i just created `neuGetBindings` but need to think of how to inject something into global dependencies and type it properly. If we don't find a good typesafe way then probably we should deprecate sagas global dependencies and just use `netGetBindings`. If you have an idea let's chat...

packages/shared-modules/src/modules/core/lib/client/IHttpClient.ts
```diff
@@ -6,8 +6,7 @@ export type HttpMethod = "GET" | "POST" | "PUT" | "PATCH" | "DELETE";
 export interface IHttpRequestCommon {
   url: string;
   baseUrl?: string;
-  // TODO: migrate YUPTS to shared package
-  responseSchema?: Yup.Schema<any> | any;
+  responseSchema?: Yup.Schema<any>;
```
desfero: Given that we want to get rid of yup ts wrappers `responseSchema` will only accept plain yup schema

packages/web/app/components/wallet-selector/WalletSelector.integration.spec.tsx
```diff
@@ -1,388 +0,0 @@
-import { createMock, createMount, remount, setupFakeClock, tid } from "@neufund/shared/tests";
```
desfero: Given that this test is skipped for a half of a year already we should remove it as fixing it is not a way to go IMO given the complexity of integration setup

packages/web/app/modules/access-wallet/sagas.ts
```diff
@@ -40,7 +40,7 @@ export function* ensureWalletConnection(
   }: TGlobalDependencies,
   password?: string,
 ): any {
-  const metadata = walletStorage.get();
+  const metadata = yield* call(() => walletStorage.get());
```
desfero: in general, i would prefer to use `yield* call([walletStorage, "get"])` as it should bee probably easier to mock but typings are not correct somewhere. Maybe in future will do a find and replace when we know the reason of broken typings

packages/web/app/modules/auth/jwt/sagas.ts
```diff
@@ -151,12 +18,12 @@ export function* refreshJWT({
 export function* ensurePermissionsArePresentAndRunEffect(
```
desfero: this should be also moved to shared but let's for now focus on core `jwt`

packages/web/app/store.ts
```diff
@@ -20,22 +34,39 @@ export type AppReducer<S> = Reducer<DeepReadonly<S>, AppActionTypes>;
...
+  const appModule = {
+    id: "app",
+    reducerMap,
+    sagas: [rootSaga],
+    libs: [setupBindings(config)],
+    middlewares: [routerMiddleware(history), reduxLogger(container)],
+  };
+
+  return [
+    setupCoreModule({ backendRootUrl: config.backendRoot.url }),
```
desfero: It's a proper way of declaring module dependencies. From the array, we can generate a full state of a given module state and actions

packages/shared-modules/src/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export class WalletNotAvailableError extends AuthModuleError {
```
arjunu: why is this not in errors file?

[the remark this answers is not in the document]
desfero: good catch

packages/shared-modules/src/modules/auth/jwt/sagas.ts
```diff
@@ -0,0 +1,172 @@
+import { call, neuCall, put, select } from "@neufund/sagas";
```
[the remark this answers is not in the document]
desfero: as discussed on slack, let's keep export per each function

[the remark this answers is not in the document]
desfero: `signChallange` was moved to `sagasInternal`

packages/shared-modules/src/modules/auth/README.md
```diff
@@ -0,0 +1,18 @@
...
+Exposes `authJsonHttpClient` and `authBinaryHttpClient` as symbols to make authorized http calls.
+
+Exposes utility sagas to handle lifetime of a JWT token (`loadJwt`, `createJwt`, `setJwt`,
+`escalateJwt`, `refreshJWT`)
+
+Allows to get access to the underline jwt token with `selectJwt` selector.
+
+### Tech notes
+
+Any kind of changes in the auth module requires intensive testing on moth web and mobile app side to
```
arjunu, line 17:
```suggestion
Any kind of changes in the auth module requires intensive testing on both web and mobile app side to
```

packages/web/app/modules/access-wallet/sagas.ts
Moejoe90: Btw you can do simply `call(walletStorage.get)` 
and that will work fine with our tests

PR #4222, opened 2020-03-30
Handle unsupported browsers

arjunu:
How it works:
- it's a separate script written in old syntax (see index.html) that runs independent of the app script
- uses it's own html & styles
- so even if app script crashes due to syntax error/unsupported API usage, it will not affect this

Changes:
- added cmd in pkg to build custom modernizr script from config
- the generated script has been committed to avoid rebuilding it every time as we don't expect it to change often. Let me know if you think we should build it with CI instead.
- the logic for checking browser support is in index.html
- all files for this are under `packages/web/app/external/unsupported-browser` dir


packages/web/.prettierignore
```diff
@@ -7,3 +7,4 @@ intl
 e2e
 .reg
 app/lib/contracts
+app/external/unsupported-browser/modernizr-custom.js
```
arjunu, line 10: because this is a generated file and it's already minified

packages/web/package.json
```diff
@@ -29,6 +29,7 @@
     "lint": "yarn format && yarn tslint && yarn sasslint && yarn tsc && yarn intl:extract:check",
     "lint:fix": "yarn contract-artifacts:update && concurrently -p \"[{name}]\" -n \"TSLINT/PRETTIER,SASS/TSC/INTL\" -c \"bgYellow.bold,bgMagenta.bold\" \"yarn tslint:fix && yarn format:fix\" \"yarn sasslint && yarn tsc && yarn intl:extract\"",
     "lint:fix:staged": "concurrently -p \"[{name}]\" -n \"TSC,LINT-STAGED\" -c \"bgYellow.bold,bgMagenta.bold\" \"yarn tsc\" \"lint-staged\"",
+    "modernizr": "modernizr -c modernizr-config.json -d app/external/unsupported-browser/modernizr-custom.js",
```
arjunu, line 32: builds script from config

packages/web/modernizr-config.json
```diff
@@ -0,0 +1,30 @@
+{
```
Moejoe90, line 1: can you also add intl?

packages/web/app/index.html
```diff
@@ -1,20 +1,129 @@
...
+          <img
+            src="/external/logo_neufund_on_white_title.svg"
+            alt="NEUFUND"
+            class="unsupported-browser__header-logo__title"
+          />
+        </a>
+      </header>
+
+      <div class="unsupported-browser__body">
+        <div>
```
desfero, line 37: seems we can omit `div`?

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
+.unsupported-browser {
+  display: none;
+  height: 100%;
+  min-width: 480px;
```
desfero, line 4: can we get rid of min-width (here an on line 27) and min-height on line 28 so we don't force scroll bars as the design should fit any kind of screen IMO

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
...
+.unsupported-browser__header-logo {
+  text-decoration: none;
+}
+
+.unsupported-browser__header-logo__title {
+  margin-left: 14px;
+}
+
+.unsupported-browser__body {
+  height: calc(100% - 128px);
```
desfero, line 22: i see you are using `display: flex` here so we can also get rid of `calc` here by applying flexbox system to header/content/footer, but it's up to you given flex may be not supported by the browser

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
...
+  margin-bottom: 60px;
+}
+
+.unsupported-browser__logo {
+  display: inline-block;
+  width: 20%;
+}
+
+.unsupported-browser__logo:last-child {
+  margin-right: 0;
```
desfero, line 81: not used?

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
...
+}
+
+.unsupported-browser__logo-image {
+  display: inline-block;
```
desfero, line 85: display inline seems to be not needed

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
...
+  margin-bottom: 12px;
+}
+
+.unsupported-browser__logo-title {
+  display: block;
```
desfero, line 90: we can just replace `span` by `p` and remove `display: block`

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
...
+  font-family: Roboto, sans-serif;
+  font-size: 14px;
+  color: #556877;
+  line-height: 24px;
+}
+
+.unsupported-browser__footer {
+  font-family: Roboto, sans-serif;
+  background-color: #142431;
+  height: 64px;
```
desfero, line 100: we have a padding therefore `height` can be ommited?

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
...
+  color: #94a5b3;
+  padding: 20px 32px;
+  display: flex;
+  align-items: start;
+}
+
+.unsupported-browser__link {
+  margin-left: 1rem;
```
desfero, line 108: what's the reason to use `rem` if everywhere here we use pixels?

packages/web/package.json
arjunu: added

packages/web/modernizr-config.json
arjunu: added

packages/web/app/index.html
```diff
@@ -1,20 +1,129 @@
...
+    <meta property="twitter:url" content="http://platform.neufund.org/" />
+    <meta name="twitter:card" content="Neufund Platform" />
+    <meta name="twitter:site" content="@neufundorg" />
+
+    <title>Neufund Platform</title>
+
+    <link rel="stylesheet" type="text/css" href="/external/unsupported-browser-styles.css" />
+  </head>
+  <body>
+    <div class="unsupported-browser" id="unsupported-browser">
```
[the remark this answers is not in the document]
arjunu: changed

packages/web/app/index.html
```diff
@@ -1,20 +1,129 @@
...
+        <h1 class="unsupported-browser__title">Update your browser</h1>
+
+        <p class="unsupported-browser__description">
+          It looks like you're using an outdated browser version. Please upgrade your browser, or
+          try using a different one from the list below.
+        </p>
+
+        <div class="unsupported-browser__logos">
+          <span class="unsupported-browser__logo">
+              <img
```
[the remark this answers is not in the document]
arjunu: We may have keep the links up to date and also the links could be different for mobiles

packages/web/app/index.html
```diff
@@ -1,20 +1,129 @@
...
+              <img class="unsupported-browser__logo-image" src="/external/logo-edge.svg"
+                   alt=Edge""/>
+              <span class="unsupported-browser__logo-title">Edge</span>
+            </span>
+        </div>
+
+        <div>
+          <span class="unsupported-browser__contact"
+            >Need help? Contact our
+            <a href="https://support.neufund.org/support/home">Customer Care</a>.</span
```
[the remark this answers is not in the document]
arjunu: added target blank too

packages/web/app/index.html
```diff
@@ -1,20 +1,129 @@
...
+    </div>
+
+    <div id="app"></div>
+
+    <script type="text/javascript" src="/external/modernizr-custom.js"></script>
+    <script>
+      function isBrowserSupported() {
+        for (var feature in Modernizr) {
+          var test = Modernizr[feature];
+          if (typeof test === "boolean" && test === false && feature !== "cssgridlegacy") {
```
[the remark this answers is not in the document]
arjunu: added

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
```diff
@@ -0,0 +1,157 @@
+.unsupported-browser {
+  display: none;
+  height: 100%;
+  min-width: 480px;
+}
+
+.unsupported-browser__header {
+  height: 64px;
```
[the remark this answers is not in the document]
arjunu: to me it's like an insurance policy, if any of the children overflow, the header remains the same height

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: oops left over, removed

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: removed

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: but `p` sounds semantically wrong in this case

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: changed to px

packages/web/app/index.html
arjunu: omitted the `span` inside

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: removed min-height and used flex, but I kept min-width and changed it to 320px (the smallest device width) which shouldn't give you scroll on any device except if you manually resize to a very small width. We have this behaviour for the app as well.

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: Changed to flex. Simple flex works even on IE. Will check again.

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: removed

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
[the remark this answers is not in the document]
arjunu: removed

packages/web/app/external/unsupported-browser/unsupported-browser-styles.css
arjunu: I had to put min height here for IE11

Issue #4205, opened 2020-03-17
[Mobile App MVP] Component - Nav Bars


Trilbi:
**Background**

**Designs**
Zeplin Link: https://app.zeplin.io/styleguide/5e5f852f2f231b17c87f8622/components?cseid=5e710d8c8bba5dbfca246977

**Acceptance criteria**

**Technical notes**

**Testing notes**

desfero:
Done as a part of #4192

PR #4192, opened 2020-03-15
4182 theming for mobile app

DmitriyNoa:

Issue #4182, opened 2020-03-12
Setup theming for mobile app


DmitriyNoa:
**Background**
Mobile app UI theming should be flexible and easy to change. Things like common (base) background and text colors, font sizes, boldness, etc. have to be customizable on a global level so we can support features like:
- Dark mode.
- Accessibility, the user selects an increased font size.

PR #4229, opened 2020-03-31
#4182 theming for mobile app

desfero:
A continuation of original #4192 

**Testing notes**
IOS: https://i.diawi.com/LKuZ8h
Android: https://i.diawi.com/wk7MgY
This pr:
- [ ] Adds tab navigation (https://github.com/Neufund/platform-frontend/issues/4205)
image
- [ ] Provides a proper app icon and app name for both android and ios


packages/wallet/app/themes/light/Light.ts
```diff
@@ -0,0 +1,21 @@
+import { EThemeVariant, ITheme } from "../types";
+import { lightColorSystem } from "./colorSystem";
+
+const navigationTheme = {
```
desfero: still need to connect theme with already created components, will do that soon in of the next pr with some component changes

packages/wallet/app/components/containers/AppContainer.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+
+import { ThemeProvider } from "../../themes/ThemeProvider";
+
+const AppContainer: React.FunctionComponent = ({ children }) => (
```
arjunu, line 5: is this component going to get any more logic in the future?

packages/wallet/app/AppRouter.tsx
```diff
@@ -1,25 +1,78 @@
...
-    </Stack.Navigator>
-  </NavigationContainer>
+type TBarProps = { color?: string };
+
+const Root: React.FunctionComponent = () => (
+  <Tab.Navigator initialRouteName={appRoutes.home}>
+    {tabConfig.map((route, index) => {
+      return (
+        <Tab.Screen
+          key={index}
```
arjunu, line 24: why not use name as key or add a new id property?

packages/wallet/app/components/shared/Icon.tsx
```diff
@@ -0,0 +1,44 @@
...
+import Wallet from "../../assets/wallet.svg";
+
+enum EIconType {
+  HOME = "home",
+  PORTFOLIO = "portf",
+  PROFILE = "profile",
+  WALLET = "wallet",
+}
+
+const getIcon = (type: EIconType) => {
```
arjunu, line 16: what if we have 30 icons? why not pass svg as children instead?

packages/wallet/app/components/containers/AppContainer.tsx
desfero: I would say so, for eg. we will include `AppearanceProvider`  in the future here.

packages/wallet/app/AppRouter.tsx
desfero: good point, fixed to use route as it should be unique always

packages/wallet/app/components/shared/Icon.tsx
desfero: discussed on slack

PR #4230, opened 2020-03-31
Fix whitelist latency

arjunu:
- fixed flashing of state (because it takes time to calculate `bookbuildingLimitReached`)
flash

- fixed incorrect public state 

this 
image

changed to
image


packages/web/app/modules/bookbuilding-flow/utils.ts
```diff
@@ -83,7 +83,7 @@ export const calculateWhitelistingState = ({
     isAuthorized &&
     isInvestor &&
     whitelistingIsActive &&
-    investmentCalculatedValues === undefined
+    (investmentCalculatedValues === undefined || bookbuildingLimitReached === undefined)
```
desfero, line 86: so `bookbuildingLimitReached` can be `undefined` when we don't have a any bookbuilding yet? Just trying to understand the logic here and why we can't do `&&` over `||`

arjunu: Here https://github.com/Neufund/platform-frontend/pull/4230 I'm setting `bookbuildingLimitReached` to `undefined` if `investorsCount` is undefined. It comes from a separate API call so it takes a sec to load.

We can't do  && because `investmentCalculatedValues` can be defined and `bookbuildingLimitReached` can be `undefined`

packages/web/app/modules/eto-view/notAuth/sagas.ts
```diff
@@ -50,6 +50,8 @@ export function* loadNotAuthorizedEtoView(
       payload.previewCode,
     );
 
+    yield* performLoadEtoSideEffects(eto);
```
arjunu, line 53: this line was missing and hence book building stats wasn't loading until reload ETO was called and therefore the latency

PR #4253, opened 2020-04-06
At/#4206 wallet connect subtype meta

AlexeyTuboltsev:


packages/web/app/lib/web3/PersonalWeb3.ts
```diff
@@ -2,14 +2,12 @@ import { EthereumAddressWithChecksum, EthereumNetworkId } from "@neufund/shared"
...
 import { TxData } from "web3";
 
-import { EWalletSubType, EWalletType, TWalletMetadata } from "../../modules/web3/types";
+import { TWalletMetadata } from "../../modules/web3/types";
 import { Web3Adapter } from "./Web3Adapter";
 
 export interface IPersonalWallet {
   readonly web3Adapter: Web3Adapter;
   readonly ethereumAddress: EthereumAddressWithChecksum;
-  readonly walletType: EWalletType;
```
Moejoe90, line 11: why remove this?!

This should stay as part of the `IPersonalWallet` interface

packages/web/app/modules/web3/types.ts
```diff
@@ -13,6 +15,7 @@ export enum EWalletSubType {
   METAMASK = "METAMASK",
   PARITY = "PARITY",
   GNOSIS = "GNOSIS",
+  NEUFUND = "NEUFUND",
```
Moejoe90, line 18: is it neufund really? 

I would say this should be `WALLET_CONNECT` 
and as a subtype you have `NEUFUND`

packages/web/app/lib/web3/wallet-connect/WalletConnectWallet.ts
```diff
@@ -31,21 +33,63 @@ export class WalletConnectSessionTransactionError extends WalletError {
...
+  supportsRemoteKyc: boolean;
+  supportsWalletMigration: boolean;
+};
+
+const generateWalletMetaFormPeerMeta = (peerMeta: TPeerMeta | null): TWcMeta => {
+  const walletMeta = {
+    walletSubType: EWalletSubType.UNKNOWN as
+      | EWalletSubType.UNKNOWN
+      | EWalletSubType.METAMASK
+      | EWalletSubType.NEUFUND,
```
Moejoe90, line 53: Why not create a type out of this your already used it in another place

packages/web/app/lib/web3/wallet-connect/WalletConnectWallet.ts
```diff
@@ -31,21 +33,63 @@ export class WalletConnectSessionTransactionError extends WalletError {
...
+  signerType: ESignerType;
+  sessionRequestTimeout: number;
+  signingTimeout: number;
+  supportsExplicitTimeouts: boolean;
+  supportSessionPings: boolean;
+  supportsRemoteKyc: boolean;
+  supportsWalletMigration: boolean;
+};
+
+const generateWalletMetaFormPeerMeta = (peerMeta: TPeerMeta | null): TWcMeta => {
```
Moejoe90, line 48: Looks like a good case for unit tests

packages/web/app/lib/web3/wallet-connect/WalletConnectWallet.ts
```diff
@@ -31,21 +33,63 @@ export class WalletConnectSessionTransactionError extends WalletError {
...
+    signerType: ESignerType.ETH_SIGN,
+    signingTimeout: WC_DEFAULT_SIGN_TIMEOUT,
+    sessionRequestTimeout: WC_DEFAULT_SESSION_REQUEST_TIMEOUT,
+    supportsExplicitTimeouts: false,
+    supportSessionPings: false,
+    supportsRemoteKyc: false,
+    supportsWalletMigration: false,
+  };
+
+  if (peerMeta !== null && peerMeta.name === "Metamask") {
```
Moejoe90, line 64: why are we using direct strings this should be an Enum from the the subtype of the wallet

packages/web/app/lib/web3/PersonalWeb3.ts
AlexeyTuboltsev: oh, don't know what I had on mind. That's definitely a mistake, thanks for catching it

packages/web/app/lib/web3/wallet-connect/WalletConnectWallet.ts
AlexeyTuboltsev: but that's where I assign the subtype of wallet. PeerMeta comes from wc peer and it's a bunch of strings. Or what do you mean?

packages/web/app/modules/web3/types.ts
AlexeyTuboltsev: that's exactly how it is. Did you forgot to expand the folded code? :D

packages/web/app/lib/web3/wallet-connect/WalletConnectWallet.ts
AlexeyTuboltsev: done

packages/web/app/lib/web3/wallet-connect/WalletConnectWallet.ts
AlexeyTuboltsev: good point. 
done

packages/web/app/modules/web3/types.ts
Moejoe90: ah my bad. it seemed as if we had it as a main type

Issue #3266, opened 2019-08-04
Move required modules for mobile MVP from web to shared-modules package


desfero:
**Background**
Given that mobile app MVP was defined we can try to move the parts of the shared logic from web app to our `shared-modules` package. It's not gonna be an easy task as web app modules are coupled with web app business state but fortunately, some basic work was already done as a part of #3918. 

- [x] `notificationCenter`
This one needs a refactor to move `notificationCenter` from lib to a module with actions `showInfo` and `showError` so then we can take actions on both `web` and `wallet` side and provide a proper platform dependent implementation. The reason for moving away from libraries is that libraries should no show anything on the UI, it's a responsibility of sagas to interact with UI.

- [x] `tx-history` module with `AnalyticsApi`.
It should be straightforward to move.
- [x] `contractsService`
This one is tricky to move as we need to provide different implementation base on the platform because `web` app is using `web3` where mobile app is using `etherjs`. `IEthManager` is an example of such shared library with different platform dependent providers

- [x] `tokenPrice` module
Should be easy to move, just find the way to abstract 
```yield take(actions.web3.newBlockArrived.getType());``` as its platform-specific action. (Maybe we can somehow extract block watchers, but that it's tricky one given we don't use `web3` in the mobile app)

- [x] `wallet` module.
Only the logic that's required to display nEur and Ether balance (Zeplin screen for reference zpl.io/29elgke). 

- [x] `investor portfolio` module
Only the logic that's required to display assets (Zeplin screen for reference zpl.io/aMG5Xvr). 

Besides the above, we probably need to move `eto` module but let's leave that until we truly need something from there as we already have a lot of to refactor.

**Technical notes**

**Testing notes**
It's a technical ticket to prepare code reuse between web and mobile so if web app works and tests passes everything should work the same way.

PR #4256, opened 2020-04-07
#3266 Move contracts service and token price to shared

### LINKED ISSUE #4232: Update MyWalletWidget unit tests to not use sagas
In https://github.com/Neufund/platform-frontend/pull/4147 ` packages/web/app/components/dashboard/my-wallet/MyWalletWidget.spec.tsx` tests were skipped because they are starting sagas in `createIntegrationTestsSetup`. Since unit tests are only supposed to test a single piece of code, this suite and any other must be updated.

desfero:
**Tech notes**
It's not possible to fully share `contractsService` given `web3` and `ethers` generated typings for contracts are not fully interchangeable. I found a way to use an adapter in the cases where we want to adopt some contract to be used in `shared-modules` (check how `contractsService.rateOracle` is used in `token-price` module)

Also closes #4232 

Also, component tests dependent on the store where refactored. It was a huge pain to move modules to shared because you always end up dirty fixing unit tests in components


packages/web/app/components/dashboard/my-wallet/MyWalletWidget.spec.tsx
```diff
@@ -1,61 +1,40 @@
...
-            neumarksDue: "2.01496624735285327461934e+23",
-            unlockDate: "1627528891",
-          },
-          neumarkAddress: "0x027a7a3991c4dd1dcb9db3f9a4dda8bab4d58f2f",
-          etherTokenBalance: "0",
-          euroTokenBalance: "0",
-          etherBalance: "5.8879038511742e+22",
-          neuBalance: "4.30223410549950281244511e+23",
-        },
+    const props = {
```
desfero, line 24: refactored to not use store but rather recieve props

packages/web/app/modules/init/sagas.ts
```diff
@@ -1,26 +1,33 @@
...
+}
+
 function* initSmartcontracts({ web3Manager, logger }: TGlobalDependencies): any {
   try {
     yield fork(neuCall, initWeb3ManagerEvents);
     yield web3Manager.initialize();
 
     yield neuCall(initializeContracts);
-    yield neuCall(populatePlatformTermsConstants);
+    yield neuCall(watchTokenPrices);
```
desfero, line 30: not sure if it should be here. Let me know what do you think

packages/wallet/app/store/create.tsx
```diff
@@ -20,9 +22,13 @@ import { TAppGlobalState } from "./types";
 import { setupStorageModule } from "../modules/storage";
 
 export const createAppStore = (container: Container) => {
+  // TODO: Take universe address from artifacts meta.json
```
desfero, line 25: Will do a ticket to find a way dynamically provide env variable (given we need to read file from disk and take the address from there)

packages/web/app/components/dashboard/my-portfolio/my-neu-widget/MyNeuWidget.spec.tsx
```diff
@@ -1,250 +1,92 @@
-import { createMock, createMount, setupFakeClock, tid } from "@neufund/shared/tests";
-import { BigNumber } from "bignumber.js";
+import { createMount, tid } from "@neufund/shared/tests";
```
desfero, line 1: Refactors the unit test to not be integration as it's impossible to maintain. Selectors are mocked so it's possible to change store shape without breaking unit tests for component

packages/web/app/components/dashboard/my-portfolio/my-neu-widget/MyNeuWidget.spec.tsx
```diff
@@ -1,250 +1,106 @@
...
+  let stubs: Record<string, SinonStub>;
+
+  beforeEach(() => {
+    // stub button link as it uses store under the hood
+    stub(buttonLink, "ButtonLink").callsFake((props: any) => (
+      <div data-test-id={props["data-test-id"]} />
+    ));
+
+    stubs = {
+      selectIsLoading: stub(walletSelectors, "selectIsLoading").returns(undefined),
```
desfero, line 25: let me know if you have an idea of how we can simplify mocking selectors?

packages/sagas/src/sagaEffects.ts
```diff
@@ -3,7 +3,7 @@ export * from "redux-saga";
 export { StringableActionCreator } from "@redux-saga/types";
 
 export * from "redux-saga/effects";
-export { call, race, SagaGenerator, select, getContext, take } from "typed-redux-saga";
+export { call, race, SagaGenerator, select, getContext, take, all } from "typed-redux-saga";
```
arjunu, line 6: why not do `export *`?

packages/wallet/scripts/generate-contracts.js
```diff
@@ -0,0 +1,56 @@
...
+  try {
+    await generateKnownInterfaces();
+  } catch (e) {
+    console.error("Failed to read meta.json and generate knownInterfaces.json");
+    console.error(e.message);
+    process.exit(1);
+  }
+}
+
+main().catch(console.error);
```
arjunu, line 56: what's this catch for? there's already catch inside main

packages/web/test/integrationTestUtils.unsafe.tsx
```diff
@@ -64,6 +65,11 @@ interface ICreateIntegrationTestsSetupOutput {
   sagaMiddleware: SagaMiddleware<{ container: Container; deps: TGlobalDependencies }>;
 }
 
+/**
+ * @deprecated Please don't write unit tests for components as integration tests.
```
arjunu, line 69: 👍

packages/web/app/components/dashboard/my-wallet/MyWalletWidget.stories.tsx
```diff
@@ -38,9 +38,9 @@ storiesOf("NDS|Molecules/Dashboard/MyWalletWidget", module)
     );
   })
   .add("default", () => (
-    <MyWalletWidgetComponentContainer {...props}>
+    <>
```
arjunu, line 41: is this required?

packages/sagas/src/sagaEffects.ts
desfero: to export * we need to refactor our app as the exports from `typed-redux-saga` (especially takeEvery and friends) do not yet in the same way as native `takeEvery`

packages/wallet/scripts/generate-contracts.js
desfero: good point. leftover

packages/web/app/components/dashboard/my-wallet/MyWalletWidget.stories.tsx
desfero: fixed, also on line `66`

Issue #4234, opened 2020-04-01
walletConnect: logout, disconnect, login flows


AlexeyTuboltsev:
login, disconnect and logout:
- [x] need a special flag "logout in progress" to not restart logout process on 'disconnect' event if it's already under way.
- [x] need to investigate why wc doesn't restart if logged out from the mobile side

PR #4260, opened 2020-04-07
#4234 wc logout login

AlexeyTuboltsev:
this PR adds more stages to login flow. See ticket for details


packages/web/app/components/wallet-selector/WalletSelectorLayout.tsx
```diff
@@ -51,6 +56,34 @@ const SelectTitle: React.FunctionComponent<ISelectTitleProps> = ({
   }
 };
 
+const LogoutReason: React.FunctionComponent<ILogoutReasonProps> = ({
```
Moejoe90, line 59: I'm thinking if we make this component connected but that's not for now

Issue #4226, opened 2020-03-31
Returning user "log in" flow


andywalton:
**Background**
The first time a users uses our mobile app and creates an account is handled by #4219.

The context for this ticket is for returning users who need to unlock their account, ie. "log in".

**Please note**, when we have security enabled for the app via device FaceID/TouchID or Pass/Pincode, we can simply initiate a presence check on the app loading screen and use Welcome back, screen as a fallback should you fail that. For now, we'll use this Welcome back screen primarily as a quick way to toggle fixtures by making switching accounts easy to navigate to.

**Designs**
Head over to Zeplin and look in the section: Login / Unlock
https://app.zeplin.io/project/5e6f48a43076c61012fd3793/dashboard?seid=5e6f49ad3076c61012fd42b6

**Acceptance criteria**
- [x] As a returning user I can tap the `Unlock account` button on the _Welcome back_ view. Doing so will load the app as for the MVP we aren't using device security features yet.
- [x] As a returning user, `Unlock account` will load the app and land me on the _Home_ view.
- [x] On the _Welcome back_ view -> it should highlight which "account" is active. Ie. the Welcome back, <username> ideally should display the users account username associated to the wallet. Technically we'll need to check what we can use for this as we might need to add a username field to our wallet fixtures for this and allow you to add this information via the Profile view.
- [x] As a returning user I can tap the `Switch account` button. Doing so will open the _Switch account_ view. Here a user will be able to switch between the various accounts connected to the users device. This is important for testing.
- [ ] ~~On the _Switch account_ view as a user I should be able to tap on the `Connect another account` button. This will change the views to the "Welcome screen" that allows me to create a new account~~ (moved to #4217).

**Technical notes**
@desfero -> please add technical notes here.

- Needs clarity on the <username> field above.

**Testing notes**
For convenience `Switch account` is also included on `Create account` as `Import fixture`.  `Connect another account` was renamed to `Connect account` given we reuse the same screen for both flows (new user flow and unlock flow)

PR #4267, opened 2020-04-09
Make `invariant` fully type safe

desfero:
**Technical notes**
Extracted from #4226 to keep PR clean from web app changes.


packages/shared/src/utils/invariant.ts
```diff
@@ -1,12 +1,17 @@
...
-  type = "InvariantError";
+  constructor(message: string) {
+    super(`InvariantError: ${message}`);
+  }
 }
 
-export function invariant(value: any, message: string): null {
-  if (!value) {
+/**
+ * Assets that the given condition is truthy
```
desfero, line 8:
```suggestion
 * Asserts that the given condition is truthy
```

packages/shared/src/utils/invariant.ts
```diff
@@ -1,12 +1,17 @@
...
+ *
+ * @param condition - Either truthy or falsy value
+ * @param message - An error message is required so it's easier to figure out the error
```
desfero, line 11:
```suggestion
 * @param message - An error message
```

packages/shared/src/utils/invariant.ts
```diff
@@ -1,12 +1,17 @@
...
+ */
+export function invariant(condition: any, message: string): asserts condition {
+  if (!condition) {
     throw new InvariantError(message);
   }
-
-  // Useful in react component context as `undefined` is not allowed in JSX
-  return null;
```
desfero, line 11: not possible to have return type other than `never | void`

packages/shared/src/utils/invariant.spec.ts
```diff
@@ -13,4 +14,12 @@ describe("invariant", () => {
     invariant(true, "description");
     invariant(1, "description");
   });
+
+  it("should properly narrow type", () => {
```
arjunu, line 18: why not put this in `./test/testUtils.spec.ts` ?

PR #4245, opened 2020-04-03
#4231 add import account flow

desfero:
Base on #4220 so please review it before

**Demo**
ezgif com-video-to-gif

**TODO:**

- [x] Add button icon component
- [x] Fix android text area alignment
- [x] Align wording with @andywalton 

**Testing notes**
Fixes also #4225.

Android: https://i.diawi.com/rgy8um
IOS: https://i.diawi.com/ccPytm

PR #4271, opened 2020-04-13
#4226 returning user flow

desfero:
**Technical notes**
Also:
- wallet app typescript/babel/prettier updated to have support for `import type/export type` given that `isolateModules` compiler flag should be always turned. 
- `init` module was fully migrated to the new module format.

Should be reviewed after #4245 

**Testing notes**
Besides, unlock and switch screen also 
- switch account to `create account flow` was added
- basic view of profile was implemented to show address and switch account menu item

IOS: https://i.diawi.com/eft9KK
Android: https://i.diawi.com/KaA4fj


packages/wallet/app/AppNoAuthRouter.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <NoAuthStack.Navigator
+      initialRouteName={hasAnAccountToUnlock ? appRoutes.unlockAccount : appRoutes.createAccount}
+      screenOptions={({ route, navigation }) => ({
+        gestureEnabled: true,
+        cardOverlayEnabled: true,
+        headerStatusBarHeight:
+          navigation.dangerouslyGetState().routes.indexOf(route) > 0 ? 0 : undefined,
+        ...TransitionPresets.ModalPresentationIOS,
```
arjunu, line 30: shouldn't this be the first line so that it won't override the properties above it?

packages/wallet/app/components/create-account/CreateAccountScreen.tsx
```diff
@@ -92,18 +103,18 @@ const styles = StyleSheet.create({
     textAlign: "center",
```
arjunu, line 103: mb is spacing 5 in the design

packages/wallet/app/components/create-account/CreateAccountScreen.tsx
```diff
@@ -1,13 +1,14 @@
 import { useNavigation } from "@react-navigation/native";
```
arjunu, line 1: can you put all the route components in a separate directory?
Either `components/routes` or `components/screens`?

arjunu, line 1: CreateAccountScreen is not an apt name, maybe WelcomeScreen?

packages/wallet/app/AppNoAuthRouter.tsx
desfero: fixed

packages/wallet/app/components/create-account/CreateAccountScreen.tsx
desfero: good catch

packages/wallet/app/components/create-account/CreateAccountScreen.tsx
desfero: let's leave it for now as it will cause conflicts.

desfero: Renamed to `LandingScreen`

PR #4278, opened 2020-04-16
#4228_kyc_form_save_and_close_fix

AlexeyTuboltsev:
fix form not saving because of invalid birth date


packages/web/app/components/eto/utils.ts
```diff
@@ -202,6 +203,16 @@ export const removeKeys = () => (data: { key: string }[]) =>
     return arrayElement;
   });
 
+// our date field returns "--" for empty date
+// we need to remove it in order to be able to save and close forms without validating
+export const removeInvalidDate = () => (data: string | undefined) => {
```
arjunu, line 208: why not keep this in kyc utils?

PR #4279, opened 2020-04-16
Move notification module to shared-modules

arjunu:
- add `showInfo` & `showError` actions to `shared-modules` pkg
- removed `notificationCenter` global dependency
- replaced calls to `notificationCenter` with actions
- created `createNotificationMessage` function to map `TranslatedMessageType` to message payload for the actions


packages/web/app/modules/notification-ui/module.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const MODULE_ID = "web:notification-ui";
+
+const setupWebNotificationUIModule = () => ({
+  id: MODULE_ID,
+  sagas: [webNotificationUISaga],
+  api: webNotificationUIModuleApi,
+});
+
+const webNotificationUIModuleApi = {
```
arjunu, line 13: called it `webNotificationUIModuleApi` to distinguish it from the shared-modules notification module

packages/shared-modules/src/modules/notification-ui/actions.ts
```diff
@@ -0,0 +1,11 @@
+import { createActionFactory } from "@neufund/shared-utils";
+
+const payloadCreator = (message: any, options?: any) => ({
```
desfero, line 3: just wondering, what if we use `unknown` here so it's properly cast by the developer where it's used?

packages/shared-modules/src/modules/notification-ui/module.ts
```diff
@@ -0,0 +1,3 @@
+import { notificationUIActions } from "./actions";
+
+export { notificationUIActions };
```
desfero, line 3: for consistency, we should always use the same pattern
```
const notificationUIAPI = {
actions: notificatioinUIActions,
}
```

packages/web/app/modules/notification-ui/sagas.ts
```diff
@@ -0,0 +1,27 @@
...
+import { notificationUIActions } from "@neufund/shared-modules";
+
+import { showErrorToast, showInfoToast } from "../../components/shared/Toast";
+import { getMessageTranslation } from "../../components/translatedMessages/messages";
+import { TGlobalDependencies } from "../../di/setupBindings";
+
+function* showInfo(
+  _: TGlobalDependencies,
+  action: TActionFromCreator<typeof notificationUIActions, typeof notificationUIActions.showInfo>,
+): Generator<unknown, void> {
```
desfero, line 11: let's use `SagaGenerator` type as it's more type-safe one, same with `showError`

packages/shared-modules/src/modules/notification-ui/actions.ts
huesforalice: done

packages/shared-modules/src/modules/notification-ui/module.ts
huesforalice: done

packages/web/app/modules/notification-ui/sagas.ts
huesforalice: done

packages/wallet/app/modules/notification-ui/sagas.ts
```diff
@@ -1,22 +1,20 @@
...
 function* showInfo(
   _: TGlobalDependencies,
   action:
-    | TActionFromCreator<typeof notificationUIActions, typeof notificationUIActions.showInfo>
-    | TActionFromCreator<typeof notificationUIActions, typeof notificationUIActions.showError>,
+    | TActionFromCreator<typeof actions, typeof actions.showInfo>
+    | TActionFromCreator<typeof actions, typeof actions.showError>,
 ): Generator<unknown, void> {
-  Alert.alert(action.payload.message);
+  Alert.alert(action.payload.message as string);
```
desfero, line 15: Let's use `invariant` (from `shared-utils`) here so we don't need to force cast
```
invariant(isString(action.payload.message), "Invalid notification message type recieved");

Alert.alert(action.payload.message);
```
You can either use `isString` from lodash `import isString from "lodash/fp/isString";` or a manual `typeof`

packages/web/app/modules/notification-ui/sagas.ts
```diff
@@ -1,27 +1,30 @@
...
+
 function* showInfo(
   _: TGlobalDependencies,
-  action: TActionFromCreator<typeof notificationUIActions, typeof notificationUIActions.showInfo>,
-): Generator<unknown, void> {
+  action: TActionFromCreator<typeof actions, typeof actions.showInfo>,
+): SagaGenerator<void> {
   const { message, options } = action.payload;
-  showInfoToast(getMessageTranslation(message), options);
+  showInfoToast(getMessageTranslation(message as TMessage), options);
```
desfero, line 16: same as with `notification-ui` in wallet (also line 24). Here it's a little bit tricky as it's an object. You can either create a yup schema and validate (hope it will narrow the type) or use a type-predicates https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

PR #4283, opened 2020-04-21
Add all packages to lint step in CI 

arjunu:


.circleci/config.yml
```diff
@@ -28,12 +28,14 @@ jobs:
       # Danger
       - run: yarn workspace @neufund/web danger ci
       # Linting
-      - run: yarn workspace @neufund/web lint
-      - run: yarn workspace @neufund/shared-utils lint
       - run: yarn workspace @neufund/design-system lint
+      - run: yarn workspace @neufund/sagas lint
```
desfero, line 32: hm, would bee good if we move install and lints outside of `web_instal_lint_tests` so each package has a separate job. Then we can move them to workflows as separate jobs to run. In future we can implement diffing to only run jobs related to the changed files

arjunu: postponed for now since having different jobs will increase CI time

Issue #3930, opened 2020-01-20
Migrate to ESLint


arjunu:
[TSLint is deprecated](https://github.com/palantir/tslint/issues/4534) and we must move to [ESLint](https://github.com/typescript-eslint/typescript-eslint)

PR #4284, opened 2020-04-21
Add esling config

desfero:
Partially implements changed required for #3930 by introducing a way to share configs between packages and with basic eslint config that we can use.

**Technical notes**
- New package `eslint-config` added to store and share config between apps/packages
- I haven't reviewed react rules (only react-native) so that's something we will have to do as a part of #3930 
- Mobile app wallet connect implementation was fixed thanks to the bugs reported by eslint


packages/wallet/app/components/home/Header.tsx
```diff
@@ -1,15 +1,14 @@
 import { useNavigation } from "@react-navigation/native";
 import React from "react";
-import SafeAreaView from "react-native-safe-area-view";
 import { StyleSheet, View } from "react-native";
+import SafeAreaView from "react-native-safe-area-view";
 
 import { appRoutes } from "../../appRoutes";
+import NeufundLogo from "../../assets/neufund-logo.svg";
```
arjunu, line 7: can we keep asset imports in a separate category?

packages/wallet/app/components/shared/buttons/ButtonIcon.tsx
```diff
@@ -1,7 +1,9 @@
+/* eslint-disable react-native-a11y/has-valid-accessibility-traits, react-native-a11y/has-valid-accessibility-states */
```
arjunu, line 1: these can't be set?

packages/wallet/app/components/shared/forms/layouts/switcher/SwitcherItem.tsx
```diff
@@ -43,9 +41,10 @@ const SwitcherItem: React.FunctionComponent<TItemExternalProps> = ({
   <TouchableOpacity
     style={st(styles.container, [position === ESwitcherItemPosition.LAST, styles.containerLast])}
     activeOpacity={0.4}
-    accessibilityComponentType={isSelected ? "radiobutton_checked" : "radiobutton_unchecked"}
-    accessibilityTraits={isSelected ? ["button", "selected"] : "button"}
```
arjunu, line 46: `accessibilityTraits` will be soon deprecated

packages/wallet/app/components/shared/buttons/ButtonIcon.tsx
arjunu: removed

PR #4291, opened 2020-04-22
keychain pt1

huesforalice:
This PR implements storing secrets in the "SecureStorage" class into the Keychain  on iOS and Android. Features:
* Keychainstorage with Asyncstorage fallback on emulator and simulator
* 10 Seconds key caching time to prevent multiple biometrics popups within a short period of time.


packages/wallet/app/modules/device-information/DeviceInformation.ts
```diff
@@ -0,0 +1,54 @@
...
+import { ILogger } from "@neufund/shared-modules";
+
+/**
+ * @class DeviceInformation
+ * Layer on top of react-native-device-info
+ * Provides access to a device information, thought the API. Available methods are here https://github.com/react-native-community/react-native-device-info#api.
+ *
+ */
+@injectable()
+export class DeviceInformation {
```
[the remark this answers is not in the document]
huesforalice: It's copied from there bc that PR is not merged yet and I needed it here.

huesforalice:
I turns out that the android "face id" still has bugs. I disabled biometry support for android for now until this bug has been resolved. There already is an issue in the react native keychain project.

packages/wallet/android/app/src/main/java/com/wallet/MainApplication.java
```diff
@@ -25,6 +26,7 @@ public boolean getUseDeveloperSupport() {
           List<ReactPackage> packages = new PackageList(this).getPackages();
           // Packages that cannot be autolinked yet can be added manually here, for example:
           // packages.add(new MyReactNativePackage());
+          // packages.add(new KeychainPackage());
```
desfero, line 29: do we need to add the package? I see added in the getting started for react-native-keychain

desfero: hm, probably we do not need it as library supports autolinking

packages/wallet/app/modules/eth/lib/EthSecureEnclave.ts
```diff
@@ -48,31 +51,63 @@ class FailedToDerivePrivateKey extends EthSecureEnclaveError {
...
    * @note This operation is unsafe as it exposes secret to the device memory
    *
    * @param reference - A reference to the secret
    *
    * @returns A secret value saved under the provided reference
    */
-  unsafeGetSecret(reference: TSecureReference): Promise<string | null> {
-    return this.secureStorage.getSecret(reference);
+  async unsafeGetSecret(reference: TSecureReference): Promise<string | null> {
+    this.logger.info(`Retrieving secret for ref ${reference.substring(0, 4)}`);
```
desfero, line 98: +1

packages/wallet/app/modules/eth/lib/EthSecureEnclave.ts
```diff
@@ -48,31 +51,63 @@ class FailedToDerivePrivateKey extends EthSecureEnclaveError {
...
+    const storage = await this.getStorage();
+    return await storage.getSecret(reference);
```
desfero, line 100: no need for `await`

packages/wallet/app/modules/eth/lib/SecureStorage.ts
```diff
@@ -17,36 +20,110 @@ const toSecureReference = (reference: string) => reference as TSecureReference;
...
+    this.useBiometry = useBiometry;
+    this.localCache = new Cache();
+  }
+
   async setSecret(secret: string): Promise<TSecureReference> {
     const reference = utils.bigNumberify(utils.randomBytes(32)).toString();
 
-    await AsyncStorage.setItem(reference, secret);
+    // dev implementation
+    if (this.useAsyncStorageFallback) {
```
desfero, line 45: let's split the secure implementation (on top of keychain) with the test one based on async storage so it's easier to maintain. Then we can either use `inversify` factory to decide which one to return or  just manually in `EthSecureEnclave`.

packages/wallet/app/modules/eth/lib/SecureStorage.ts
```diff
@@ -1,10 +1,13 @@
 import { Opaque } from "@neufund/shared-utils";
 import { utils } from "ethers";
+import { Cache, CacheClass } from "memory-cache";
```
desfero, line 3: Do we really need to use this library? Seems like overkill here + it's not maintained (last release 3 years ago). We can try to just use plain https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

packages/wallet/app/modules/eth/lib/__mocks__/react-native-keychain.ts
```diff
@@ -0,0 +1,43 @@
+import { Dictionary } from "@neufund/shared-utils";
```
desfero, line 1: is there a reason to just not use https://github.com/oblador/react-native-keychain#using-a-jest-__mocks__-directory?

packages/wallet/app/modules/eth/lib/SecureStorage.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { Cache, CacheClass } from "memory-cache";
+
+function sleep(ms: number) {
+  return new Promise(resolve => setTimeout(resolve, ms));
```
desfero, line 4: i would say we should just mock timers https://jestjs.io/docs/en/timer-mocks

packages/wallet/app/modules/eth/lib/SecureStorage.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Write some  basic tests to ensure the used caching lib works the way we expect
+ */
+describe("SecureStorage", () => {
```
desfero, line 10: it's really a test for external library `memory-cache`. What i think we should do is to write a proper test here for secure storage with mocks for `react-native-keychain`. And then we can simplify EthSecureEnclave.spec.ts tests to not be aware of `react-native-keychain` because it's a transitive dependency of `EthSecureEnclave`.

packages/wallet/app/modules/eth/lib/SecureStorage.ts
desfero, line 3: the library seems to be not maintained anymore (last release 3 years ago + handing PR with not response from the author). But for now, we can leave with it and hope it won't cause problems and manual patches

packages/wallet/package.json
```diff
@@ -46,26 +46,30 @@
...
     "@neufund/sagas": "^1.0.0",
-    "@neufund/shared-utils": "^1.0.0",
     "@neufund/shared-modules": "^1.0.0",
+    "@neufund/shared-utils": "^1.0.0",
     "@react-native-community/async-storage": "^1.8.0",
     "@react-native-community/masked-view": "^0.1.6",
     "@react-navigation/bottom-tabs": "^5.0.1",
     "@react-navigation/native": "^5.0.1",
     "@react-navigation/stack": "^5.0.1",
+    "@types/memory-cache": "^0.2.1",
```
desfero, line 56: @types/memory-cache should be in `devDependencies`

packages/wallet/tsconfig.json
```diff
@@ -22,7 +22,8 @@
     "resolveJsonModule": true,
     "strict": true,
     "target": "esnext",
-    "types": ["react-native"]
+    "types": ["react-native"],
+    "skipLibCheck": true
```
desfero, line 26: what's the reason to skip lib checks? We lose some typesafety heree

packages/wallet/android/app/src/main/java/com/wallet/MainApplication.java
huesforalice: no, it works without!

packages/wallet/app/modules/eth/lib/EthSecureEnclave.ts
huesforalice: ah yeah, true

packages/wallet/app/modules/eth/lib/__mocks__/react-native-keychain.ts
huesforalice: ah, interesting, I did not know of this possibility.

packages/wallet/package.json
huesforalice: done

packages/wallet/app/modules/eth/lib/SecureStorage.ts
```diff
@@ -17,36 +20,106 @@ const toSecureReference = (reference: string) => reference as TSecureReference;
...
-    await AsyncStorage.setItem(reference, secret);
+    // dev implementation
+    if (this.useAsyncStorageFallback) {
+      await AsyncStorage.setItem(reference, secret);
+    }
 
+    // keychain implementation
+    else {
+      await Keychain.setInternetCredentials(reference, "", secret, {
+        accessControl: Keychain.ACCESS_CONTROL.USER_PRESENCE,
```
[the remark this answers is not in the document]
huesforalice: yes, it's working :)

packages/wallet/tsconfig.json
huesforalice: Unfortunately there is a missing interface in react native keychain:
https://github.com/oblador/react-native-keychain/pull/333

It's already fixed but not released yet. Is there maybe a way to only skip react native keychain lib check?

packages/wallet/app/modules/eth/lib/__mocks__/react-native-keychain.ts
huesforalice: Which part exactly do you want me to use from this? It's already a jest mock with a mock implementation of the actual storage. I think it's pretty good for our purposes, no?

packages/wallet/app/modules/eth/lib/SecureStorage.ts
huesforalice: Probably there is not so much that can change in a very simple lib like this. I've added a few simple tests to ensure it behaves the way we expect. Ofc we could also write our own, but I think that probably would be a waste of time.

packages/wallet/app/modules/eth/lib/SecureStorage.spec.ts
huesforalice: Ah, nice, thanks. Done!

packages/wallet/app/modules/eth/lib/SecureStorage.ts
huesforalice: Done

packages/wallet/app/modules/eth/lib/SecureStorage.spec.ts
huesforalice: You're right, I've added a bunch of tests for securestorage. I did not change the secure enclave tests, as I don't really think we need to. The Securestorage is an internal dependency. But maybe I don't completely understand what you're looking for.

PR #4295, opened 2020-04-23
add a preinstall script to update submodules

AlexeyTuboltsev:


package.json
```diff
@@ -12,6 +12,7 @@
     "update-submodules": "git submodule update --init --recursive",
     "prepare": "yarn update-submodules && wsrun --stages prepare",
     "deduplicate": "yarn-deduplicate",
+    "preinstall": "yarn prepare",
```
arjunu, line 15: keep `prepare` as it is. Move just `yarn update-submodules` to `preinstall`

.circleci/config.yml
```diff
@@ -55,7 +55,6 @@ jobs:
 
     steps:
       - checkout
-      - run: yarn update-submodules
```
arjunu, line 58: I think these are still required

Issue #4217, opened 2020-03-25
Add session storage to wallet-connect module


desfero:
**Background**
In #4178 we did a basic implementation for wallet connect that will allow building more complex WC handling in the future. Saving wallet connects in the session storage and reconnecting after opening from the background in the next important milestone pushing wallet-connect integration forward.

**Acceptance criteria**
- [x] App must keep the session key and data in-app storage to be able to re-connect when killed and coming from a background
- [x] when coming from the background, web socket must be reconnected and pending messages must be read.
- [x] On the Profile view as a user I should be able to tap on the Log out button. This will change the views to the "Welcome screen" that allows me to create a new account.

**Technical notes**

**Testing notes**

PR #4299, opened 2020-04-27
#4217 add session storage for wallet connect

desfero:
**TODO:**
- [x] Check if there is a memory leak after the session was disconnected
- [x] Test whether sign message is working after @rudolfix fixes web app
- [ ] Clean up session when the new user was signed


packages/wallet/app/components/shared/Link.tsx
```diff
@@ -2,74 +2,49 @@
...
-  const onShowUnderlay = () => {
-    setIsActive(true);
-  };
-
-  const onHideUnderlay = () => {
-    setIsActive(false);
-  };
-
-  return (
-    <TouchableHighlight
```
desfero, line 44: I wasn't aware that it's not possible to nest `TouchableHighligh` into the `Text` in react-native. But that's good as we can simplify the implementation to only have a `Text` node here

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -31,6 +33,11 @@ function* initStartSaga(): SagaGenerator<void> {
   }
 }
 
+function* initAuthSigned(): SagaGenerator<void> {
+  yield* call(walletConnectModuleApi.sagas.tryToConnectExistingSession);
+}
+
 export function* initSaga(): SagaGenerator<void> {
   yield fork(neuTakeLatest, initActions.start, initStartSaga);
+  yield fork(neuTakeLatest, authActions.signed, initAuthSigned);
```
desfero, line 42: not sure if it's the best approach here as now init module watches for `authActions.signed`. In general, it looks good to me as we avoid coupling `auth` module with `wallet-connect` module given that they are not aware of each other and it's part of `init` module to connect them properly

packages/wallet/app/modules/wallet-connect/lib/WalletConnectAdapter.ts
```diff
@@ -0,0 +1,300 @@
+import { ILogger } from "@neufund/shared-modules";
```
desfero, line 1: almost a one to one copy from `WalletConnectManager` given the manager was repurposed to control session storage and instantiation of WalletConnectAdapter

packages/wallet/app/modules/wallet-connect/lib/WalletConnectAdapter.ts
```diff
@@ -0,0 +1,300 @@
...
+ * Wraps wallet connect class to simplify interface and to provide type safety
+ *
+ * @internal WalletConnectAdapter should only be used in WalletConnectManager and never exposed to DI container
+ */
+class WalletConnectAdapter extends EventEmitter2 {
+  private readonly logger: ILogger;
+
+  private readonly walletConnect: WalletConnect;
+
+  private connectedAt: number | undefined = undefined;
```
desfero, line 72: `connectedAt` was added to the adapter in this PR

arjunu:
Closed in favor of https://github.com/Neufund/platform-frontend/pull/4362

PR #4300, opened 2020-04-28
Fix invest all with empty eth wallet

AlexeyTuboltsev:
add guards to avoid 0 eth as investment input

### DIFF
--- a/packages/web/app/modules/investment-flow/sagas.ts
+++ b/packages/web/app/modules/investment-flow/sagas.ts
@@ -135,8 +135,13 @@ function* investEntireBalance(): any {
     case EInvestmentType.Eth:
       const gasCostEth = selectTxGasCostEthUlps(state);
       balance = selectLiquidEtherBalance(state);
-      balance = subtractBigNumbers([balance, gasCostEth]);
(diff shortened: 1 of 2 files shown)


packages/web/app/modules/investment-flow/sagas.ts
```diff
@@ -209,7 +214,11 @@ function* validateAndCalculateInputs({ contractsService }: TGlobalDependencies):
   const eto = selectEtoById(state, state.investmentFlow.etoId);
   const value = state.investmentFlow.euroValueUlps;
 
-  if (value && eto) {
+  if (compareBigNumbers(value, "0") < 0) {
```
Moejoe90, line 217: why do we compare with the assumetion that value could be also in minus this is not the case the minum value is `0` 

or what am I missing?

AlexeyTuboltsev: the problem was exactly that this value was negative and `loadComputedCotributionFromContract` that performs some calculations in contract just froze each times we try to do this with a negative val.
so I added this as guard, just to make sure value is positive | 0

PR #4310, opened 2020-04-30
Onboarding logic fixes

Moejoe90:
Note that the tests are currently skipped for `loadOrCreateUser` due to the new dependancy injection. I didn't have enough time to see what is exactly wrong there.


arjunu:
> Note that the tests are currently skipped for `loadOrCreateUser` due to the new dependancy injection. I didn't have enough time to see what is exactly wrong there.
Fixed and unskipped

packages/web/app/modules/auth/user/sagas.spec.ts
```diff
@@ -15,9 +15,10 @@ import { loadKycRequestData } from "../../kyc/sagas";
 import { EWalletSubType, EWalletType } from "../../web3/types";
 import { loadOrCreateUser } from "./sagas";
 
-describe.skip("Auth - User - Integration Test", () => {
+describe("Auth - User - Integration Test", () => {
   describe("loadOrCreateUser", async function(): Promise<void> {
-    this.timeout(5000);
+    // TODO timeout bad, replace with mocked services
```
Moejoe90, line 20: We need to be very carful IMO what to mock and what not to mock. 

In this case I wouldn't mock the values of the backend because we need to make sure the saga handels the correct values. The values returned at some point were not inline to the swagger definitions. 

Mocking is better most of the time but not when we can't guarentee the integrity of the data. This doesn't mean all tests need to touch the real backend but at least one of two checks should be done to avoid false postives

PR #4321, opened 2020-05-04
move tx history module to shared

huesforalice:


packages/shared-modules/src/modules/tx-history/sagas.ts
```diff
@@ -0,0 +1,470 @@
...
+export function setupTXHistorySagas(_: TSetupSagasConfig): () => SagaGenerator<void> {
+  return function* txHistorySaga(): SagaGenerator<any, any> {
+    yield fork(neuTakeLatest, txHistoryActions.loadTransactions, loadTransactionsHistory);
+    yield fork(neuTakeLatest, txHistoryActions.loadNextTransactions, loadTransactionsHistoryNext);
+    yield fork(
+      neuTakeUntil,
+      txHistoryActions.startWatchingForNewTransactions,
+      txHistoryActions.stopWatchingForNewTransactions,
+      watchTransactions,
+      // config.refreshOnAction, // TODO: this should be commented in, but types don't work
```
huesforalice, line 467: I somehow can't get this to work with the typing. I updated neuTakeUntil but it won't accept it. A few lines above are a few lines of related commented out code.

packages/web/app/modules/tx-history/module.ts
```diff
@@ -0,0 +1,25 @@
...
+const MODULE_ID = "web:tx-history";
+
+const setupWebTxHistoryModule = (config: TModuleConfig) => {
+  const parentModules = setupTxHistoryModule(config);
+
+  const webTxHistoryModule = {
+    id: MODULE_ID,
+    api: txHistoryApi,
+    sagas: [setupTXHistorySagas()],
+    reducerMap: parentModules[1].reducerMap,
```
huesforalice, line 19: I had to re-export the reducer map like this for the typing on the TGlobalState to have tx-history. I'm not sure this is supposed to be this way, but I did not find any other way..

packages/shared-modules/src/modules/tx-history/constants.ts
```diff
@@ -0,0 +1,4 @@
+import { minutesToMs } from "@neufund/shared-utils";
+
+export const TX_REFRESH_DELAY = minutesToMs(1);
```
huesforalice, line 3: I hope this is a sensible default :)

packages/shared-modules/src/modules/tx-history/module.ts
```diff
@@ -0,0 +1,54 @@
...
+const setupTxHistoryModule = (config: Config) => {
+  const module = {
+    id: MODULE_ID,
+    api: txHistoryApi,
+    libs: [setupContainerModule()],
+    sagas: [setupTXHistorySagas(config)],
+    reducerMap: txHistoryReducerMap,
+  };
+
+  return [setupTokenPriceModule({ refreshOnAction: undefined }), module];
```
huesforalice, line 44: This seems to work in the main project without problems of having the same module there twice.

packages/web/app/modules/actions.ts
```diff
@@ -44,14 +44,14 @@ import { web3Actions } from "./web3/actions";
 /** You should add new actions also here (with a namespace).*/
 export const actions = {
   ...txActions,
+  txHistory: txHistoryApi.actions,
```
huesforalice, line 47: I could not get this to propery type the action object with appconnect without inserting the txHistory actions here. Maybe you can try this out @desfero ?

packages/shared-modules/src/modules/tx-history/messages.ts
```diff
@@ -0,0 +1,4 @@
+export enum ETxHistoryMessage {
```
huesforalice, line 1: I feel like it's best if every module has its own set of messages and we bring them together and provide the translations in each of the projects.

packages/web/app/modules/tx-history/module.ts
desfero: yeah, seems there is a bug in type system somewhere

packages/shared-modules/src/modules/tx-history/module.ts
desfero: It works but it's not fully safe. Its just enough to change the order of modules and you will end up using time pooling over `NEW_BLOCK_ARRIVED` action without even noticing it. We should find a better way in future

packages/shared-modules/src/modules/tx-history/sagas.ts
huesforalice: Ah nice, thanks for the tip. Fixed.

packages/web/app/modules/actions.ts
huesforalice: ok

packages/shared-modules/package.json
```diff
@@ -48,15 +48,15 @@
...
+    "yup": "^0.26.7"
   },
   "peerDependencies": {
     "@neufund/sagas": "^1.0.0",
     "@neufund/shared-utils": "^1.0.0",
     "bignumber.js": "^5.0.0",
     "redux": "^4.0.5",
     "tslib": "^1.10.0",
-    "react-redux": "^6.0.0",
-    "yup": "^0.26.7"
```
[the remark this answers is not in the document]
huesforalice: Ok, now I read what peer dependencies are and agree :)

packages/shared-modules/src/modules/tx-history/reducer.ts
```diff
@@ -113,3 +114,11 @@ export const txHistoryReducer: AppReducer<ITxHistoryState> = (
...
   return state;
 };
+
+const txHistoryReducerMap = {
+  txHistory: txHistoryReducer,
+};
+
+export { txHistoryReducerMap };
+
+export type TModuleState = StateFromReducersMapObject<typeof txHistoryReducerMap>;
```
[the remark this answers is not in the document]
huesforalice: Yeah it is, fixed, the duplication.

packages/shared-modules/src/modules/tx-history/module.ts
```diff
@@ -0,0 +1,54 @@
...
+} from "./types";
+
+export { ETxHistoryMessage } from "./messages";
+
+const MODULE_ID = generateSharedModuleId("tx-history");
+
+type Config = Parameters<typeof setupTXHistorySagas>[0];
+
+const setupContainerModule = () =>
+  new ContainerModule(bind => {
```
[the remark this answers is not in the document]
huesforalice: done

packages/web/app/components/wallet/transactions-history/TransactionsHistory.tsx
```diff
@@ -118,23 +116,23 @@ const TransactionListLayout: React.FunctionComponent<TStateProps & TDispatchProp
...
     },
   }),
   onLeaveAction({
     actionCreator: dispatch => {
-      dispatch(actions.txHistory.stopWatchingForNewTransactions());
+      dispatch(txHistoryApi.actions.stopWatchingForNewTransactions());
     },
   }),
-  appConnect<TStateProps, TDispatchProps>({
+  appConnect<TStateProps, TDispatchProps, {}>({
```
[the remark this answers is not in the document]
huesforalice: somehow the linter complained on the other branch, now it seems to be working

packages/shared-modules/src/modules/tx-history/constants.ts
```diff
@@ -0,0 +1,4 @@
...
+export const TX_LIMIT = 20;
```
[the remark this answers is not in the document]
huesforalice: we can divide those later I would say, we can come up with some kind of defaults that can be overriden.

packages/web/app/di/setupBindings.ts
```diff
@@ -268,7 +263,6 @@ export const createGlobalDependencies = (container: Container) => ({
   apiEtoNomineeService: container.get<EtoNomineeApi>(symbols.apiEtoNomineeService),
   apiUserTxService: container.get<UsersTxApi>(symbols.usersTxApi),
   vaultApi: container.get<VaultApi>(symbols.vaultApi),
-  analyticsApi: container.get<AnalyticsApi>(symbols.analyticsApi),
```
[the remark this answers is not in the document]
huesforalice: true! done.

packages/web/app/modules/tx-history/sagas.ts
```diff
@@ -1,424 +1,17 @@
...
-      // Log error and continue looping
-      logger.error("Failed to watch for analytics transaction", e);
-    }
-  }
-}
 
 function* showTransactionDetails(
   _: TGlobalDependencies,
-  action: TActionFromCreator<typeof actions.txHistory.showTransactionDetails>,
+  action: TActionFromCreator<typeof txHistoryApi.actions.showTransactionDetails>,
```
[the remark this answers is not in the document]
huesforalice: I would keep it as we may be able to use it in the wallet too, it would just show a different interface.

packages/web/app/modules/tx-history/module.ts
```diff
@@ -0,0 +1,25 @@
...
+  };
+
+  return [...parentModules, webTxHistoryModule];
+};
+
+export { setupWebTxHistoryModule, txHistoryApi };
```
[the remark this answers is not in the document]
huesforalice: for consistency reasons I put it there, shouldn#t the module shape be the same or very similar for all of them?

packages/web/app/modules/tx-history/module.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const MODULE_ID = "web:tx-history";
+
+const setupWebTxHistoryModule = (config: TModuleConfig) => {
+  const parentModules = setupTxHistoryModule(config);
+
+  const webTxHistoryModule = {
+    id: MODULE_ID,
+    api: txHistoryApi,
```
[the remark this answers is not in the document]
huesforalice: for consistency reasons I put it there, shouldn't the module shape be the same or very similar for all of them?

Issue #4306, opened 2020-04-29
Fix nightly build / Connect slack orb in circleci


desfero:
**Background**
We decided to pay huge attention to the stability of nightly builds and notify slack if the build failed to catch the reason ASAP before it starts slowing down development.

**Designs**

**Acceptance criteria**

**Technical notes**
Use the following orb https://circleci.com/orbs/registry/orb/circleci/slack for slack notifications.

**Testing notes**

PR #4324, opened 2020-05-05
#4306 Fix nightly build

desfero:


packages/web/e2e/e2e-test-tags
```diff
@@ -0,0 +1 @@
+#p1&#p2&#p3
```
desfero, line 1: we need that file to be always available given that danger will only generate it for PR

PR #4334, opened 2020-05-06
Fix circleci job mangling

desfero:
Tests #p3


.circleci/config.yml
```diff
@@ -15,97 +15,99 @@ executors:
...
       - run: yarn workspace @neufund/sagas lint
       - run: yarn workspace @neufund/shared-modules lint
       - run: yarn workspace @neufund/shared-utils lint
       - run: yarn workspace @neufund/web lint
+
       # Unit Tests
       - run: yarn workspace @neufund/shared-utils test
       - run: yarn workspace @neufund/web test
 
-      - save_cache:
```
desfero, line 62: this one is not true given that root package.json is never changing (at least not that often) and all `packages` node_modules are not cached so it kindy useless

.circleci/config.yml
```diff
@@ -15,97 +15,99 @@ executors:
...
 commands:
-  notify_slack:
+  notify-slack:
     description: "Notifies slack about the job failure on master or production branches"
     steps:
       - slack/status:
           fail_only: true
           only_for_branches: master,production
 
+  checkout-and-install:
```
arjunu, line 25: 👍

.circleci/config.yml
```diff
@@ -15,97 +15,99 @@ executors:
     working_directory: /home/circleci/app
 
 commands:
-  notify_slack:
+  notify-slack:
     description: "Notifies slack about the job failure on master or production branches"
     steps:
       - slack/status:
           fail_only: true
           only_for_branches: master,production
```
arjunu, line 24:
```suggestion
failure_message: ':red_circle: $CIRCLE_JOB failed for $CIRCLE_BRANCH!'
```

desfero: done

PR #4336, opened 2020-05-06
adds send typed tx

rudolfix:
Run #p3

`ITxData` and our own typing for Ethereum data types should be later removed. Possibly when we replace `web3` with `ethers`. 
We also have two types for transaction types: `ETxType` for transactions executed from within the app and another enum for transactions in tx history.


packages/web/app/lib/web3/browser-wallet/BrowserWallet.ts
```diff
@@ -88,9 +88,9 @@ export class BrowserWallet implements IPersonalWallet {
     }
   };
 
-  public sendTransaction = async (data: Web3.TxData): Promise<string> => {
+  public sendTransaction = async (txData: ITxData, _: ITxMetadata): Promise<string> => {
```
desfero, line 91: what's the reason to specify the second arg as `_`?

packages/web/app/lib/web3/wallet-connect/utils.ts
```diff
@@ -52,6 +52,7 @@ export const generateWalletMetaFormPeerMeta = (clientMeta: IClientMeta | null):
     } else if (name.includes("neufund")) {
       walletMeta.walletSubType = EWalletSubType.NEUFUND;
       walletMeta.signerType = ESignerType.ETH_SIGN;
+      walletMeta.sendTransactionMethod = ESignTransactionMethod.ETH_SEND_TYPED_TRANSACTION;
```
desfero, line 55: I think it will affect the mobile app as at the moment there is not support for typed transactioin RPC method

PR #4346, opened 2020-05-11
Mobile wallet configure & use absolute imports

arjunu:
Configured and used absolute imports for directories directly under `wallet/app`. The aim is to avoid imports with more than two `../` so `../../` is okay. 

Note the imports have not been reordered. That can be done with ESlint https://github.com/Neufund/platform-frontend/pull/4284.


arjunu:
Closed in favor of https://github.com/Neufund/platform-frontend/pull/4368

PR #4350, opened 2020-05-12
move wallet module to shared modules

huesforalice:
Moves the wallet and the gas price module to shared.

Note: it's based on tx history module moving.


packages/shared-modules/src/modules/contracts/lib/ContractService.ts
```diff
@@ -1,16 +1,34 @@
...
 
 /**
  * An interface syncing ContractsService implementation between platforms (web and mobile)
  */
 interface IContractsService {
+  /**
+   * Special functions (ether balance)
+   */
+  // TODO: move to new eth module
+  balanceOf: (address: EthereumAddressWithChecksum) => Promise<BigNumber>;
```
huesforalice, line 16: I'll move all of this stuff into a new eth module in one of the next PRs

packages/shared-modules/src/modules/wallet/sagas.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+type TSetupSagasConfig = {
+  waitUntilSmartContractsAreInitialized: () => Generator<any, any, any>;
+};
+
+export function setupWalletSagas(config: TSetupSagasConfig): () => SagaGenerator<void> {
+  return function* walletSagas(): any {
+    yield fork(neuTakeEvery, "WALLET_LOAD_WALLET_DATA", loadWalletDataSaga);
+    yield config.waitUntilSmartContractsAreInitialized();
```
huesforalice, line 107: It's not super nice having to inject this here. I would probably try to implement something like this in the new eth module.

packages/shared-modules/src/modules/wallet/module.ts
```diff
@@ -0,0 +1,37 @@
...
+  };
+
+  return [setupTokenPriceModule({ refreshOnAction: undefined }), module];
+};
+
+const walletApi = {
+  actions: walletActions,
+  selectors,
+  utils,
+  sagas,
```
huesforalice, line 32: I had to expose sagas here, because wallet loading is used from another place too. I feel like this is not super nice either, but for the pure migration purpose it should be ok.

packages/wallet/app/modules/contracts/lib/ICBMLockedAccountAdapter.ts
```diff
@@ -0,0 +1,38 @@
...
+  async balanceOf(address: string): Promise<[BigNumber, BigNumber, BigNumber]> {
+    const result = await this.accountContract.balanceOf(address);
+    return [
+      new BigNumber(result[0].toString()),
+      new BigNumber(result[1].toString()),
+      new BigNumber(result[2].toString()),
+    ];
+  }
+
+  readonly currentMigrationTarget: Promise<string> = this.accountContract.currentMigrationTarget();
```
huesforalice, line 35: Somehow this feels not quite right, but it works :)

packages/web/app/lib/web3/ContractsService.ts
```diff
@@ -60,6 +63,10 @@ export class ContractsService implements IContractsService {
 
     this.web3 = this.web3Manager.getFullWeb3Object();
 
+    // TODO: move to new eth module
+    this.balanceOf = (address: EthereumAddressWithChecksum) =>
```
huesforalice, line 67: l8er :)

PR #4355, opened 2020-05-13
Fixes user sign up when Metamask is cancelled

Moejoe90:
This preserves the full init state even when a logout is initiated.


Moejoe90:
Manual tests:

- Register with lightwallet
- Register with metamask
- Login with metamask
- Login with lightwallet
- Changing between flows while registering
- Cancelled metamask and continued lightwallet login flow
- Cancelled metamask and continued lightwallet register flow
- Cancelled metamask and succssefully login in after registering
- Regular Logout

Issue #4263, opened 2020-04-08
Wallet view redesign


andywalton:
**Background** 
The context for this ticket is to update our Wallet view to the redesigned layout of this view.

_A breakdown of the changes:_
* The layout has changed on desktop/tablet to a main column on the left with a narrower aside column to the right of it. This collapses to a single column on on our mobile breakpoint.
* The view has a Wallet balance at the top. This should simply sum up the EUR conversions of the users account wallets.
* Wallet balances now sit in a single wallet component with the wallets listed vertically one above each other as simple list items
* Below is a list of transactions.
* The aside column shows the users wallet address and their linked bank account.

_New functionality:_
1. The Wallet balance total.
2. The modal that tells users to complete their account setup before they can link a bank account for nEUR purchasing. This is handled separately by #4262

**Designs**
<img width="1213" alt="Screenshot 2020-04-08 09 41 38" src="https://user-images.githubusercontent.com/609657/78775374-68e33f00-7996-11ea-8508-7df8a49de678.png">

See Zeplin for the Wallet view screen designs here: https://zpl.io/anRMngl 

Tooltips are here: https://app.zeplin.io/project/5e8d7e1906ea3db82609c137/dashboard?seid=5e8d7fa7c97d1f22978c29fa

See Wallet components here in the Styleguide under the Wallet section here: https://app.zeplin.io/styleguide/5dcae025a4cf6cfed74cab61/components?cseid=5e8d8aaae9617d2379059a57

 See a prototype (hover over the info icons and wallet balance to see tooltips) here:
https://www.figma.com/proto/sBpCxZAZ2wrxW5fgKY5Lhy/Wallet-concept?node-id=159%3A3707&viewport=51%2C582%2C0.14477239549160004&scaling=min-zoom

**Acceptance criteria**

_General:_
- [x] The wallet view layout has changed and matches the designs. Note, the app background is now white: `#ffffff`.
- [ ] As user I can see my wallet balance at the top of the view (this is the EUR sum of my crypto holding. https://zpl.io/bLQrk0J

_Wallet:_
- [ ] The Wallet component takes into account the various states of a crypto asset might be in. See components linked to above.
- [ ] ETH, if balance is greater than 0 you the Send ETH button isn’t disabled: https://zpl.io/2pALWD8
- [ ] nEUR, if the balance is > 0 you can Redeem nEUR: https://zpl.io/be0N8OP
[...]

PR #4363, opened 2020-05-18
#4263 wallet page redesign

AlexeyTuboltsev:
we have a big mess with tables, there are at least two different components plus pending transaction component, that are incompatible with each other 
This should be solved in the upcoming transaction redesign


packages/design-system/src/components/buttons/Button.module.scss
```diff
@@ -44,7 +44,7 @@
 
   &--secondary {
     border: px-to-rem(1px) solid $neu-light-blue-grey-two;
-    background: $neu-white-two;
+    background: $neu-white;
```
AlexeyTuboltsev, line 48: `$neu-white-two` was actually white (#ffffff) and `$neu-white` was a shade of grey. I renamed the former to `$neu-white` and the latter to `$neu-very-light-grey`

packages/design-system/src/assets/img/ETH_32.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
```
AlexeyTuboltsev, line 1: I'm sure we'll be using currency icons many more times in the codebase, so I thought it's not a bad idea to add them to design-system

packages/web/app/components/shared/hooks/useCycleFocus.ts
```diff
@@ -0,0 +1,70 @@
+import * as React from "react";
```
AlexeyTuboltsev, line 1: this would be very helpful for modals 
if we ever get a chance to redesign them

packages/web/app/components/wallet/Balance.spec.tsx
```diff
@@ -0,0 +1,397 @@
+import { tid } from "@neufund/shared-utils/tests";
```
AlexeyTuboltsev, line 1: the purpose of this unit test is to create a full spec of a component's behaviour. That's why it's so repetitive.
But I think every component in the codebase should be spec'ed like this
One thing is missing here - the tooltip texts should be checked for explicitly. This will be much easier after I'll write a variant of FormattedMessage that'd return a simple string (instead of a piece of html, like it does now) so that we'd able to check texts in the test against real translations

packages/web/app/modules/wallet-view/sagas.ts
```diff
@@ -0,0 +1,153 @@
+import { all, call, fork, put, select, take } from "@neufund/sagas";
```
AlexeyTuboltsev, line 1: What's happening in this file:
there's a `walletViewController` saga, it takes care of the whole lifecylcle of the view - initial load, rebuild on new data, tear down tasks. 
there's a `loadWalletView` saga taht's actually performing the work of generating the data for the view. In this case it's the data to show wallet balances. This data is then "hydrated" in the appConnect of the wallet view component, e.g. the real markup and translations are selected based on this data.
I didn't want to touch the tx history in this pr, but it should be refactored IMO because the view and data are too tightly coupled there. After that the tx history data should be populated in this saga too, so that we'd have one source of truth

packages/web/app/modules/wallet-view/types.ts
```diff
@@ -0,0 +1,70 @@
...
+import { TBankAccount } from "../kyc/types";
+
+export type TBasicBalanceData = {
+  name: EBalanceType;
+  hasFunds: boolean;
+  amount: string;
+  euroEquivalentAmount: string;
+};
+
+export enum EBalanceType {
```
AlexeyTuboltsev, line 14: this enum doesn't reflect our internal data structures, these are just the types of UI we use in the wallet view. That's why `neur` and `restricted_neu`r are separated, as well as `icbm` and `locked_icbm`

packages/web/app/components/wallet/transactions-history/PendingTransaction.tsx
```diff
@@ -0,0 +1,85 @@
+import * as cn from "classnames";
```
Moejoe90, line 1: if you're only testing PendingTransaction for story book IMO you only have two cases to test in that case 
Default 
and
TxType = `NOMINEE_THA_SIGN`

packages/web/app/components/wallet/transactions-history/TransactionsHistory.module.scss
```diff
@@ -16,3 +29,96 @@
...
+  flex-flow: column;
+  justify-content: center;
+  align-items: flex-end;
+  align-self: flex-end;
+
+  padding-right: px-to-rem(20);
+
+  font-family: $font-family-roboto;
+  font-size: px-to-rem(14);
+  font-weight: 500;
```
Moejoe90, line 105: please take from typography

packages/web/app/components/wallet/utils.tsx
```diff
@@ -0,0 +1,155 @@
...
+  [EBalanceType.ETH]: "wallet-balance.eth.balance-value",
+  [EBalanceType.NEUR]: "wallet-balance.neur.balance-value",
+  [EBalanceType.RESTRICTED_NEUR]: "wallet-balance.restricted-neur.balance-value",
+  [EBalanceType.ICBM_ETH]: "icbm-wallet.eth.balance-value",
+  [EBalanceType.ICBM_NEUR]: "icbm-wallet.neur.balance-value",
+  [EBalanceType.LOCKED_ICBM_ETH]: "locked-icbm-wallet.eth.balance-value",
+  [EBalanceType.LOCKED_ICBM_NEUR]: "locked-icbm-wallet.neur.balance-value",
+};
+
+export const createBalanceActions = (dispatch: Function): TBalanceActions => ({
```
Moejoe90, line 84: why not test them? the test should be a simple assert that dispatch was used correctly

packages/web/app/modules/wallet-view/sagas.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    amount: "0",
+    euroEquivalentAmount: "0",
+  },
+];
+
+describe("Wallet View", () => {
+  describe("populateWalletData()", () => {
+    it("will populate wallet data, no funds anywhere", async () => {
+      await expectSaga(populateWalletData)
+        .provide([
```
Moejoe90, line 75: I would do this differently 
No need to overide the selectors just use `withReducer` and attach the correct reducer the selectors should select correctly from the state. 

With the test will be more realistic. There are already tests that use `withReducer` in the web application

packages/web/app/modules/wallet-view/types.ts
Moejoe90: Maybe we should add this as a comment to not mix things up, as other developers might assume this is our internal data structure

packages/web/app/modules/wallet-view/types.ts
```diff
@@ -0,0 +1,70 @@
...
+  balanceAdditionalInfo: TTranslatedString | undefined;
+  amount: string;
+  currency: ECurrency;
+  euroEquivalentAmount: string;
+  walletActions: TBalanceAction[];
+  dataTestId?: string;
+};
+
+export type TBalanceAction = {
+  dispatchAction: (x: unknown) => void;
```
Moejoe90, line 63: why is it unknown it should be known its some action type that should be only inclined with this module or is it not the case

packages/web/app/modules/wallet-view/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+      hasFunds: compareBigNumbers(lockedIcbmNeuroWalletData.amount, "0") > 0,
+      amount: lockedIcbmNeuroWalletData.amount,
+      euroEquivalentAmount: lockedIcbmNeuroWalletData.euroEquivalentAmount,
+    },
+  ];
+}
+
+export function* loadWalletView(): Generator<any, void, any> {
+  try {
+    yield all([neuCall(loadWalletDataSaga), neuCall(loadBankAccountDetails)]);
```
Moejoe90, line 105: why do we need to yield all here? 

Its obsucers the steps that are happening IMO

packages/web/app/modules/wallet-view/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const userIsFullyVerified = yield* select(selectIsUserFullyVerified);
+    const userAddress = yield* select(selectEthereumAddress);
+    const bankAccount = yield* select(selectBankAccount);
+
+    const balanceData = (yield call(populateWalletData)).filter(
```
Moejoe90, line 111: you should use `yield *`

packages/web/app/modules/wallet-view/sagas.ts
```diff
@@ -0,0 +1,153 @@
...
+        userIsFullyVerified,
+        userAddress,
+        balanceData,
+        totalBalanceEuro,
+        bankAccount,
+        processState: EProcessState.SUCCESS,
+      }),
+    );
+  } catch (e) {
+    yield put(actions.walletView.walletViewSetData({ processState: EProcessState.ERROR }));
```
Moejoe90, line 130: ok I can see that is how you are setting the error IMO 
you should explicitly use the action you have already set in `actions` 

In all cases we either need to remove that action or use it here. for readability I would use the dedicated action as in the future we might need to add some metadata with the action and thus type of `walletViewSetData` will get more complicated

packages/design-system/src/components/tables/Table.unsafe.tsx
```diff
@@ -91,11 +96,10 @@ const Table: React.FunctionComponent<TExternalProps> = ({
     columns,
     data,
   });
-
   return (
     <div className={styles.container}>
-      <table className={styles.table} {...getTableProps()}>
-        <TableHeader headerGroups={headerGroups} />
+      <table className={cn(styles.table, className)} {...getTableProps()}>
+        {CustomHeader ? <CustomHeader /> : <TableHeader headerGroups={headerGroups} />}
```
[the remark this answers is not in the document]
AlexeyTuboltsev: sorry, this is a leftover. 
As i said we have a bit of a mess with tables at the moment and at first I spent tons of time try to figure out how to make them usable for the new wallet designs
-fixed

packages/design-system/src/components/typography/DashboardTitle.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
```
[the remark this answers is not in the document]
AlexeyTuboltsev: it's not used anymore

packages/web/app/components/layouts/Content.module.scss
```diff
@@ -10,7 +10,7 @@
 
     @include media-breakpoint-up (md) {
       padding: 0 $layout-content-padding-lg;
-      margin: $layout-content-padding-lg auto;
+      margin: px-to-rem(49) auto $layout-content-padding-lg;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: done

packages/web/app/components/shared/AccountAddress.module.scss
```diff
@@ -40,11 +50,56 @@
...
+    padding: px-to-rem(12);
+    border-radius: 4px;
+  }
+
+  .address {
+    @include address;
+
+    font-family: $font-family-roboto;
+    font-size: px-to-rem(14);
+    font-weight: normal;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: done

packages/web/app/components/shared/AccountAddress.module.scss
```diff
@@ -40,11 +50,56 @@
...
+    width: px-to-rem(24);
+    height: px-to-rem(24);
+  }
 
-    // force to not overflow
-    max-width: 100%;
+  .history-link {
+    font-family: $font-family-roboto;
+    font-size: px-to-rem(12);
+    color: $neu-shuttle-gray;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: done

packages/web/app/components/shared/hooks/useCycleFocus.ts
AlexeyTuboltsev: it's fully written by me, except for some googling

packages/web/app/components/shared/tooltips/Tooltip.tsx
```diff
@@ -1,12 +1,13 @@
+import { InlineIcon } from "@neufund/design-system";
 import { makeTid } from "@neufund/shared-utils";
 import * as cn from "classnames";
 import * as React from "react";
 
 import { CommonHtmlProps, TDataTestId, TTranslatedString } from "../../../types";
-import { InlineIcon } from "../icons";
+import { EInlineIconFill } from "../icons";
```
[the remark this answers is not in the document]
AlexeyTuboltsev: this is not my code... There are two ways to apply color to svg - fill and shape, i guess that's what Pawel meant

packages/web/app/components/wallet/Balance.tsx
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const BalanceActions: React.FunctionComponent<TBalance> = props => {
+  const { walletActions } = props;
+
+  const [isOn, toggle] = React.useState(false);
+
+  const toggleBalanceActionsRef = React.useRef<HTMLButtonElement>(null);
+  const balanceActionRefs = walletActions.map(_ => React.useRef<HTMLButtonElement>(null));
+  const allRefs = [toggleBalanceActionsRef, ...[...balanceActionRefs].reverse()]; //reverse the button order to reflect the visual ordering
```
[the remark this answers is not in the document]
AlexeyTuboltsev: it's just UX - the cycling order should match the visual order. Take a look at the designs:
https://app.zeplin.io/project/5e8d7e1906ea3db82609c137/screen/5e8d7e65e11e8db9ce41753e

packages/web/app/components/wallet/Balance.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const moveFocusOnTabKey = useCycleFocus(allRefs);
```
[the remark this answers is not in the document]
AlexeyTuboltsev: that's what I did - useCycleFocus takes an array of refs and cycles focus over them. 
The code before is preparing this array of refs. 

You don't have to write this anymore for each component, just use the hook

packages/web/app/components/wallet/BankAccount.tsx
```diff
@@ -1,40 +1,79 @@
...
+}) => (
+  <Container className={styles.noLinkedBankAccountWrapper} columnSpan={EColumnSpan.ONE_COL}>
+    <div className={styles.subtitle}>
+      <FormattedMessage id="wallet.linked-bank-account-title" />
+    </div>
+    <div className={styles.noLinkedBankAccount}>
+      <FormattedHTMLMessage
+        tagName="span"
+        id="wallet.no-linked-bank-account"
+        values={{ href: "" }}
```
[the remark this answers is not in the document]
AlexeyTuboltsev: oh, good catch
done

packages/web/app/components/wallet/Wallet.module.scss
```diff
@@ -0,0 +1,219 @@
+@mixin table-text-style {
+  font-family: $font-family-roboto;
+  font-size: 1rem;
+  font-weight: 500;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: done

packages/web/app/components/wallet/Wallet.module.scss
```diff
@@ -0,0 +1,219 @@
...
+.no-linked-bank-account-wrapper {
+  display: grid;
+  grid-template: auto auto / 1fr;
+  grid-row-gap: px-to-rem(12);
+}
+
+.total-balance-title {
+  font-family: $font-family-roboto;
+  font-size: px-to-rem(14);
+  font-weight: 500;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: done

packages/web/app/components/wallet/Wallet.module.scss
```diff
@@ -0,0 +1,219 @@
...
+
+  &.active {
+    @include balance-actions-visible;
+    position: absolute;
+    right: px-to-rem(64);
+    opacity: 1;
+  }
+}
+
+
```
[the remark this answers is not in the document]
AlexeyTuboltsev: done

packages/web/app/components/wallet/transactions-history/PendingTransaction.stories.tsx
```diff
@@ -1,58 +1,57 @@
...
-import { Transaction } from "./Transaction";
+import { PendingTransaction } from "./PendingTransaction";
 
 const fromAddress = "0x16cd5aC5A1b77FB72032E3A09E91A98bB21D8989";
 storiesOf("Molecules|Transaction", module)
   .add("Withdraw", () => {
     const transaction: TxPendingWithMetadata = generalPendingTxFixture(fromAddress);
 
-    return <Transaction transaction={transaction} />;
+    return <PendingTransaction transaction={transaction} />;
```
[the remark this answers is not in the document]
AlexeyTuboltsev: I renamed transaction to pending transaction, because that's what it is.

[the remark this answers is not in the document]
AlexeyTuboltsev: added storybook test for regular transaction

packages/web/app/components/wallet/transactions-history/PendingTransaction.tsx
AlexeyTuboltsev: that was the test that Dima wrote. Should I refactor it?

packages/web/app/components/wallet/transactions-history/TransactionsHistory.module.scss
AlexeyTuboltsev: done

packages/web/app/components/wallet/utils.tsx
AlexeyTuboltsev: not sure what you mean...

these are just lists of mappings of raw data from sagas to "hydrated" data for UI. The correctness is checked in the `packages/web/app/components/wallet/Balance.spec.tsx`

packages/web/app/modules/wallet-view/reducer.ts
```diff
@@ -0,0 +1,21 @@
...
+const walletViewInitialState: TWalletViewState = {
+  processState: EProcessState.NOT_STARTED,
+};
+
+export const walletViewReducer: AppReducer<TWalletViewState> = (
+  state = walletViewInitialState,
+  action,
+) => {
+  switch (action.type) {
+    case actions.walletView.walletViewSetData.getType():
```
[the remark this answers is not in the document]
AlexeyTuboltsev: error is a variant of the data, no need for a special action

packages/web/app/modules/wallet-view/sagas.ts
AlexeyTuboltsev: both of these sagas are triggering api calls that can take quite some time.. 
I think `all` fits well here, as it executes lengthy side effects in a batch.

packages/web/app/modules/wallet-view/sagas.ts
AlexeyTuboltsev: sure

packages/web/app/modules/wallet-view/sagas.ts
AlexeyTuboltsev: I added a generic error for now. Looks clean to me,  how about you?  :)

packages/web/app/modules/wallet-view/types.ts
AlexeyTuboltsev: 👍

packages/web/app/modules/wallet-view/types.ts
AlexeyTuboltsev: no, I can't know beforehand what action with what kind of arguments, or with how many arguments goes here.

packages/web/app/test-e2e/investor/wallet/utils.ts
```diff
@@ -89,7 +89,7 @@ export const typeWithdrawForm = (testAddress: string, testValue: string) => {
 };
 
 export const continueWithdrawFlow = (testAddress: string, testValue: string) => {
-  cy.get(tid("account-address.your.ether-address.from-div"))
+  cy.get(tid("account-address.your.ether-address.from-div.trigger"))
```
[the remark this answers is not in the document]
AlexeyTuboltsev: I added a tooltip to the address, so I'm using the `tid`'s that are already there. Not my code :))

packages/web/app/test-e2e/vis-reg/Investor.spec.e2e.visreg.ts
```diff
@@ -28,10 +28,9 @@ describe("Investor", () => {
     cy.screenshot();
   });
 
-  it("should render wallet", () => {
+  it.only("should render wallet", () => {
```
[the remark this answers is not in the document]
AlexeyTuboltsev: wow. How did this pass linter????

packages/web/app/components/shared/hooks/useCycleFocus.ts
AlexeyTuboltsev: test added

packages/web/app/modules/wallet-view/sagas.spec.ts
AlexeyTuboltsev: fixed

packages/web/app/components/wallet/Wallet.module.scss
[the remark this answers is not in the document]
Moejoe90: I would add all styles to to typography because we will lose control over things if we don't keep doing this

packages/web/app/components/wallet/transactions-history/PendingTransaction.tsx
Moejoe90: Yes you are right for Transactions, is this is also the case for `PendingTransactions`? That was the point of the comment

packages/web/app/components/wallet/utils.tsx
Moejoe90: It would be good to have a direct connection where a test touches the utility function directly. 
Will this take a long time to write?

packages/web/app/modules/wallet-view/sagas.ts
Moejoe90: I would have a special error action for it but its not really blocking. As long as there are no redundant actions

packages/web/app/components/shared/hooks/useCycleFocus.spec.tsx
```diff
@@ -0,0 +1,56 @@
+import { expect } from "chai";
```
Moejoe90, line 1: :+1:

PR #4367, opened 2020-05-19
Add module documentation

Moejoe90:


packages/shared-modules/README.md
```diff
@@ -1,10 +1,54 @@
...
+# Shared Modules
 
 [![Project dependencies Status](https://david-dm.org/Neufund/platform-frontend/status.svg?path=packages/shared-modules)](https://david-dm.org/Neufund/platform-frontend?path=packages/shared-modules)
 [![Project devDependencies Status](https://david-dm.org/Neufund/platform-frontend/dev-status.svg?path=packages/shared-modules)](https://david-dm.org/Neufund/platform-frontend?type=dev&path=packages/shared-modules)
 
 Shared modules for use between web and mobile apps.
 
+## Module Design
+
+Each module should function as an independent unit with no direct dependency on other modules
```
huesforalice, line 10: This would be nice, but it won't happen. The way we want to do it is to only have dependencies on the api objects that modules share among each other, this way it should also be easy to mock stuff for tests. See the shared modules for examples.

packages/shared-modules/README.md
```diff
@@ -1,10 +1,54 @@
...
+
+A set of actions that will either trigger a state change and/or trigger a listening saga. using
+`takeEvery` for example.
+
+### Sagas
+
+Sagas should generally be split into two categories:
+
+- Internal sagas that should run only internal and will never get exposed. Only the module is can
+  use these sagas
```
AlexeyTuboltsev, line 27: typo: 
only the module can use these sagas

packages/shared-modules/README.md
```diff
@@ -1,10 +1,54 @@
...
+- External sagas, these exposed sagas should be in effect with actions the main interface to
+  interact with this module.
+
+### Selectors
+
+Selectors should be used only for data fetching and by no way should convert the data in anyway from
+the reducer except in clear cases where the data needs to be normalized before use. If that is the
+case the selector should have a clear comment with what is the original data form and what is the
+data being normalized to.
+
```
AlexeyTuboltsev, line 38: I'd say selectors shouldn't convert any data at all. (very very trivial cases aside). The data should be normalized before storage -  explicitly by the saga that acquired it, via utility fns, or after being selected -  by other sagas that want to use it for their purposes. 
For now we have tons of conversions and logic inside selectors

arjunu: Sometimes conversion is unavoidable. We may have a component that needs data in a different way.

packages/shared-modules/README.md
arjunu: We're already using one module in another, aren't we?

Moejoe90: What do you mean by `api objects`? 
What was initially designed and could be still the case is that modules shouldn't depend on each other. in terms of for example you can only communicate with another module via its external API. 

You already mentioned for example today the issues you found in the ETO Modules where things are a bit too mish mashed that clearly indicates that the ETO modules needs rethinking 

Thats' whats meant by the sentance up it could change of course to something that conveys this meaning. 

We also need to seperate between injected instances like `Storage` and `Modules`

packages/shared-modules/README.md
Moejoe90: Lets take the case where you are dealing with `BigNumbers` the selector would take a string from the state normalize it to a BigNumber and returns it. That is cleaner IMO than having to initialize a BigNumber after each time you select that value from the state. 

There are a couple of other examples I can think of that has the same trait.

packages/shared-modules/README.md
desfero: Yeah, l also think we should rephrase it a little bit as for .e.g all modules are in reality dependent on the core module (because they use `logger`/`http` symbols.) A lot of modules gonna be with relation `logic -> ui` (the way `shareholder-resolution-voting` and `shareholder-resolutiion-voting-view` are done)

packages/shared-modules/README.md
```diff
@@ -1,10 +1,54 @@
...
+Each module should function as an independent unit with no direct dependency on other modules
+
+### Reducer
+
+A reducer should hold the internal state of the module only. The reducer shouldn't react to any
+external actions outside of the actions defined in the module.
+
+### Actions
+
+A set of actions that will either trigger a state change and/or trigger a listening saga. using
```
desfero, line 19:
```suggestion
A set of actions that will either trigger a state change and/or trigger a listening saga using
```

packages/shared-modules/README.md
```diff
@@ -1,10 +1,54 @@
...
+used internally. If more than one module ends up using the same utility method the method should be
+moved to `shared-util` package
+
+### constants
+
+All constants should be defined in the constant file. constants can be exposed externally.
+
+internal sagas that shouldn't be exposed and external sagas that should be the main method of
+communication with the module externally.
+
```
desfero, line 51: I would also mention the `api` of a module which expect `actions` may contain `symbols`, `errors` and whatever is needed

Moejoe90:
Will temporary close for the sake of the new RFC idea

PR #4368, opened 2020-05-19
Wallet configure and use absolute imports

arjunu:
- Configured and used absolute imports for directories directly under `wallet/app`. The aim is to avoid imports with `../` 
- also merged the two different asset directories


packages/wallet/ios/wallet.xcodeproj/project.pbxproj
```diff
@@ -15,9 +15,9 @@
 		44C078612420CDC000CD0C0F /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = 44C078602420CDC000CD0C0F /* GoogleService-Info.plist */; };
 		965F0307243B1D190055C7D0 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 965F0306243B1D190055C7D0 /* LaunchScreen.storyboard */; };
 		BA577FB11DA5E440D4B93FF2 /* libPods-wallet.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 0F90EC4E5707298CEB599B49 /* libPods-wallet.a */; };
-		C2C8D5B1B8D14A8096C562CB /* Roboto-Regular.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 8664C35363264F6D9A1F72B2 /* Roboto-Regular.ttf */; };
-		CDEE3ED7D2894CE8AD0C2A44 /* Roboto-Medium.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 5C3B1B79A78840FBAD4EFBDB /* Roboto-Medium.ttf */; };
-		CE662478B3B2439E9EF842E1 /* Montserrat-SemiBold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 6C6451630EA742FAA6F9D489 /* Montserrat-SemiBold.ttf */; };
+		41D7C7554656486D8F2A1095 /* Montserrat-SemiBold.ttf in Resources */ = {isa = PBXBuildFile; fileRef = D7ECC95DD7A44468B165BFFE /* Montserrat-SemiBold.ttf */; };
```
arjunu, line 18: got conflicts here, had to re-link resources.

packages/wallet/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "allowJs": true,
-    "baseUrl": "./",
+    "baseUrl": "./app",
```
desfero, line 4: i would say we should still keep base url as `./` as it's where the project starts (`index.tsx` file)

packages/wallet/babel.config.js
```diff
@@ -3,5 +3,25 @@ module.exports = {
   plugins: [
     ["@babel/plugin-proposal-decorators", { legacy: true }],
     "babel-plugin-parameter-decorator",
+    [
+      "module-resolver",
+      {
+        root: ["./app"],
+        extensions: [".ts", ".tsx", ".js", ".ios.js", ".android.js"],
+        alias: {
+          assets: "./app/assets",
```
desfero, line 12: we can read it from tsconfig and put it here so then there is no need to duplicate the same list

packages/wallet/.eslintrc.js
```diff
@@ -16,6 +16,78 @@ module.exports = {
...
+      {
+        groups: [
+          ["builtin", "external"],
+          ["internal", "parent", "sibling", "index"],
+        ],
+        "newlines-between": "always",
+        alphabetize: {
+          order: "asc",
+        },
+        pathGroups: [
```
desfero, line 30: We can read path groups from tsconfig map them a little bit and display here

packages/wallet/.eslintrc.js
```diff
@@ -16,6 +16,78 @@ module.exports = {
...
+          {
+            pattern: "utils/**",
+            group: "external",
+            position: "after",
+          },
+        ],
+        pathGroupsExcludedImportTypes: ["builtin"],
+      },
+    ],
+    "import/no-relative-parent-imports": "error",
```
desfero, line 90: this won't catch import errors from the same or child directories 

For .e.g in `App.tsx` replace 
```
import { CriticalError } from "components/CriticalError";
```
with 
```
import { CriticalError } from "./components/CriticalError";
```

packages/wallet/.eslintrc.js
arjunu: done

packages/wallet/babel.config.js
arjunu: done

packages/wallet/tsconfig.json
arjunu: changed

packages/wallet/.eslintrc.js
arjunu: there's no eslint rule to catch this :(

Issue #4353, opened 2020-05-12
Fix android loading animation


desfero:
**Description**
There is some inconsistency in how loading animation behaves between android and ios devices (not all android devices are affected).

**Environment**

**Steps To Reproduce**
1. Go to the landing screen
2. Click create a new user 

**Actual behaviour**
The animation is quite sharp

**Expected behaviour**
The same behavior as on the web or ios

**Comments, context, logs**

**Screenshots**
![ezgif-6-77ddef21b7d1](https://user-images.githubusercontent.com/7031578/81725225-d0087d80-9485-11ea-856c-448d4c796275.gif)

PR #4370, opened 2020-05-20
#4353 fix android loading animation

desfero:
**Testing notes**
This PR also blocks landscape orientation on android devices

Android https://i.diawi.com/Hfx8sy
IOS: https://i.diawi.com/GgPQjC


packages/wallet/index.tsx
```diff
@@ -32,19 +28,28 @@ function startupApp(): void {
 }
 
 function renderApp(store: IModuleStore<TAppGlobalState>): void {
-  const Component = () => (
-    <ReduxProvider store={store}>
-      <AppContainer>
+  const Component = () => {
```
desfero, line 31: just to note for a reviewer, the code here is partially invalid as we end up bundling storybook in release build. We should probably separate index.tsx for release and debug in future

PR #4373, opened 2020-05-24
filters our test, mock and library artifacts from wrapping

rudolfix:
Smart Contract Artifacts that will be deployed with backend in https://github.com/Neufund/platform-contracts/pull/340 contain Solidity libraries for which internal types are exposed in ABI and typechain cannot parse them.
We still use 0.4.26 artifacts that do not hide internal types.

This PR filters out unnecessary artifacts from wrapping
1. libraries (never called directly anyway)
2. test and mock contracts


packages/wallet/scripts/generate-contracts.js
```diff
@@ -48,7 +51,18 @@ async function generateFixtures() {
   );
 }
 
-async function generateTypechainWrappers() {
+async function filterArtifacts(sourcePath) {
+  const tmpdir = await fs.promises.mkdtemp(`${os.tmpdir()}/artifacts`);
```
desfero, line 55: not really sure whether it can handle separators in a consistent way across systems, maybe we should use 
`fs.promises.mkdtemp(path.join(os.tmpdir(), 'neufund-artifacts));`?

packages/wallet/scripts/generate-contracts.js
```diff
@@ -15,7 +16,9 @@ const artifactsVersion = process.env.NF_CONTRACT_ARTIFACTS_VERSION || "localhost
 
 const outDir = "app/lib/contracts";
 
-async function generateKnownInterfaces() {
+const filteredArtifacts = ["Gov.json", "Test.+.json", "Mock.+.json"].map(r => RegExp(r));
```
desfero, line 19: do we want to have `.(any-character)json` or `\.(dot)json`?

packages/wallet/scripts/generate-contracts.js
```diff
@@ -48,7 +51,18 @@ async function generateFixtures() {
   );
 }
 
-async function generateTypechainWrappers() {
+async function filterArtifacts(sourcePath) {
+  const tmpdir = await fs.promises.mkdtemp(`${os.tmpdir()}/artifacts`);
+  const dir = await fs.promises.opendir(sourcePath);
+  for await (const dirent of dir) {
+    if (filteredArtifacts.reduce((prev, curr) => !curr.test(dirent.name) && prev, true)) {
+      await fs.promises.copyFile(`${sourcePath}${dirent.name}`, `${tmpdir}/${dirent.name}`);
```
desfero, line 59: also `path.join` can be used here

packages/wallet/scripts/generate-contracts.js
```diff
@@ -48,7 +51,18 @@ async function generateFixtures() {
   );
 }
 
-async function generateTypechainWrappers() {
+async function filterArtifacts(sourcePath) {
+  const tmpdir = await fs.promises.mkdtemp(`${os.tmpdir()}/artifacts`);
+  const dir = await fs.promises.opendir(sourcePath);
+  for await (const dirent of dir) {
+    if (filteredArtifacts.reduce((prev, curr) => !curr.test(dirent.name) && prev, true)) {
```
desfero, line 58: A good example for Array.prototype.some, gonna be more readable than reduce

packages/wallet/scripts/generate-contracts.js
```diff
@@ -48,7 +51,18 @@ async function generateFixtures() {
...
+      await fs.promises.copyFile(`${sourcePath}${dirent.name}`, `${tmpdir}/${dirent.name}`);
+    }
+  }
+  return tmpdir;
```
desfero, line 62: what's the lifetime of a tmpdir? Will OS clean it up ASAP or it gonna leave in the temp directory longer than we really need?

rudolfix: it will be left with all the content. is this really a big problem?

desfero: it depends, artifacts size is around 8mb now, given dev runs `yarn install` 2-5 times a day gives us around 25mb each day. 25mb * 21 days  = 512mb, not a huge problem. Let's hope that OS clears the temp folders when it needs resources

Issue #4288, opened 2020-04-22
Generic modal component


andywalton:
**Background**
<img width="1107" alt="Screenshot 2020-04-22 12 54 15" src="https://user-images.githubusercontent.com/609657/79973761-673a7080-8498-11ea-9e4f-63d1ed93e989.png">

The context for this ticket is to create the a generic modal component for use with the mobile app.

We use modals to highlight information that might require the user to take action on on make a decision about.

Our modals can therefore be used for simple primer screens or to sign transactions or messages or for useful information.
 For this ticket, the modal has the following breakdown:
1. An optional Icon/Image at the top
2. A modal title
3. A modal body (paragraph of text)
4. Up to 2 call to action buttons (Both would action and close the modal)
5. 1-4 sit on a modal container card (white background) which sits on an overlay scrim (transparent dark background) 

**Designs**
See Zeplin for the screen designs here under the Modal & Signers sections: https://app.zeplin.io/project/5e6f48a43076c61012fd3793/dashboard?seid=5ea0145ebe207252d49d8c27

This is the generic modal we’re going for: https://zpl.io/2G00X9d 

> **Take note** 
> Both the transaction and message signers will use this this generic modal template. The transaction signer will simply have a modal title that describes the transaction type and a modal body that is a simple line of text to ask you to sign to confirm the transaction.
> 
> Example simple transaction signer: https://zpl.io/2G00j63 
> Example message signer: https://zpl.io/VQ55NGm

See design system components here: https://app.zeplin.io/styleguide/5e5f852f2f231b17c87f8622/components?cseid=5ea01fb09264e45331b7300a

 **Acceptance criteria**
- [ ] The modal is as per the designs and works as per the modal content it’s displays.
- [ ] The transaction & message signers uses this generic modal also - ie. for transactions we don't display complex transaction meta details and data.

**Technical notes**
@rudolfix @desfero -> please add technical notes here.

**Testing notes**
@glupiawymowka -> Same as acceptance criteria but edit this if you want.

PR #4374, opened 2020-05-24
#4288 Generic signer modal

desfero:
TODO:
- [x] Write unit test for BaseAnimation component

Android: https://i.diawi.com/Yc3q8c
IOS: https://i.diawi.com/zJxXBh


packages/wallet/app/modules/signer-ui/reducer.ts
```diff
@@ -1,7 +1,7 @@
 import { AppReducer } from "@neufund/sagas";
 
 import { signerUIActions } from "./actions";
-import { TSignerRequestData } from "./types";
+import { TSignerSignPayload } from "./types";
 
 export enum ESignerUIState {
   IDLE = "idle",
```
arjunu, line 7: do you think "hidden" is more appropriate than "idle"?

arjunu:
- there's a typo in the file name `BottomSheetModa.stories.tsx` and it fails to load 
image

- can you add stories for the three types of modal?

packages/wallet/storybook/stories.ts
```diff
@@ -4,6 +4,7 @@ import "../app/components/shared/LineBreak.stories";
 import "../app/components/shared/Icon.stories";
 import "../app/components/shared/Link.stories";
 import "../app/components/shared/asset/Asset.stories";
+import "../app/components/shared/modals/BottomSheetModa.stories";
```
arjunu, line 7: BottomSheetModal*

packages/wallet/app/components/shared/animations/BaseAnimation.tsx
```diff
@@ -0,0 +1,112 @@
...
+    if (prevProps.isActive !== this.props.isActive) {
+      if (this.props.isActive) {
+        this.showAnimation();
+      } else {
+        this.hideAnimation();
+      }
+    }
+  }
+
+  showAnimation() {
```
arjunu, line 65: `startAnimation` and `finishAnimation` seems more appropriate.

packages/wallet/app/components/shared/animations/BaseAnimation.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const { animationProgress } = this.state;
+
+    Animated.timing(animationProgress, {
+      duration: 300,
```
arjunu, line 69: can you move these to constants?
SHOW_ANIMATION_DURATION
HIDE_ANIMATION_DURATION

packages/wallet/app/components/shared/animations/BaseAnimation.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+    const renderProp = () => <Text testID="content">Animation content</Text>;
+    const { queryByTestId, rerender } = render(
+      <BaseAnimation isActive={true} render={renderProp} />,
+    );
+
+    rerender(<BaseAnimation isActive={false} render={renderProp} />);
+
+    expect(queryByTestId("content")).toBeDefined();
+
+    timeTravel(200);
```
arjunu, line 44: timeTravel(HIDE_ANIMATION_DURATION)

packages/wallet/app/modules/signer-ui/reducer.ts
desfero: no sure now let's see how the enum evolves when we add support for transaction from the wallet app

packages/wallet/storybook/stories.ts
desfero: wow i'm curious why it was not caught by typescript compiler

packages/wallet/app/components/shared/animations/BaseAnimation.tsx
desfero: Even better, moved to props so there is control over how long animation should be in progress

packages/wallet/app/components/shared/animations/BaseAnimation.spec.tsx
desfero: done

Issue #2638, opened 2019-03-25
Change Transaction Success Confetti to SUCCESS GREEN across the board


Louisswryan:
Hello,

Currently transactions are rewarded with the below animated confetti:

![TRANSACTION_SUCCESS.png](https://images.zenhubusercontent.com/5c5184c7df0733a30553c211/7abb5c42-3f2c-47f5-9a34-c4c6ad910ff9)

Could we please change this confetti to just our SUCCESS GREEN
    --success: #20e06e;

Also, successfully linking a new bank account is rewarded with confetti in NEUFUND GREEN

![NEW_BANK_ACC.png](https://images.zenhubusercontent.com/5c5184c7df0733a30553c211/237eabec-eb93-44a7-aa81-f28e819820a4)

Can we please change this to SUCCESS also

In the end, we want all confetti to look like the below.

![CONFETTI_CORRECT.png](https://images.zenhubusercontent.com/5c5184c7df0733a30553c211/927a7303-a2eb-4a50-b624-919a5b3505b8)

I will highlight other examples on inconsistency if I see them but if you notice others, please change them also and note them in this ticket.

Thank You!

PR #4381, opened 2020-05-26
#2638 iframe csp

AlexeyTuboltsev:
this PR adds csp headers/url sanitization to twitter and slideshare iframes


packages/web/app/external/sanitize-url/sanitize-url.js
```diff
@@ -0,0 +1,40 @@
...
+var invalidPrototcolRegex = /^(%20|\s)*(javascript|data)/im;
+var ctrlCharactersRegex = /[^\x20-\x7EÀ-ž]/gim;
+var urlSchemeRegex = /^([^:]+):/gm;
+var relativeFirstCharacters = [".", "/"];
+
+function isRelativeUrlWithoutProtocol(url) {
+  return relativeFirstCharacters.indexOf(url[0]) > -1;
+}
+
+function sanitizeUrl(url) {
```
desfero, line 14: maybe some unit tests to make sure it's working properly always?

packages/web/app/external/pitch-deck-iframe.html
```diff
@@ -3,6 +3,10 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self'; frame-src *; script-src 'self' 'sha256-80vXT2q9DtdrNkWOqKM45NjhN0z2WCwccwwrAQbJw2E='; style-src 'unsafe-inline'; img-src 'self';"
```
desfero, line 8: What kind of js file `sha256-80vXT2q9DtdrNkWOqKM45NjhN0z2WCwccwwrAQbJw2E=` represents?

packages/web/app/components/shared/TwitterTimeline.tsx
```diff
@@ -16,8 +16,11 @@ export class TwitterTimelineEmbed extends React.Component<IProps> {
         width="100%"
         height="100%"
         frameBorder={0}
-        allowFullScreen
+        allowFullScreen={false}
```
desfero, line 19: maybe we can completely omit the attribute if it's false?

packages/web/app/external/pitch-deck-iframe.html
AlexeyTuboltsev: I think it represents  scripts inside <script></script> tags. 
see here: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

that means effectively that even if everything else fails only this script will be run

packages/web/app/components/shared/TwitterTimeline.tsx
AlexeyTuboltsev: absolutely

packages/web/app/external/sanitize-url/sanitize-url.spec.js
```diff
@@ -0,0 +1,140 @@
...
+
+  it("disregards capitalization for JavaScript urls", () => {
+    expect(sanitizeUrl("jAvasCrIPT:alert(document.domain)")).to.eq("");
+  });
+
+  it("ignores ctrl characters in javascript urls", () => {
+    expect(sanitizeUrl(decodeURIComponent("JaVaScRiP%0at:alert(document.domain)"))).to.eq("");
+  });
+
+  it("replaces javascript urls with  when javascript url begins with %20", () => {
```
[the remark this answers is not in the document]
AlexeyTuboltsev: ups

PR #4388, opened 2020-06-01
test pull 4 signup

glupiawymowka:
as per title

### DIFF
--- a/packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
+++ b/packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
@@ -9,13 +9,19 @@ describe("Signup", () => {
     cy.visit("/");
 
     cy.get(tid("Header-register")).click();
-    //how do I click on eye symbol?
 
     cy.get("#email").type(email);
     cy.get("#password").type(password);
     cy.get("#repeatPassword").type(password);
 
-    //why the tid?
+    cy.get(tid("wallet-selector-register-password")).find("[type='password']").should('be.visible') 
+    cy.get(tid("wallet-selector-register-password")).find("[type='text']").should('not.be.visible')
+
+    cy.get(tid("wallet-selector-register-password.adornment")).click();
+    
+    cy.get(tid("wallet-selector-register-password")).find("[type='password']").should('not.be.visible')
+    cy.get(tid("wallet-selector-register-password")).find("[type='text']").should('be.visible') 
+
     cy.get(tid("wallet-selector-register-tos")).click();
 
     cy.get(tid("wallet-selector-register-button")).click();
--- a/packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
+++ b/packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
@@ -1,26 +1,23 @@
 import { tid } from "../utils/index";
 import { cyan } from "color-name";
+import { generateRandomEmailAddress } from "../utils/userHelpers";
 
 describe("Signup", () => {
   it("should sign up with light wallet", () => {
-    cy.visit("https://localhost:9090");
+    const email = generateRandomEmailAddress();
+    const password = "myStrongPassword";
+    cy.visit("/");
 
     cy.get(tid("Header-register")).click();
-    //QUESTION - how do I toggle random input, to avoid email reuse?
     //how do I click on eye symbol?
 
-    cy.get("#email").type("email@test.com");
-    cy.get("#password").type("dupadupa");
-
-    //how do I make this line reuse the previously used password instead of typing?
-    cy.get("#repeatPassword").should("be.visible");
-    cy.get("#repeatPassword").type("dupadupa");
-    //this is wrong and I can't find the right command: cy.get("#repeatPassword").should("have.length", 8)
+    cy.get("#email").type(email);
+    cy.get("#password").type(password);
+    cy.get("#repeatPassword").type(password);
 
     //why the tid?
     cy.get(tid("wallet-selector-register-tos")).click();
 
-    cy.get(tid("wallet-selector-register-button")).should("be.enabled");
     cy.get(tid("wallet-selector-register-button")).click();
   });
 });
--- a/packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
+++ b/packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
@@ -1,0 +1,27 @@
+import { tid } from "../utils/index";
+import { cyan } from "color-name";
+
+describe("Signup", () => {
+  it("should sign up with light wallet", () => {
+    cy.visit("https://localhost:9090");
+
+    cy.get(tid("Header-register")).click();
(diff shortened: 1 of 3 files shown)


packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
+import { tid } from "../utils/index";
+import { cyan } from "color-name";
+
+describe("Signup", () => {
+  it("should sign up with light wallet", () => {
+    cy.visit("https://localhost:9090");
```
desfero, line 6: just a nitpick here: it's better to use just `/` over full URL `https://localhost:9090` given that the port may change as it's just a configuration setting
```
      cy.visit("/");
```

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    cy.get(tid("Header-register")).click();
+    //QUESTION - how do I toggle random input, to avoid email reuse?
```
desfero, line 9: at the moment we have a manually written utility function for that called `generateRandomEmailAddress`.

https://github.com/Neufund/platform-frontend/blob/d93bf5ebab4fdf38f8fdcbae69ca582cd64c09d6/packages/web/app/test-e2e/utils/userHelpers.ts#L26-L30

We just generate a random number from `0` to `1` and then convert that number base 32 string. It not fully unique but chance of collision is quite rare (given that we also reset the whole environment every day)

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
...
+    //how do I click on eye symbol?
```
desfero, line 10: Given that developers don't put `data-test-id` (tid) attributes by default on every element it's now a manual job to find that element and add a `data-test-id`.
The element is here:
https://github.com/Neufund/platform-frontend/blob/d93bf5ebab4fdf38f8fdcbae69ca582cd64c09d6/packages/design-system/src/components/inputs/TextField.tsx#L113-L117

That means you need to add a data-test-id for that attribute. Also given that it's inside `design-system` package you need to rebuild that package to see changes locally. Talk to me on slack for more information. The easiest way would be to explain in on a call.

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    cy.get("#email").type("email@test.com");
+    cy.get("#password").type("dupadupa");
+
+    //how do I make this line reuse the previously used password instead of typing?
+    cy.get("#repeatPassword").should("be.visible");
```
desfero, line 16: `cy.get("#repeatPassword").should("be.visible")` is not needed here given you also `type` line below because cypress always makes sure element is visible before invoking other commands

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
...
+    cy.get("#repeatPassword").type("dupadupa");
```
desfero, line 17: so to reuse password instead of typing the same string manually again we need to move it to the variable.

```
const password = "my-strong-password";
```
and then you can use it as  
``` 
cy.get("#repeatPassword").type(password);
```
see https://javascript.info/variables for some details :)

```

File: packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts [hunk 6]
```diff
@@ -0,0 +1,27 @@
...
+    //this is wrong and I can't find the right command: cy.get("#repeatPassword").should("have.length", 8)
```
desfero, line 18: not sure if there is a reason to check the length here as you typed that password above. To proper way would to logout after signup and then trying to sign in with the email and password to see whether it works.

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    //why the tid?
+    cy.get(tid("wallet-selector-register-tos")).click();
```
desfero, line 21: so `tid` (called often test ids) are a kind of abstraction that makes the process of finding elements easier. 
Often people use element selectors (for .e.g `#header .container .#tos`) but that's error-prone given that the structure in the DOM may change due to some kind of redesign.

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    cy.get(tid("wallet-selector-register-button")).should("be.enabled");
```
desfero, line 23: no reason to check if it's enabled given you also do `click` below. Cypress will make sure that button is enabled before clicking.

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,27 @@
...
+    cy.get(tid("wallet-selector-register-button")).click();
+  });
+});
+//should I group the test into smaller pieces and name them separately?
```
desfero, line 27: It really depends on whether the tests are complex or no. Often if you see that some parts can be reused then it's a good time to move them as a separate function (the same way `generateRandomEmailAddress` was moved to function to reuse the same functionality)

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,37 @@
...
+      .should("not.be.visible");
+
+    cy.get(tid("wallet-selector-register-password.adornment")).click();
+
+    cy.get(tid("wallet-selector-register-password"))
+      .find("[type='password']")
+      .should("not.be.visible");
+    cy.get(tid("wallet-selector-register-password"))
+      .find("[type='text']")
+      .should("be.visible");
```
desfero, line 31: we may try to simplify that further with
```
cy.get(tid("wallet-selector-register-password"))
      .find("[type='password']")
      .should("have.attr", "type", "password");
```
Same also on line `17`

packages/web/app/test-e2e/nikola/signup.spec.e2e.io.1.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    cy.get(tid("wallet-selector-register-tos")).click();
+
+    cy.get(tid("wallet-selector-register-button")).click();
+  });
```
desfero, line 36: here would be good to assert also that signing was successful (given that it can fail during network calls after you click `wallet-selector-register-button`). For that, we have a utility function `assertInvestorDashboard` (utils/assertions.ts) that will wait for an investor dashboard to appear on the screen.

Issue #4379, opened 2020-05-25
Add Unit tests for "Access Wallet" Module


Moejoe90:
Refer to the Epic for general guidelines on tests.

Moejoe90:
Merged

PR #4395, opened 2020-06-04
#4379 access wallet tests

AlexeyTuboltsev:


packages/sagas/src/sagaTestUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+      loopCount += 1;
+      if (
+        matchedReturn.result !== undefined &&
+        matchedReturn.result === ESagaTestDynamicProviderResult.ERROR
+      ) {
+        throw matchedReturn.payload;
+      } else if (
+        matchedReturn.result !== undefined &&
+        matchedReturn.result === ESagaTestDynamicProviderResult.RETURN
+      ) {
```
Moejoe90, line 52: why do we have a seperate if for this? is it for readability?

packages/sagas/package.json
```diff
@@ -16,7 +16,8 @@
     "start": "tsc --watch",
     "tsc": "tsc --noEmit",
     "tslint": "tslint --project ./tsconfig.json --format stylish",
-    "tslint:fix": "yarn tslint --fix"
+    "tslint:fix": "yarn tslint --fix",
```
AlexeyTuboltsev, line 19: adding "yarn browser:test" command to be able to run tests for new saga test utils

packages/web/app/modules/access-wallet/actions.spec.ts
```diff
@@ -0,0 +1,97 @@
+import { expect } from "chai";
+
+import { TestMessage } from "../../components/translatedMessages/messages";
+import { createMessage } from "../../components/translatedMessages/utils";
+import { actions } from "../actions";
+import { accessWalletReducer, initialState } from "./reducer";
+
+describe("access-wallet actions", () => {
```
Moejoe90, line 8: IMO we don't need action tests since we explicitly don't have any logic in the reducers.

packages/web/app/modules/access-wallet/sagas.spec.ts
```diff
@@ -0,0 +1,607 @@
...
+
+      const ledgerWalletConnectorMock = {} as any;
+
+      await expectSaga(ensureWalletConnection, {
+        web3Manager: web3ManagerMock,
+        walletStorage: walletStorage,
+        ledgerWalletConnector: ledgerWalletConnectorMock,
+      } as TGlobalDependencies)
+        .provide([
+          [matchers.call.fn(walletStorage.get), ledgerWalletMetadata],
```
Moejoe90, line 171: why do you need to provide the matchers?
It should already work without this

packages/web/app/modules/access-wallet/sagas.spec.ts
```diff
@@ -0,0 +1,607 @@
...
```
Moejoe90, line 196: same question as up

packages/web/app/modules/access-wallet/sagas.spec.ts
```diff
@@ -0,0 +1,607 @@
...
+        },
+      };
+
+      await expectSaga(accessWalletAndRunEffect, testEffect(), title, message, inputLabel)
+        .withState(state)
+        .provide([[matchers.call(connectWallet), undefined]])
+        .put(actions.accessWallet.showAccessWalletModal(title, message, inputLabel))
+        .call(connectWallet)
+        .put(actions.accessWallet.hideAccessWalletModal())
+        .returns("effect_return")
```
Moejoe90, line 577: where did you get this string?

packages/web/app/modules/access-wallet/sagas.spec.ts
```diff
@@ -0,0 +1,607 @@
...
+
+      const browserWalletConnectorMock = {} as any;
+
+      await expectSaga(ensureWalletConnection, {
+        web3Manager: web3ManagerMock,
+        walletStorage: walletStorage,
+        browserWalletConnector: browserWalletConnectorMock,
+      } as TGlobalDependencies)
+        .provide([
+          [matchers.call.fn(walletStorage.get), browserWalletMetadata],
```
Moejoe90, line 221: also here

packages/web/app/modules/access-wallet/sagas.spec.ts
```diff
@@ -0,0 +1,607 @@
...
+        ensureWalletConnection,
+        {
+          web3Manager: web3ManagerMock,
+          walletStorage: walletStorage,
+          lightWalletConnector: lightWalletConnectorMock,
+        } as TGlobalDependencies,
+        "strongpassword",
+      )
+        .provide([
+          [matchers.call.fn(walletStorage.get), lightWalletMetadata],
```
Moejoe90, line 252: here

packages/web/app/modules/access-wallet/sagas.spec.ts
```diff
@@ -0,0 +1,607 @@
...
+
+      const lightWalletConnectorMock = {} as any;
+
+      await expectSaga(ensureWalletConnection, {
+        web3Manager: web3ManagerMock,
+        walletStorage: walletStorage,
+        lightWalletConnector: lightWalletConnectorMock,
+      } as TGlobalDependencies)
+        .provide([
+          [matchers.call.fn(walletStorage.get), lightWalletMetadata],
```
Moejoe90, line 294: also here

packages/web/app/modules/access-wallet/sagas.spec.ts
```diff
@@ -0,0 +1,607 @@
...
+
+      const walletConnectConnectorMock = {} as any;
+
+      await expectSaga(ensureWalletConnection, {
+        web3Manager: web3ManagerMock,
+        walletStorage: walletStorage,
+        walletConnectConnector: walletConnectConnectorMock,
+      } as TGlobalDependencies)
+        .provide([
+          [matchers.call.fn(walletStorage.get), wcMetadata],
```
Moejoe90, line 319: also here

packages/sagas/src/sagaTestUtils.ts
AlexeyTuboltsev: well yeah. I tried to avoid more nested ifs here

packages/web/app/modules/access-wallet/actions.spec.ts
AlexeyTuboltsev: hmmm. Tmere's always a real danger to screw smth up and start mutating in the reducer. Those tests should in theory check for this.

packages/web/app/modules/access-wallet/sagas.spec.ts
AlexeyTuboltsev: but the saga needs the metadata and the wallet objet to compare the addresses. It won't get to this check if they are not provided, it will throw earlier

AlexeyTuboltsev: and they both come from external sources, - wallet storage and another saga. So a "withState()" won't be enough

packages/web/app/modules/access-wallet/sagas.spec.ts
AlexeyTuboltsev: see answer above. I need the data from "external" sources to get to this point. "withState()" or smth like this won't be enough. 
the stubbed "connectLedger" that I provide here, in turn needs tons of stuff from context that's hard to mock. So if you mean this is too coupled, I'd say it's still the best way to do it, the others bring us to even much more coupling

packages/web/app/modules/access-wallet/sagas.spec.ts
AlexeyTuboltsev: it;s just a test value. To test if the saga really goes all way through and returns. I can call it "return_value" or anything else

packages/web/app/modules/access-wallet/sagas.spec.ts
AlexeyTuboltsev: see above

packages/web/app/modules/access-wallet/sagas.spec.ts
AlexeyTuboltsev: see above

PR #4401, opened 2020-06-08
Adds Investment Module tests

Moejoe90:


packages/web/app/modules/investment-flow/sagas.ts
```diff
@@ -67,12 +67,12 @@ function* processCurrencyValue(
...
   yield put(actions.investmentFlow.setIsInputValidated(false));
-  yield computeAndSetCurrencies(value, curr);
+  yield call(computeAndSetCurrencies, value, curr);
   // dispatch in order to debounce, instead of calling directly
   yield put(actions.investmentFlow.validateInputs());
 }
 
-function* computeAndSetCurrencies(value: string, currency: ECurrency): any {
+export function* computeAndSetCurrencies(value: string, currency: ECurrency): any {
   const state: TAppGlobalState = yield select();
```
AlexeyTuboltsev, line 76: maybe use the opportunity to clean this up? 
`const type = selectInvestmentType(s);` -> `const type = yield* select(selectInvestmentType);` etc
if you have time

packages/web/app/modules/investment-flow/sagas.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+            wallets: [],
+          },
+        })
+        .provide([[matchers.call.fn(computeAndSetCurrencies), undefined]])
+        .not.call(computeAndSetCurrencies)
+        .run();
+    });
+  });
+
+  describe("recalculateCurrencies", async function(): Promise<void> {
```
AlexeyTuboltsev, line 71: typo. should be `computeAndSetCurrencies`

packages/web/app/modules/investment-flow/sagas.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+            euroValueUlps: "",
+            investmentType: EInvestmentType.NEur,
+            isValidatedInput: false,
+            wallets: [],
+          },
+        })
+        .provide([[matchers.call.fn(computeAndSetCurrencies), undefined]])
+        .not.call(computeAndSetCurrencies)
+        .run();
+    });
```
AlexeyTuboltsev, line 68: if `type === undefined` (line 406) check is to prevent some bugs and not only to silence the TS, I'd add a test for it.

packages/web/app/modules/investment-flow/sagas.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+          {
+            investmentFlow: {
+              ethValueUlps: "99",
+              etoId: "0x2754523CE7C78FeD60F742b0Bc8A8F9fa323bB4C",
+              euroValueUlps: "",
+              investmentType: EInvestmentType.Eth,
+              isValidatedInput: false,
+              wallets: [],
+            },
+            tokenPrice: { tokenPriceData: { etherPriceEur: "0", neuPriceEur: "0" } as any } as any,
```
AlexeyTuboltsev, line 88: so many any :))
the last one should be enough. 

especially lines 221-223 are an any orgy  :D
Not blocking though

packages/web/app/modules/investment-flow/sagas.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+          },
+        )
+        .not.put(actions.investmentFlow.setIsInputValidated(false))
+        .run();
+    });
+    it("should set values as empty when value is not given", async () => {
```
AlexeyTuboltsev, line 173: something's wrong with the description

packages/web/app/modules/investment-flow/sagas.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+          },
+        )
+        .call(
+          computeAndSetCurrencies,
+          new BigNumber("100000000000").minus("600000").toString(),
+          ECurrency.ETH,
+        )
+        .put(actions.investmentFlow.validateInputs())
+        .run();
+    });
```
AlexeyTuboltsev, line 299: I'd add tests for ICBM stuff too. I know the flow is essentially the same as what's already covered and it would be a simple copy-paste, but it would help people not forget about those options in case of refactoring. 
When I was doing the investment flow refactoring (the PR that was dropped),  I was really struggling with so many different variants of everything that have to be taken into account. Tests that at least give an idea what those options are would have come very handy then.

packages/web/app/modules/investment-flow/sagas.spec.ts
```diff
@@ -0,0 +1,301 @@
...
+            } as any,
+            gas: { gasPrice: { standard: "1" } as any } as any,
+            txSender: { txDetails: { gasPrice: "1", gas: "1" } as any } as any,
+          },
+        )
+        .call(computeAndSetCurrencies, "2", ECurrency.ETH)
+        .put(actions.investmentFlow.validateInputs())
+        .run();
+    });
+    it("should invest entire balance for Ether with gas", async () => {
```
AlexeyTuboltsev, line 261: maybe use this place to add a comment to explain what does it mean? Why do we check for balanceWithoutGas etc. (it's about the gas stipend, right?)

packages/web/app/modules/investment-flow/sagas.spec.ts
Moejoe90: its a test, I don't want to add the full state if its not gonna be used

packages/web/app/modules/investment-flow/sagas.ts
Moejoe90: I purposly didn't want to touch any code in this PR

packages/web/app/modules/investment-flow/sagas.spec.ts
Moejoe90: good catch

PR #4415, opened 2020-06-10
At/#4387 signer modal fix

AlexeyTuboltsev:


andywalton:
Hi @glupiawymowka - when you test this PR, Lexey and myself had a Slack conversation regarding the short sentence below the Modal title, this sentence should read:

By default (MetaMask, Ledger Nano):
`Please sign on <wallet_name> to confirm this action.`

WalletConnect:
`Please sign on <wallet_name> to confirm this action.` as per the default but sometimes the wallet details might be unknown, ie. we don't know the wallet name or have it's wallet icon. If this is the situation, the text will read:

`Please sign on your WalletConnect connected wallet to confirm this action.`

For the Neufund Mobile App:
`Please sign on the Neufund Mobile App to confirm this action.`

Please ask myself or @AlexeyTuboltsev regarding this.

packages/web/app/components/shared/WarningAlert.module.scss
```diff
@@ -8,7 +8,7 @@
 
   .icon {
     @include rect(20px);
-
+    align-self: baseline;
```
Moejoe90, line 11: we need to ve carful here this is general component this might break the view where the component is used in pther places

packages/web/app/components/modals/wallet-access/AccessWalletModal.module.scss
```diff
@@ -1,15 +1,83 @@
-.info {
-  color: $gray6;
+$modal-border-radius: 4px;
+
+.main {
+  max-width: px-to-rem(650) !important; // sass-lint:disable-line no-important //reactstrap!!!
```
Moejoe90, line 4: lets make clear comments reactstrap looks more like a rant than a comment to me

AlexeyTuboltsev: done

packages/web/app/components/shared/WarningAlert.module.scss
AlexeyTuboltsev: this is how it should look, so in fact  this fixes a previous bug

Moejoe90: The main point is that it works in all the previous instances where this was used

Issue #4410, opened 2020-06-10
[Wallet Connect] Page improvements


andywalton:
**Description** 
This ticket improves the current WalletConnect page on the platform by adding more content to the page and changing the experience slightly to make it more usable.   

The current user flow:  
1. As a user I can navigate to our WalletConnect page.
2. This currently displays the WalletConnect modal automatically - which is good because you can simply scan the QR code and connect to the platform but bad because it's a modal automatically opening and the page behind it, you're unsure what content it's displaying.
3. If you close this modal, we show a page with very little information and an error. The user also sees this for the first time.

This ticket will focus on improving point 3. and making a change to point 2.

**Designs**
![Screenshot 2020-06-10 10.19.13.png](https://images.zenhubusercontent.com/5c5861910b133e639c2b7056/1359ad00-a08f-4a05-99ba-329477523f73)

You can view the screen designs in Zeplin here: https://zpl.io/angAmrv

See the designs for the relevant icon and components used (which should link through to the design system in Zeplin also).

**Tasks**
Please implement the following changes to the WalletConnect user flow:

- [x] As a user I can navigate to our WalletConnect page but the WalletConnect modal doesn't display immediately.

- [x] Instead, I now see a page that displays a little more information about connecting to Neufund using WalletConnect. The `Learn more` should open a link in a new tab to `https://walletconnect.org/`. 

The primary call to action `Connect using WalletConnect` should open the WalletConnect QR code modal we've already implimented.

- [x] The user can then scan a QR code using a WalletConnect compatible wallet to access the platform. If this modal is closed or there is an error, a user will then see the WalletConnect landing page but it with an error showing that asks them to retry as there was an issue connecting to their wallet.

- [x] While this feature is in beta, please link the `Customer Care` link to `https://support.neufund.org/support/tickets/new`

- [x] Take note of the responsive changes for our smallest mobile breakpoint (screens <=540px wide I think). See the mobile screen designs.

- [x] The header navbar should show both our `Log in` and `Get started` buttons.

**Technical requirements**
There are no technical changes other initiating the WalletConnect modal only on pressing the CTA visible on at the WalletConnect URL.

**Acceptance**
@glupiawymowka , as per the tasks.

PR #4424, opened 2020-06-15
#4410 walletConnect landing page redesign

AlexeyTuboltsev:


packages/web/intl/locales/en-en.json
```diff
@@ -1464,9 +1465,13 @@
   "voting.failed-to-load-proposal": "Failed to load proposal details.",
   "voting.unknown-proposal": "Unknown proposal.",
   "voting.unsupported-proposal-state": "Unsupported proposal state.",
+  "wallet-connect.beta-testing-message": "This feature is in beta testing. Please share your feedback with our  <a href={href} target=\"_blank\">Customer Care</a>.",
```
Moejoe90, line 1468: I'm not sure if its the best idea to put the full links here as we already have this as a react component

AlexeyTuboltsev: yeah I know about that component. The problem here is that you have to split the wording for it in multiple `FormattedMessages`, sometimes in a very awkward way, like this: "Please visit our "--"customer care"--" if you have problems".  And don't forget the white spaces around all of it

Even worse, if we ever will have another language, it's very likely that the word order will change. That means you can't split this in a way to be really cross-language, which kinda defeats the whole idea of having translations.

Moejoe90: but on the other hand you have maybe unsafe html in the translation file.
Its also very hard to keep track of what you have there. 

So you end up forgetting to add rel="noopener noreferrer for example which is the case now. btw

For safety please use `ExternalLink`

AlexeyTuboltsev: hm. once again, this makes it unusable as a translation.
done

AlexeyTuboltsev: done

andywalton:
@AlexeyTuboltsev @glupiawymowka @Moejoe90 

I have two small changes for this PR please:

1. Update the alert icon please. (Will send share with you @AlexeyTuboltsev)

2. Update the alert error text to: `An error occurred while communicating to your remote wallet. Please retry.`

Please also make sure the colour of the text is: `#FF5B5B`

Screenshot 2020-06-17 12 06 18

PR #4429, opened 2020-06-16
removes clearEmailServer from e2e tests

Moejoe90:
#p3


packages/shared-modules/src/tests/wrapperFetch.ts
```diff
@@ -7,5 +7,5 @@ export const wrappedFetch = async (input: RequestInfo, init?: RequestInit): Prom
     if (response.ok) {
       return response;
     }
-    throw new Error(response.statusText);
+    throw new Error(response.statusText + response.status);
```
AlexeyTuboltsev, line 10: could you wrap it into a string and add a separator, like `'${response.statusText}, status is ${response.status}'`, for the sake of readability?

Moejoe90: good idea

PR #4431, opened 2020-06-16
Fix memory leaks in unit tests

desfero:
**Technical notes**
Not really sure that the root cause of memory leaks on CI as it's quite a subtle leak with small memory overhead that's also get cleared out when the tests are finished.

Also, i did some changes to the eslint config to catch missing `await`s


packages/wallet/app/modules/eth/lib/EthSecureEnclave.spec.ts
```diff
@@ -17,6 +17,10 @@ import { getInternetCredentials } from "./__mocks__/react-native-keychain";
 import { isMnemonic } from "./utils";
 
 describe("EthSecureEnclave (with SecureStorage)", () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
```
Moejoe90, line 21: was this the reason?

desfero: yeah, under the hood memory-cache npm plugin set's timeout in each test run when something get's put onto the cache. Tests run is completed but still, the timeout is there causing some leaks and slow down of tests

Issue #4427, opened 2020-06-16
[Message signers] - new message signer broken for lightwallet


glupiawymowka:
**Description**
We started to improve and fix the message signers for all the wallets. While the message signers like email or recovery phrase access are styled well, the transactional ones, like ETH send, broke for light wallet. 

**env**
.io

**steps to reproduce**|
1. Register as a light wallet user 
2. Try the email change or recovery phrase modals, they should be styled correctly
3. Send some ETH to your light wallet account, so you can use it for transactions 
4. Try to send ETH from light wallet to another account

**Actual results**
- [x] All the transactional signers like nEUR redeem, investment, ETH send are misaligned
- [x] There is an unnecessary copy repetition in the second paragraph 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/513b68aa-cd4f-4210-9aad-a781de8f2fea) 

**Expected results**
- [x] Icon and the copy are centered
- [x] The second "please..." is deleted 
- [ ] The styling fits the non-transactional signers
![image](https://user-images.githubusercontent.com/18315589/84757136-0cb71f80-afc4-11ea-8668-d08127d113cb.png)

PR #4436, opened 2020-06-18
#4427 fix signer modal

AlexeyTuboltsev:


packages/web/app/components/modals/tx-sender/TxSender/TxSenderBody.tsx
```diff
@@ -39,10 +41,9 @@ const TxSenderBody: React.FunctionComponent<TxSenderProps> = ({
 
     case ETxSenderState.ACCESSING_WALLET:
       return (
-        <AccessWallet
-          title={<FormattedMessage id="modals.tx-sender.confirm-title" />}
-          message={<FormattedMessage id="modals.tx-sender.confirm-description" />}
-        />
+        <div className={styles.accessWalletWrapper}>
```
Moejoe90, line 44: do you need `styles.accessWalletWrapper` here and in `AccessWallet.tsx` ?

packages/web/app/components/shared/hooks/useLogger.ts
```diff
@@ -8,7 +8,7 @@ const useLogger = () => {
   const container = React.useContext(ContainerContext);
 
   // For storybook return noop logger as container is not connected with stories
-  if (process.env.STORYBOOK_RUN === "1") {
+  if (process.env.STORYBOOK_RUN === "1" || process.env.NODE_ENV === "development") {
```
Moejoe90, line 11: why did you disable this during development?

packages/web/app/components/modals/tx-sender/TxSender/TxSenderBody.tsx
AlexeyTuboltsev: yeah. This is a workaround necessary because we have two modals now - an old one with a yellow line on the top, and the new one with rounded corners. I need to make the content look good in both, so I decided to do it in a unified way. This wrapper can be removed once we have a single modal on the platform

packages/web/app/components/shared/hooks/useLogger.ts
AlexeyTuboltsev: we need to be able to run storybook during dev too and we don't need logger at this time.

packages/web/app/modules/access-wallet/reducer.ts
```diff
@@ -31,10 +31,7 @@ export const accessWalletReducer: AppReducer<ISignMessageModalState> = (
         inputLabel: action.payload.inputLabel,
       };
     case actions.accessWallet.hideAccessWalletModal.getType():
-      return {
-        ...state,
-        isModalOpen: false,
```
Moejoe90, line 36: :scream:

PR #4438, opened 2020-06-22
Refactor logger interface

desfero:
**Technical notes**
- `verbose` level was removed in favor of using `debug` 
- debug/info/warn/error method arguments were normalized 

The main reason for a change was to catch invalid logs during compile time as for e.g.:
`logger.error('Error while trying to delete nominee request with nominee id ${nomineeId}');` is now valid as sentry requires `Error` object to capture properly exception.


packages/shared-modules/src/modules/core/lib/logger/ILogger.ts
```diff
@@ -1,14 +1,13 @@
...
-  info(...args: LogArg[]): void;
-  verbose(...args: LogArg[]): void;
-  debug(...args: LogArg[]): void;
-  warn(...args: ErrorArgs[]): void;
-  error(...args: ErrorArgs[]): void;
-  fatal(message: string, error: Error, data?: object): void;
+  info(message: string, data?: object): void;
+  debug(message: string, data?: object): void;
+  warn(message: string, data?: object): void;
+  error(error: Error, message?: string, data?: object): void;
```
Moejoe90, line 9: :+1:

PR #4460, opened 2020-07-01
Update ESLint

desfero:
**Technical notes**
ESLint did a really good progress recently with typescript linting support. After the update we got more than 300 linting errors related to invalid types/any's (a couple of them caught a real bug in the code). 

Not all violations were fixed, a lot of any related violations in spec were just marked as ignored given its quite time consuming to fix all of them at once.


packages/wallet/.eslintignore
```diff
@@ -1,4 +1,4 @@
-node_modules
+.eslintrc.js
```
arjunu, line 1: it doesn't ignore itself by default?

packages/shared-utils/src/utils/types.ts
```diff
@@ -192,3 +192,11 @@ export type UnionToIntersection<U> = (U extends any
...
   ? I
   : never;
+
+/**
+ * Specifies an object with unknown keys
+ *
+ * Useful when you don't care about exact props passed to the component.
+ * @note Don't use `{}` as a type. `{}` actually means "any non-nullish value".
+ */
+export type UnknownObject = Record<string, unknown>;
```
arjunu, line 202: 👍

packages/wallet/app/components/shared/forms/fields/Form.tsx
```diff
@@ -22,7 +25,7 @@ const Form = <Values extends {}>({
     enableReinitialize
     validateOnMount
   >
-    {formikProps => (typeof children === "function" ? children(formikProps) : children)}
+    {formikProps => (isFunction(children) ? children(formikProps) : children)}
```
arjunu, line 28: Formik uses the same :D https://github.com/formik/formik/blob/d2637207791a74489c5d57532a0c23dbf0a70b22/packages/formik/src/utils.ts#L12

packages/wallet/app/modules/eth/lib/EthSecureEnclave.ts
```diff
@@ -192,7 +192,7 @@ class EthSecureEnclave {
     try {
       const hdNode = utils.HDNode.fromMnemonic(secret).derivePath(derivationPath);
       return this.addSecret(hdNode.privateKey);
-    } catch (e) {
```
arjunu, line 195: don't we want the original exception in the message?

packages/wallet/app/modules/wallet-connect/lib/WalletConnectAdapter.ts
```diff
@@ -108,6 +114,8 @@ class WalletConnectAdapter extends EventEmitter2 {
         result,
       });
     } catch (error) {
+      assertError(error);
```
arjunu, line 117: when would this not be an instance of error?

packages/wallet/app/modules/eth/lib/EthSecureEnclave.ts
arjunu: same question for other places

packages/wallet/.eslintignore
desfero: that just changed in a recent update

packages/wallet/app/modules/wallet-connect/lib/WalletConnectAdapter.ts
desfero: i would say it depends. You can `throw` anything you want, from number/string to even Promise (that's btw exactly how react suspense is implemented)

packages/wallet/app/modules/eth/lib/EthSecureEnclave.ts
desfero: good point, fixed

Issue #4450, opened 2020-06-23
Landing page changes pt2


pstroinski:
**Background**
Last fixes to the landing page.

- [x] Add "Stay tuned" section as on design - the `Sign up` should open our Sign up view.
- [x] Remove the current "Register" button 
- [ ] Remove the "Coming soon" card.

![image](https://user-images.githubusercontent.com/27957897/85426016-38b14280-b57a-11ea-91c3-55596c1674c8.png)
- [x] remove unused assets for mocked etos

**Designs**
See designs here: https://www.figma.com/file/rjk3UJLvBt8c0I7zyoOJIN/Neufund_website_transition2020_landing?node-id=35%3A0

![Neufund_platform_temp_landing_02](https://user-images.githubusercontent.com/609657/85551740-94370b00-b622-11ea-98b3-bc57d4da0ef2.png)

**Acceptance criteria**
Please make sure that the "sign up" button down there won't cause the same registration loop that we have currently. How to check it: 
- Go to the landing page
- Press "sign up"
- Go back with the browser "<-" button 

- [x] The landing page reopens correctly once the mentioned steps are done

PR #4465, opened 2020-07-02
#4450 landing page fixes #2

AlexeyTuboltsev:
adds a "stay tuned" block to the bottom of the landing page
removes unused assets
fixes the redirection loop in login route


packages/web/app/modules/routing/actions.ts
```diff
@@ -53,6 +53,11 @@ export const routingActions = {
   goToIssuerLightWalletRegister: () => push(appRoutes.registerIssuerWithLightWallet),
   goToNomineeLightWalletRegister: () => push(appRoutes.registerNomineeWithLightWallet),
 
+  redirectToRegisterBrowserWallet: () => replace(appRoutes.registerWithBrowserWallet),
```
desfero, line 56: +1

packages/web/app/components/landing/Landing.module.scss
```diff
@@ -73,3 +73,21 @@
 
   padding-top: px-to-rem(24);
 }
+
+.stay-tuned {
+  display: flex;
+  flex-flow: column;
+  align-items: center;
+  width: 100%;
```
desfero, line 81: Is there a reason to have `width: 100%`? Isn't `display: flex` a block element?

AlexeyTuboltsev: yeah, it's a leftover
removed

Issue #4223, opened 2020-03-31
Refresh JWT token before it expires 


desfero:
**Background**
As a part of #4220 we moved some code from auth and jwt module to implement account creation. What is still missing is a refresh JWT logic we have in the web app. As a part of this ticket please move refresh code to `shared-modules` and connect it with the mobile app.

**Designs**
not needed

**Acceptance criteria**
- [x] Given that I'm logged in and the JWT token is close to expiring it will automatically refresh the token with the same permissions so the app is still usable

**Technical notes**

**Testing notes**
As for testing, I think the easiest approach is to log in and go to the wallet page and see if after around 10 minutes (for `io`) the transactions history list can be reloaded without an error. If there is an error that means JWT token expired and was not properly refreshed.

Issue #4426, opened 2020-06-15
Research a better way to mark modules as dependent on other modules


desfero:
**Background**
Currently, to mark the module as a dependency we just return an array from `setupMyAwesomeModule` function.

https://github.com/Neufund/platform-frontend/blob/590b829fea978298aaf1e4cce8b584bab107e9ac/packages/wallet/app/modules/contracts/module.ts#L13-L26

This is quite handy:
- It makes sure that the module dependencies are always initialized even if they were removed from the root module
- It's quite easy to generate typings with proper actions/state as we have full type definitions of all module state/actions/symbols and dependencies state/actions/symbol
- It's easier to write integration module tests as we have already access to the desired system shape and we know what can be mocked and what is not a part of a module as a whole.

But unfortunately, it's not perfect (yet!).

The problem starts to appear when the module needs some parameters (for. e.g some actions to dispatch on success / DI symbols to use / etc).

https://github.com/Neufund/platform-frontend/blob/590b829fea978298aaf1e4cce8b584bab107e9ac/packages/shared-modules/src/modules/tx-history/module.ts#L25-L36

In the code above `setupTokenPriceModule` requires an action that will trigger token price refresh logic and the proper action is provided at the moment in the root module 

https://github.com/Neufund/platform-frontend/blob/590b829fea978298aaf1e4cce8b584bab107e9ac/packages/web/app/store.ts#L76-L78

At first glance, it may not look so scary. But in reality, it is and may end up as huge, hard to catch, regressions in the future. In case someone changes the order if root modules and `setupTokenPrice` module lands after `setupTxHistoryModule` we end up with a token-price module initialized without proper `refreshOnAction`. Here the damage won't be a deal-breaker but imagine the param called `useSecretStorage?: boolean`.

**Technical notes**

**Testing notes**

PR #4466, opened 2020-07-06
#4223 Refresh JWT token 

desfero:
**Technical notes**
This pr
- Moves JWT refreshing code to shared-modules
- Makes sure mobile app refreshes jwt token when needed
- Adds tests for `handleJwtRefresh` logic
- Enables `esModuleInterop` for `shared-utils` and `web` tests so we slowly can get rid of `import * as something from 'package'` as it's not in sync spec

**Testing notes**
Run #p3

IOS: https://install.appcenter.ms/orgs/Neufund/apps/Neufund-Wallet-MVP-iOS/releases/18
Android: https://install.appcenter.ms/orgs/Neufund/apps/Neufund-Wallet-MVP/releases/17


packages/sagas/src/sagasUtils.ts
```diff
@@ -105,7 +105,7 @@ export function* neuTakeUntil(
 /**
  * Starts saga on `startAction`, cancels on `stopAction`, loops...
  */
-export function* neuTakeLatestUntil<
+export function* takeLatestUntil<
```
desfero, line 108: let's not use `neu*` prefix for sagas that do not inject `TGlobalDependencies` to distinguish the approach

packages/shared-modules/src/modules/auth/user/sagas.spec.ts
```diff
@@ -29,10 +29,15 @@ describe("Auth - User - Integration Test", () => {
...
-      const { expectSaga, container } = bootstrapModule([
-        setupCoreModule({ backendRootUrl: BACKEND_BASE_URL }),
-        setupAuthModule({ jwtStorageSymbol, ethManagerSymbol }),
-      ]);
+      const { expectSaga, container } = bootstrapModule(
+        setupAuthModule({
+          backendRootUrl: BACKEND_BASE_URL,
+          jwtStorageSymbol,
+          ethManagerSymbol,
+          jwtTimingThreshold: 100,
```
Moejoe90, line 37: non-blocking why not use a test constant for all the values that equal `100`

packages/shared-modules/src/modules/auth/user/sagas.ts
```diff
@@ -135,11 +135,11 @@ export function* resetUser(): SagaGenerator<void> {
...
 
-  yield put(userActions.reset());
+  yield put(userPrivateActions.reset());
 
   logger.info("User has been cleared out");
 }
 
 export function* authUserSagas(): SagaGenerator<void> {
-  yield takeEvery([userActions.setUser, userActions.reset], setLoggerUser);
+  yield takeEvery([userActions.setUser, userPrivateActions.reset], setLoggerUser);
```
Moejoe90, line 144: can you give an explenation on what `userPrivateActions` since it seems that `reset` is connected to main event saga? 

Also I see a circular dependancy because `setLoggerUser` is activated by `userPrivateActions.reset` and when run calls `userPrivateActions.reset`

packages/sagas/src/sagasUtils.ts
desfero: Will add that to your docs pr

packages/web/app/components/modals/tx-sender/redeem/BankTransferRedeemInit.tsx
```diff
@@ -57,6 +57,17 @@ export interface IReedemData {
   amount: string | undefined;
 }
 
+const formatEuroValueToString = (value: string) =>
```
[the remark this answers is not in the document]
desfero: it was not moved by @arjunu with formatters but anyway not sure we should use this one in components given we have components for that. That's why I moved it to just single place where it's used.

packages/web/app/store.ts
```diff
@@ -72,12 +73,17 @@ export const setupAppModule = ({ history, config, container }: TAppModuleConfig)
...
   };
 
+  const authConfig = {
+    backendRootUrl: config.backendRoot.url,
+    jwtStorageSymbol: symbols.jwtStorage,
+    ethManagerSymbol: symbols.web3Manager,
+    jwtTimingThreshold: AUTH_JWT_TIMING_THRESHOLD,
+    jwtRefreshThreshold: AUTH_TOKEN_REFRESH_THRESHOLD,
+  };
+
```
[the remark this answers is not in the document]
desfero: I think we gonna need some refactoring with #4426 so let's see how it evolves

packages/shared-modules/src/modules/bookbuilding/sagas.ts
```diff
@@ -305,8 +304,8 @@ export function setupBookbuildingSagas(config: TSetupSagasConfig): () => SagaGen
 
   return function* bookBuildingSagas(): Generator<any, any, any> {
     yield fork(neuTakeEvery, bookbuildingActions.loadBookBuildingStats, loadBookBuildingStats);
-    yield fork(
-      neuTakeLatestUntil,
+    yield;
```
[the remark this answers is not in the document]
desfero: good point. A proper typing `SagaGenerator<void>` will catch that error

packages/shared-modules/src/modules/auth/user/actions.ts
```diff
@@ -2,7 +2,10 @@ import { createActionFactory } from "@neufund/shared-utils";
 
 import { IUser } from "../module";
 
+export const userPrivateActions = {
```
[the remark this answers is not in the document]
desfero: will do with documentation PR

packages/shared-modules/src/modules/auth/user/sagas.ts
desfero: So `userPrivateActions` are internal to the module and should not be exposed by `authModuleApi`. We also use the same pattern for `privateSymbols` in a couple of places.

I don't see any circular dependency with `setLoggedUsere`, where it's exactly?

packages/shared-modules/src/modules/auth/user/sagas.spec.ts
desfero: Did some refactoring to move the reusable part to a separate functioin

packages/shared-modules/src/modules/auth/user/sagas.ts
Moejoe90: My bad I missed the fact that `setLoggerUser` was using `logger` and not `userActions`

PR #4468, opened 2020-07-06
Fix mismatch between dependencies

desfero:
**Technical notes**
At the moment our imports between packages are really messy because no lint rules enforce each import to be in `package.json` of a specific package. Given that we 
- import `@neufund/web` in the `design-system` package
- `web` app does not implicitly say it depends on `@neufund/shared-modules`
- react import has slipped through to `shared-utils`
- a lot of packages are missing in `package.json`
- hard to catch invalid package being used during code review (Eto module migration pr almost added `connected-react-router` to shared-modules).

This pr fixes all issues and turns on lint rules for tslint/eslint to catch that before it reaches `master`


packages/shared-utils/tslint.json
```diff
@@ -67,7 +67,7 @@
       }
     ],
     "no-commented-code": true,
-    "no-submodule-imports": false,
+    "no-submodule-imports": [true, "lodash"],
```
arjunu, line 70: is this to catch sub module imports from our own packages? like `@neufund/web/x`? But we also have other sub module node imports like `redux-saga/effects`

packages/design-system/package.json
```diff
@@ -77,11 +78,13 @@
     "react-intl": "^2.4.0",
     "reactstrap": "^8.4.0",
     "react-intl-phraseapp": "^0.2.4",
-    "react-virtualized-select": "^3.1.3",
     "yup": "^0.26.7"
   },
   "dependencies": {
     "react-table": "^7.0.0-rc.16",
+    "react-select": "^1.2.1",
```
arjunu, line 85: `react-select` is included with web, so it can be a peer dependency here

packages/shared-utils/tslint.json
desfero: yeah, each submodule import should be specified manually, you can check other `no-submodule-imports` overwrites

packages/design-system/package.json
desfero: moved to peerDependencies

PR #4470, opened 2020-07-09
Governance Overview page

arjunu:
To be merged after https://github.com/Neufund/platform-frontend/pull/4455


packages/web/app/modules/governance/sagas.ts
```diff
@@ -82,6 +83,7 @@ export function* loadGeneralInformationView(): Generator<any, void, any> {
 
   let eto = yield select(selectIssuerEto);
   if (!eto) {
+    yield neuCall(loadIssuerEto);
```
Moejoe90, line 86: was this a bug?

arjunu: yes, when we load the page directly instead of coming from Dashboard, it should load ETO

PR #4472, opened 2020-07-13
Governance Responsiveness

arjunu:
To be merged after https://github.com/Neufund/platform-frontend/pull/4470


arjunu:
> while the General Information page fits the screen well, the Overview doesn't fit the screen and the user needs to scroll horizontally.
> 
> reproducible with Metamask app (iPhone XS) and dev tools (any device chosen).
> fixture used - Issuer_payouts
> 
> image

Fixed now. It was supposed to be two columns in small screens.

packages/web/app/components/governance/Governance/GeneralInformation/GeneralInformation.module.scss
```diff
@@ -1,21 +1,39 @@
 .title-row {
   display: flex;
   justify-content: space-between;
+  margin-bottom: px-to-rem(20x);
```
AlexeyTuboltsev, line 4: 20x looks like a typo?

packages/web/app/components/appRoutes.ts
```diff
@@ -102,3 +104,5 @@ export const appRoutes = {
    */
   etoPublicViewLegacyRoute: "/eto/view/:previewCode",
 } as const;
+
+export type TAppRoute = Values<typeof appRoutes>;
```
AlexeyTuboltsev, line 108: 👍

packages/web/app/components/governance/Governance/GovernancePageSelect.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+type TOptionType = { label: string; value: string };
+
+type TProps = {
+  value: TAppRoute;
+  onChange: (to: string) => void;
+};
+
+export const GovernancePageSelect: React.FunctionComponent<TProps> = props => (
+  <VirtualizedSelect
```
AlexeyTuboltsev, line 17: what are the benefits of virtualizedSelect here? Why not a simple select? There are only 6 options, not 6000 where virtualizedSelect makes sense
Or maybe I miss something

packages/web/app/components/governance/Governance/GovernancePageSelect.tsx
```diff
@@ -0,0 +1,34 @@
...
+      page =>
+        ({
+          label: page.label,
+          value: page.to as string,
+        } as TOptionType),
+    )}
+    value={props.value}
+    searchable={false}
+    clearable={false}
+    onChange={(selectedOption: unknown) => {
```
AlexeyTuboltsev, line 29: is selectedOption really unknown here? 
there are many type castings in this file, but all arguments are actually pretty well known. Are castings really necessary?

AlexeyTuboltsev: btw as TSC point out, selectedOption may be null. Need to add a check for this.

AlexeyTuboltsev: hm, the typings of virtualizedSelect don't seem to be correct at all. I strongly suggest to get rid of it and use a simple html select. A check for null would be necessary there as well

packages/web/app/components/governance/Governance/GeneralInformation/GeneralInformation.module.scss
arjunu: nice catch, fixed

packages/web/app/components/governance/Governance/GovernancePageSelect.tsx
arjunu: as per discussion, put it in an unsafe file for now with the type castings. And here is the issue to create a new component: https://github.com/Neufund/platform-frontend/issues/4535

PR #4486, opened 2020-07-16
Unlock Euro Wallet - Hotfix

Moejoe90:


packages/web/app/components/modals/tx-sender/unlock-wallet-flow/UnlockWalletTransactionDetails.tsx
```diff
@@ -1,5 +1,6 @@
 import { walletApi } from "@neufund/shared-modules";
```
Moejoe90, line 1: This file is different than the PR against Master

arjunu: NEU can also be non zero decimals, but it's ok

PR #4490, opened 2020-07-20
adds android wallet signing

rudolfix:
adds gradle configuration for production signing. see README. also commits encrypted production keystore in PKCS12 format


packages/wallet/android/app/build.gradle
```diff
@@ -158,15 +158,22 @@ android {
...
     buildTypes {
         debug {
+            // applicationIdSuffix ".debug"
             signingConfig signingConfigs.debug
         }
         release {
-            // Caution! In production, you need to generate your own keystore file.
-            // see https://facebook.github.io/react-native/docs/signed-apk-android.
-            signingConfig signingConfigs.debug
+            // applicationIdSuffix ".debug"
```
desfero, line 174: We probably don't need that `code-comment` here as it's not related to production build varian

packages/wallet/README.md
```diff
@@ -111,3 +111,37 @@ yarn e2e:ios-release
...
+#### Android Debug Release
+
+Apk is signed using `debug` keystore which is committed to repository with cleantext passwords. The application id is `org.neufund.mobile` (TODO: we should add `.debug` postfix soon)
+
+#### Android Production Release
+
+Android release uses `aab` (bundle) format. In that scheme, the final signing key is maintained by Google in Play Store and bundle is signed with upload key. The application id is `org.neufund.mobile`. To generate and sign you should follow the steps.
+1. Make sure you have Java JDK and Android build toolchain available with SDK level 28.
+2. Valid `.env` file must be present containing valid production configuration (like Firebase service access)
+3. You need keystore in `wallet/android/app/production.keystore.p12`
```
desfero, line 126: Seems that steps are duplicated  (1,2,3,4,3,4,5). 

Would be also good to keep consistency with env variables names so they are always prefixed with `NF_`

rudolfix: good to know that `README` is also linted :)

PR #4494, opened 2020-07-21
Small tweaks to remove react and metro bundler console warnings

desfero:
Also fixes home screen jumping and skeleton screen border radius


packages/wallet/app/modules/wallet-connect/sagas.ts
```diff
@@ -17,14 +16,14 @@ export function* tryToConnectExistingSession(): SagaGenerator<void> {
     const hasExistingSession = yield* call(() => walletConnectManager.hasExistingSession());
 
     if (hasExistingSession) {
-      logger.info(`${MODULE_ID}: Existing wallet connect session found, attempting to use it...`);
```
arjunu, line 20: My idea was the prefix would help us filter the logs by module just like the debug in structlint. Or do you have any other ideas for filtering logs because we have a lot in the console.

PR #4512, opened 2020-07-29
Basic Mobile Message Signer

arjunu:


packages/wallet/app/components/modals/SignerModal/SignMessageSigner.tsx
```diff
@@ -17,9 +19,41 @@ type TExternalProps = {
...
+    <FormattedMessage id="wallet.message-signer.title.do-book-building" />
+  ),
+  [EJwtPermissions.ISSUER_UPDATE_NOMINEE_REQUEST]: (
+    <FormattedMessage id="wallet.message-signer.issuer-update-nominee-request" />
+  ),
+  [EJwtPermissions.ISSUER_REMOVE_NOMINEE]: (
+    <FormattedMessage id="wallet.message-signer.issuer-remove-nominee" />
+  ),
+  [EJwtPermissions.SIGN_TOS]: null,
+  [EJwtPermissions.MANAGE_USER_WALLET]: null,
```
arjunu, line 45: perhaps I should throw an error for these two?

packages/wallet/app/locales/en-en.json
```diff
@@ -11,5 +11,13 @@
...
+  "wallet.tx-list.name.transfer.transferred": "Sent {currency}",
+  "wallet.message-signer.title.change-email": "Resend access link",
+  "wallet.message-signer.title.submit-kyc": "Confirm your identity verification request",
+  "wallet.message-signer.title.submit-eto": "Publish your listing page",
+  "wallet.message-signer.title.upload-immutable-document": "Confirm document upload",
+  "wallet.message-signer.title.do-book-building": "Confirm your commitment",
+  "wallet.message-signer.issuer-update-nominee-request": "Confirm nominee request update",
+  "wallet.message-signer.issuer-remove-nominee": "Confirm nominee rejection",
+  "wallet.message-signer.no-permission": "Please confirm message request"
 }
```
arjunu, line 23: I copied these from web, I feel some of these need to be modified especially the default "no-permission" text

packages/wallet/app/modules/wallet-connect/sagaFunctions/connectEvents.ts
```diff
@@ -50,7 +50,7 @@ function* watchWalletConnectEvents(wcAdapter: WalletConnectAdapter): SagaGenerat
         }
 
         if (denied) {
-          managerEvent.meta.rejectRequest();
+          managerEvent.meta.rejectRequest(new Error("User rejected request"));
```
arjunu, line 53: so if we don't throw error here, `assertError` will fail

packages/wallet/app/components/modals/SignerModal/SignMessageSigner.tsx
desfero: do you know what's `MANAGE_USER_WALLET` is used for? I think SIGN_TOS we should support in case we disable the sign_tos permission that's active from the beginning but not sure about `MANAGE_USER_WALLET` as i don't see it's used in the code

packages/wallet/app/components/modals/SignerModal/SignMessageSigner.tsx
```diff
@@ -17,9 +19,41 @@ type TExternalProps = {
   reject: () => void;
 };
 
-const SignMessageSigner: React.FunctionComponent<TExternalProps> = ({ data: _data, ...rest }) => (
+const Titles: { [key in EJwtPermissions]: React.ReactNode } = {
```
desfero, line 22: `Record<EJwtPermissions, React.ReactNode>` for consistency?

desfero: Not sure it should be called with an upper case `T`? Maybe just `titles`?

packages/wallet/app/locales/en-en.json
desfero: we should maybe as @andywalton for some guidelines?

packages/wallet/app/modules/wallet-connect/lib/WalletConnectAdapter.ts
```diff
@@ -142,12 +142,26 @@ class WalletConnectAdapter extends EventEmitter2 {
         case ETH_SIGN_RPC_METHOD:
           try {
             const parsedPayload = parseRPCPayload(WalletConnectEthSignJSONRPCSchema, payload);
+            const digest = parsedPayload.params[1];
+            let { permissions }: { permissions: EJwtPermissions[] } = parseDigestString(digest);
+            permissions = permissions.filter(
+              (permission: EJwtPermissions) => permission !== EJwtPermissions.SIGN_TOS,
```
desfero, line 148: what's the reason to filter TOS? I think we should pass array up to UI layer and then show all permissions that you are going to sign

packages/wallet/app/modules/wallet-connect/lib/utils.ts
```diff
@@ -87,10 +88,13 @@ const isValidWalletConnectUri = (uri: string): uri is TWalletConnectUri => {
 const parseRPCPayload = <T extends UnknownObject>(schema: yup.ObjectSchema<T>, value: unknown) =>
   schema.validateSync(value, { stripUnknown: true });
 
+const parseDigestString = (digest: string) => JSON.parse(utils.toUtf8String(digest)) as TDigestJSON;
```
desfero, line 91: it doesn't really validate the shape after parsing, we should use a proper yup schema with enum validation. Also there can be also a case when we have new jwt permission not yet supported by the app so for that case, we should show some kind of fallback UI.

packages/wallet/app/modules/wallet-connect/lib/WalletConnectAdapter.ts
```diff
@@ -142,12 +142,26 @@ class WalletConnectAdapter extends EventEmitter2 {
...
+            );
+
+            if (permissions.length > 1) {
+              this.handleCallSigningError(
+                payload.id,
+                new InvalidJSONRPCPayloadError(payload.method, new Error("Too many permissions")),
```
desfero, line 154: Let's use custom error here

packages/wallet/app/components/modals/SignerModal/SignMessageSigner.tsx
arjunu: Removed `MANAGE_USER_WALLET`

packages/wallet/app/modules/wallet-connect/sagaFunctions/connectEvents.ts
arjunu: used `UserRejectedRequestError`

packages/wallet/app/components/modals/SignerModal/SignMessageSigner.tsx
arjunu: updated

packages/wallet/app/modules/wallet-connect/lib/WalletConnectAdapter.ts
arjunu: used `TooManyPermissionsError`

packages/wallet/app/modules/wallet-connect/lib/utils.ts
arjunu: Added `DigestJSONSchema` and handled unknown permissions

packages/wallet/ios/Podfile.lock
```diff
@@ -329,7 +329,7 @@ PODS:
     - React
   - react-native-config (1.2.0):
     - React
-  - react-native-flipper (0.51.0):
+  - react-native-flipper (0.47.0):
```
desfero, line 332: <3

arjunu:
Build: https://appcenter.ms/orgs/Neufund/apps/Neufund-Wallet-MVP-iOS/distribute/releases/24

desfero (approved):

glupiawymowka (approved):

desfero (approved):

Issue #4250, opened 2020-04-05
QR code scanning / WalletConnect flow


andywalton:
**Background**
The context for this ticket is to create the QR code scanning views to allow the Neufund mobile app to connect the the Neufund web app via WalletConnect.

**Designs**
<img width="1080" alt="Screenshot 2020-04-05 21 54 40" src="https://user-images.githubusercontent.com/609657/78508717-4af4bf00-7789-11ea-8e43-160d234030b1.png">

See Zeplin for the screen designs here under the WalletConnect - Scan QR code section: https://app.zeplin.io/project/5e6f48a43076c61012fd3793/dashboard?seid=5e89f42aad9e6e23fcc965cb  

Individual components are available in the Design System.

It makes use of the TopBar/ModalStack component and then default system messages to ask the user to gain access to their camera, the same with the error should a QR code not be recognised.  

**Acceptance criteria**
- [x] As a user I can Scan a QR code by tapping on the `QR code icon` on the Home view in the TopBar or via `Profile > Neufund Web > Scan QR code`
- [x] The first time I try do this, a system permissions dialog will ask me to confirm access to my device camera.
- [x] If I grant permission to the app to use my camera, I will see a view that will allow me to scan a QR code.
- [x] If the QR code is valid, it will connect, a confirmation signer modal will display that will confirm what the connection is and I’ll be able to Accept or Reject this connect. 
- [x] If it doesn’t recognise a QR code, ie. it’s an invalid QR I’m trying to scan, a default device error should display this error to me.

> Note - further improvements such as error handling and the ability to paste in a QR code URL to connect without a camera will happen post MVP.

**Technical notes**
@desfero -> please add technical notes here.

**Testing notes**
@glupiawymowka -> Same as acceptance criteria but edit this if you want.

glupiawymowka:
tested on .io

PR #4531, opened 2020-08-06
Wallet QR code handling

arjunu:


packages/wallet/app/modules/qr-code-scanner/sagas.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  logger.info("Unsupported QR code");
+
+  if (requiredQRCodeType === EQRCodeType.WALLET_CONNECT) {
+    yield put(notificationUIModuleApi.actions.showInfo("Unrecognised Wallet Connect QR Code"));
+  } else {
+    yield put(notificationUIModuleApi.actions.showInfo("Unrecognised QR Code"));
+  }
+
+  navigate(EAppRoutes.home);
```
arjunu, line 43: for now it closes QR Code screen, as part of #4250 re-scanning will be implemented

packages/wallet/app/modules/qr-code-scanner/utils/walletConnectUtils.ts
```diff
@@ -0,0 +1,74 @@
+import { IParseURIResult, IQueryParamsResult, IRequiredParamsResult } from "@walletconnect/types";
```
arjunu, line 1: I moved validation from `wallet-connect` module to `qr-code-scanner` module

packages/wallet/app/components/screens/WalletConnectSessionScreen/WalletConnectSessionScreen.tsx
```diff
@@ -76,7 +77,11 @@ const WalletConnectLayout: React.FunctionComponent<TStateProps & TDispatchProps>
           <Button
             layout={EButtonLayout.PRIMARY}
             testID="wallet-connect-session-screen.scan-qr"
-            onPress={() => navigation.navigate(EAppRoutes.qrCode)}
+            onPress={() =>
+              navigation.navigate(EAppRoutes.qrCode, {
+                requiredQRCodeType: EQRCodeType.WALLET_CONNECT,
```
arjunu, line 82: so that even if user scans a valid transfer request QR code it won't work

packages/wallet/app/components/QRCode.tsx
```diff
@@ -1,25 +1,35 @@
 import { UnknownObject } from "@neufund/shared-utils";
+import { RouteProp } from "@react-navigation/native";
 import React from "react";
 import { compose } from "recompose";
 
 import { QRCodeScanner } from "components/shared/qr-code-scanner/QRCodeScanner";
 
-import { walletConnectActions } from "modules/wallet-connect/actions";
+import { qrCodeScannerModuleApi } from "modules/qr-code-scanner/module";
+import { EQRCodeType } from "modules/qr-code-scanner/types";
```
desfero, line 9: let's reexport `types` from `module.ts` so then we can have a single import from `"modules/qr-code-scanner/module"`

packages/wallet/app/components/shared/qr-code-scanner/QRCodeScanner.tsx
```diff
@@ -11,10 +11,13 @@ const QRCodeScanner: React.FunctionComponent<TExternalProps> = ({ onRead }) => {
 
   return (
     <RNQRCodeScanner
+      fadeIn
```
desfero, line 14: isn't a fadeIn={true} default value?

packages/wallet/app/modules/qr-code-scanner/sagas.ts
```diff
@@ -0,0 +1,48 @@
...
+function* onScan(
+  action: TActionFromCreator<typeof qrCodeScannerActions, typeof qrCodeScannerActions.onScan>,
+): SagaGenerator<void> {
+  const { data, requiredQRCodeType } = action.payload;
+
+  const { logger } = yield* neuGetBindings({
+    logger: coreModuleApi.symbols.logger,
+  });
+
+  logger.info("Parsing QR code data" + data);
```
desfero, line 23: no sure it's a good practice to put `qr` code into logger as it may contain some kind of sensitive data. 
I think we can omit this `logger.info` give we log the action and it's the next thing done after action

packages/wallet/app/modules/qr-code-scanner/sagas.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { EAppRoutes } from "router/appRoutes";
+import { navigate } from "router/routeUtils";
+
+import { isValidWalletConnectUri } from "./utils/walletConnectUtils";
+
+function* onScan(
+  action: TActionFromCreator<typeof qrCodeScannerActions, typeof qrCodeScannerActions.onScan>,
+): SagaGenerator<void> {
+  const { data, requiredQRCodeType } = action.payload;
```
desfero, line 17: we should wrap saga with try-catch and implement error handling. Otherwise, in case of any errors, they will propagate up to the root saga and break the app

packages/wallet/app/modules/qr-code-scanner/sagas.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  // Eth/ERC20 token transfer
+  // if (isValidTransactionRequestUri(data)) {
+  // yield put(...);
+  // }
+
+  logger.info("Unsupported QR code");
+
+  if (requiredQRCodeType === EQRCodeType.WALLET_CONNECT) {
```
desfero, line 37: let's use switch here? Would be easier to add new types in the future

packages/wallet/app/modules/qr-code-scanner/utils/walletConnectUtils.ts
desfero: maybe we should still keep it inside `wallet-connect` module and expose as util so it's domain logic is in a single place? But don't have a strong opinions here.

packages/wallet/app/modules/wallet-connect/sagaFunctions/connectToURI.ts
```diff
@@ -32,10 +28,6 @@ export function* connectToURI(
   const uri = action.payload.uri;
 
   try {
-    if (!isValidWalletConnectUri(uri)) {
```
desfero, line 35: i think we should leave it here to make sure we always pass valid URI. We can omit it by introducing opaque type for wallet connect uri specifically but that seems like a lot more work.

yarn.lock
```diff
@@ -21061,6 +21069,11 @@ qs@6.7.0:
   resolved "https://registry.yarnpkg.com/qs/-/qs-6.7.0.tgz#41dc1a015e3d581f1621776be31afb2876a9b1bc"
   integrity sha512-VCdBRNFTX1fyE7Nb6FYoURo/SPe62QCaAyzJvUjwRaIsc+NePBEniHlvxFmmX56+HZphIGtV0XeCirBtpDrTyQ==
 
+qs@^6.5.2:
```
desfero, line 21072: is it possible to deduplicate `qs`? I see the below one can be upgraded with `yarn deduplicate --packages qs`

packages/wallet/app/components/shared/qr-code-scanner/QRCodeScanner.tsx
arjunu: you are right, removed

packages/wallet/app/modules/qr-code-scanner/sagas.ts
arjunu: removed

packages/wallet/app/modules/wallet-connect/sagaFunctions/connectToURI.ts
arjunu: but then we will be validating it twice. We already have opaque type `type TWalletConnectUri = Opaque<"WalletConnectUri", string>;` and I made sure `action.payload.uri` is  `TWalletConnectUri`

packages/wallet/app/modules/qr-code-scanner/sagas.ts
arjunu: added

packages/wallet/app/components/QRCode.tsx
arjunu: done

packages/wallet/app/components/shared/qr-code-scanner/QRCodeScanner.tsx
arjunu: yes, removed

packages/wallet/app/modules/qr-code-scanner/utils/walletConnectUtils.ts
arjunu: Ok I moved it back. It is better as this way the errors and the constants don't have to be exposed.

packages/wallet/app/modules/qr-code-scanner/sagas.ts
arjunu: added

yarn.lock
arjunu: good catch, done

packages/wallet/app/modules/qr-code-scanner/sagas.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    if (walletConnectModuleApi.utils.isValidWalletConnectUri(data)) {
+      yield put(walletConnectModuleApi.actions.connectToPeer(data));
+      return;
+    }
+
+    // Eth/ERC20 token transfer
+    // if (isValidTransactionRequestUri(data)) {
+    // yield put(...);
+    // }
```
Moejoe90, line 31: Is this on purpose?

arjunu: yes it's just a comment to show how to add the new case. The function `isValidTransactionRequestUri` is included in this PR but we don't have the UI for this yet

PR #4551, opened 2020-08-18
adds unlock feature as a button

Moejoe90:


packages/web/app/modules/wallet-view/sagas.spec.ts
```diff
@@ -226,6 +229,24 @@ describe("Wallet View", () => {
...
+                timestampOfLastChange: 1225,
+                status: EModuleStatus.IDLE,
+              },
+              txMonitor: {
+                txs: {
+                  pendingTransaction: generalPendingTxFixture(
+                    "0x16cd5aC5A1b77FB72032E3A09E91A98bB21D8988",
+                  ),
+                },
+              } as any,
```
desfero, line 249: can we avoid `as any`? Maybe Partial<TAppState> or `DeepPartial<TAPPState>`?

packages/web/app/modules/wallet-view/sagas.ts
```diff
@@ -172,51 +172,27 @@ export function* walletViewController(): Generator<any, void, any> {
...
-            totalBalanceEuro,
-          };
-        } else if (dataReloadRequired.tokenPriceData) {
-          const { balanceData, totalBalanceEuro } = yield* call(populateBalanceData);
-          const { transactionsHistoryPaginated, pendingTransaction } = yield* call(
-            populateTxHistory,
-          );
-
-          newState = {
+        const { balanceData, totalBalanceEuro } = yield* call(populateBalanceData);
```
desfero, line 185: The problem we have now is that when `tx history` is updated then a new reference for `balanceData` is generated which in turn rerenders the UI not related to tx history. Not a big deal but seems like a code smell. I would say we need to move all of that to selectors to derive data only when there is really a change in sub modules.

desfero (changes requested):
Small changes requested

packages/web/app/components/wallet/utils.tsx
```diff
@@ -134,16 +134,37 @@ export const createBalanceActions = (dispatch: Function): TBalanceActions => ({
...
     {
       dispatchAction: () => dispatch(actions.txTransactions.startUpgrade(ETokenType.ETHER)),
       disableIf: () => false,
       text: <FormattedMessage id="wallet.enable-icbm" />,
-      level: EBalanceActionLevel.SECONDARY,
+      level: EBalanceActionLevel.PRIMARY,
       dataTestId: "wallet.icbm-eth.upgrade-button",
     },
+    {
+      dispatchAction: () => dispatch(actions.txTransactions.startUnlockEtherFunds()),
```
desfero, line 162: can we remove dispatch(actions.txTransactions.startUnlockEtherFunds()) as it should never happen here given it's always disabled, same on line `179`. Also would be good to add some kind of comment why `disabledIf` is always true

packages/web/app/modules/wallet-view/sagas.spec.ts
Moejoe90: but this is a test file. I could argue its better like this to make it clear this is not typed correctly

desfero: the case if DeepPartial is that when you move part of states to for .e.g modules under the different name you get a clear error from tsc so you know what to fix. Also the same with any kind of renaming or type changes

packages/web/app/components/wallet/BalanceList.tsx
```diff
@@ -11,8 +11,8 @@ export type TBalanceListProps = { balances: TBalance[] };
 export const BalanceList: React.FunctionComponent<TBalanceListProps> = ({ balances }) => (
   <PanelRounded>
     <li className={styles.balanceList}>
-      {balances.map((b, i) => (
-        <Balance {...b} key={i} />
+      {balances.map(b => (
+        <Balance {...b} key={b.balanceId} />
```
AlexeyTuboltsev, line 15: unfortunately we can't ban numeric keys on lint level, but they definitely should be a red flag for reviewer

AlexeyTuboltsev: since all the data comes from sagas, there's no problem to generate normal string keys, even `math.random().toString()` would do if there's no sensible alternative to it

desfero (approved):

desfero (dismissed):

glupiawymowka (changes requested):
- [ ] `Icbm` should've been fixed to capital letters as per design
image

glupiawymowka (approved):

PR #4562, opened 2020-08-28
Import lodash helpers from submodules

desfero:
It's good for bundle size to import lodash from submodules. Unfortunately some of deps import full lodash so there is no difference in bundle size yet. But anyway would be good to import lodash properly for future.

This pr also removes `prerender` script given it was never used. 

**Testing notes**
Run #p3


packages/web/dangerfile.ts
```diff
@@ -2,12 +2,12 @@ import { danger, fail, message } from "danger";
 import { writeFileSync } from "fs";
 import { uniq } from "lodash";
 
-const main = (): void => {
```
Moejoe90, line 5: :*(

packages/web/scripts/prerender.sh
```diff
@@ -1,9 +0,0 @@
-#!/usr/bin/env bash
```
Moejoe90, line 1: :+1:

Moejoe90 (approved):

Moejoe90 (approved):

Issue #4554, opened 2020-08-20
[onboarding] - some inputs are having layout problems


glupiawymowka:
**Description**
After the latest release, TOS checkbox and recovery flow inputs broke visually

**Env**
All

**Steps to reproduce**
1. Go to sign up (any wallet) and check the TOS checkbox

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/187d0869-995d-4654-8d03-f25a8d6f20fc)

2. Go to the recovery flow and check the inputs

![image](https://user-images.githubusercontent.com/18315589/90755867-f12a0700-e2db-11ea-8326-5153e436af92.png)

3. The inputs in the "BackupSeedVerify" are also slightly broken 
![image](https://user-images.githubusercontent.com/18315589/90756138-4b2acc80-e2dc-11ea-8409-b27c1e3cbab7.png)

**Expected behavior**
- [ ] checkbox is visible
- [ ] borders of the seed phrase inputs are well aligned

PR #4540, opened 2020-08-12
show warning if payout is too low

AlexeyTuboltsev:

PR #4571, opened 2020-09-09
#4554 Fix broken inputs

desfero:
**Testing notes**
Checkbox was broken in #4540 so let's make sure that payout lower than gas checkbox is still working properly.
Also this PR makes checkboxes in sync with design-system https://app.zeplin.io/project/5dcae4b293ec9fff68a73bdd/screen/5e4e8ea8a469bb7a97b85620 and adds missing states (hover on label and focus)

### DIFF
--- a/packages/design-system/.storybook/main.js
+++ b/packages/design-system/.storybook/main.js
@@ -1,4 +1,6 @@
+const { DefinePlugin } = require("webpack");
 const path = require("path");
+
 const paths = require("../webpack/paths");
 
 module.exports = {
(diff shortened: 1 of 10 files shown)


arjunu (changes requested):
Can you make the checkmark transparent/hidden when unchecked? It's white right now
image

also in design system storybook config can you add
```
config.plugins.push(
      new webpack.DefinePlugin({
        __DEV__: false,
      }),
    );
```

arjunu (approved):

glupiawymowka (approved):

Issue #4534, opened 2020-08-10
[Investor dashboard] - Portfolio widget shows empty asset values 


glupiawymowka:
**description**
The user who opens a new account can see some assets with "0" balances in their dashboard. 
This logic is only correct for mobile app but should not appear for web version. 

"Your assets" in Portfolio tab also seem broken

**Env** 
Stage, io

**Steps to reproduce**
1. Open a new investor account
2. Check the portfolio widget \
3. Check "Your assets" widget in the Portfolio tab

**Actual results**

- [x] empty asset balances are shown in the dashboard 

![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/58e5508e-d00f-406b-b1de-34ce69c17517)

- [x] Some empty broken-looking position in "Your assets" 
![image](https://user-images.githubusercontent.com/18315589/89783349-1453fa00-db17-11ea-9053-f605fbdf4211.png)

**Expected results**

- [ ] "You have no assets in your portfolio yet." copy should be brought back for the described case
- [ ] "0 EUR" total balance should stay 
![image.png](https://images.zenhubusercontent.com/5ae1e2334b5806bc2bd0c3bb/58cfde31-6e7a-403c-bbee-6a9783ec8c1b)

- [ ] "Your assets" widget, when empty, should say "You have no assets in your portfolio."
![image](https://user-images.githubusercontent.com/18315589/89783478-4cf3d380-db17-11ea-9c4d-3626fa39e51f.png)

- [ ] portfolio widget logic for the mobile app should stay unbroken

PR #4577, opened 2020-09-15
#4534 Fix portfolio items

desfero:


yarn.lock
```diff
@@ -5255,15 +5255,6 @@
     bn.js "4.11.8"
     detect-browser "4.8.0"
 
-"@webassemblyjs/ast@1.8.5":
```
desfero, line 5258: i think yarn.lock change was caused by merging one of dependabot security fixes. Probably they don't run a full `yarn install` to fix transitive dependencies

packages/web/app/test-e2e/investor/portfolio/PortfolioStats.spec.e2e.ts
```diff
@@ -21,4 +24,20 @@ describe("Portfolio Stats", () => {
...
+
+    goToDashboard();
+
+    cy.get(tid("dashboard.portfolio-stats")).within(() => {
+      cy.get(tid("dashboard.portfolio-stats.eur-equiv-total")).should(
+        "have.text",
+        "1 087 137.35 EUR",
+      );
+
+      cy.get(tidStartsWith("portfolio.stats.token")).should("have.length", 1);
```
desfero, line 39: This will make sure that the bug won't happen again. The downside is that in case fixture is modified test can fail. I think it should not happen often so we are safe

packages/web/app/test-e2e/utils/index.ts
```diff
@@ -260,51 +249,6 @@ export const removePendingExternalTransaction = () => {
   }).then(() => assertWaitForExternalPendingTransactionCount(0));
 };
 
-export const getFormattedNumber = (
```
Moejoe90, line 263: :+1:

Moejoe90 (approved):

glupiawymowka (approved):

Issue #4364, opened 2020-05-19
`TModuleAction` do not work properly with module without actions.


desfero:
**Description**
When the module is with dependencies and at least one of dependencies is without actions the final `TModuleAction` type is inferred to never.

```
const setupWebTxHistoryModule = () => {
  const webTxHistoryModule = {
    id: "test",
    api: {
      actions: {
        test: createActionFactory("fdfsd"),
      },
    },
  };

  return webTxHistoryModule;
};

const setupWeb2TxHistoryModule = () => {
  const webTxHistoryModule = {
    id: "test",
    api: {
      // NO ACTION
    },
  };

  return [setupWebTxHistoryModule(), webTxHistoryModule];
};

type Test = TModuleActions<typeof setupWeb2TxHistoryModule>; // never 
```

**Environment**

**Steps To Reproduce**
as in description

**Actual behavior**
Actions are `never` type

**Expected behavior**
Proper actions should be generated by type

**Comments, context, logs**
After the quick research, I narrowed the bug to `Values` type that for `{}` return never (and that's correct). So to fix this one we need to branch type generation in case there are no actions.

**Screenshots**

PR #4580, opened 2020-09-17
[All devs approval required] #4364 introduce createModuleSetup util

desfero:
**Technical notes**
This pr adds unit tests for our module system type utils and also introduces `createModuleSetup` utility helper.
Also `assertType` and `AssertEqual` was removed in favor of https://github.com/dsherret/conditional-type-checks as it provides more utils to type testing.

In general `createModuleSetup` is just an identity function that provides type narrowing and type checking.
In the future it can also check in `dev` mode if all modules have a unique id. Can also we used in 4426 to handle dependencies that are transitive.


arjunu (changes requested):
There a couple of web modules that use the module system and should use `createModuleSetup`.

arjunu (approved):

Moejoe90 (approved):
Looks good from my side

AlexeyTuboltsev (approved):

PR #4584, opened 2020-09-23
#3930 migrate shared-utils to eslint

desfero:


packages/eslint-config/base.js
```diff
@@ -122,8 +122,8 @@ module.exports = {
     "@typescript-eslint/no-magic-numbers": [
       "error",
       {
-        // 0 is often used to access first element of an array
-        ignore: [0],
+        // ignore some common numbers
```
arjunu, line 125: there is `ignoreArrayIndexes` option for this rule, should we use?

packages/shared-utils/src/utils/StringUtils.ts
```diff
@@ -9,6 +10,7 @@ export function isExternalUrl(s: string): boolean {
 
 export function isEmail(s: string): boolean {
   // General Email Regex (RFC 5322 Official Standard)
+  // eslint-disable-next-line no-control-regex
```
arjunu, line 13: can this be fixed with `new RegExp(...)`?

packages/shared-modules/src/tests/auth/createLightWalletWithKeyPair.ts
```diff
@@ -22,10 +22,10 @@ export const createLightWalletWithKeyPair = async (
     salt,
   });
   // create keypair
-  const keyFromPassword = promisify<any>(
+  const keyFromPassword = promisify<[string], any>(
     lightWalletInstance.keyFromPassword.bind(lightWalletInstance),
   );
-  const walletKey: any = await keyFromPassword(password);
+  const walletKey = await keyFromPassword(password);
```
Moejoe90, line 28: :+1:

packages/shared-utils/src/opaque-types/types.ts
```diff
@@ -3,4 +3,4 @@
  * see: https://codemix.com/opaque-types-in-javascript/
  */
 export type Opaque<K, T> = T & { __TYPE__: K };
-export type OpaqueType<O extends Opaque<any, any>> = O["__TYPE__"];
+export type OpaqueType<O extends Opaque<unknown, unknown>> = O["__TYPE__"];
```
Moejoe90, line 6: :+1:

Moejoe90 (approved):
looks good to me

packages/eslint-config/base.js
desfero: it won't fix the problem fully as 0 is also used often outside of accessing arrays AFAIK

packages/shared-utils/src/utils/StringUtils.ts
desfero: Not sure really as this rule checks some kind of control characters that we use in regex for email both `/regex/` and `new RegExp` gonna fail

arjunu (approved):

PR #4594, opened 2020-09-30
Replace EtherscanTxLink with EtherscanAddressLink for addresses

arjunu:


packages/web/app/components/shared/links/EtherscanLink.tsx
```diff
@@ -25,7 +31,8 @@ const EtherscanTxLink: React.FunctionComponent<IEtherscanTxLink & CommonHtmlProp
 );
 
 interface IEtherscanAddressLink {
-  address: string;
+  // TODO remove string once EthereumAddress is used everywhere
+  address: string | EthereumAddress | EthereumAddressWithChecksum;
```
Moejoe90, line 35: what stops us from making this change already?

Moejoe90 (changes requested):
one question

packages/web/app/components/shared/links/EtherscanLink.tsx
arjunu: it's a big change, we can do it as a separate ticket.

glupiawymowka (approved):

Moejoe90 (approved):

PR #4600, opened 2020-09-30
Arjunu/offline mode

arjunu:


packages/wallet/app/modules/network/sagas.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export function* tryAgain(): SagaGenerator<void> {
+  const { isConnected } = yield* call(NetInfo.fetch);
+  console.log("tryAgain", { isConnected });
+
+  yield put(networkActions.onNetworkStateChange(isConnected));
+}
+
+export function* networkSaga(): SagaGenerator<void> {
+  yield takeLeading(networkActions.startWatchNetworkState, watchNetworkStateSaga);
```
desfero, line 36: we should use `takeLatestUntil`(should be in sagas package if not check one of my pr that are in review at the moment) to `stopWatchNetworkState`.

packages/wallet/app/modules/network/sagas.ts
```diff
@@ -0,0 +1,38 @@
...
+      return;
+    }
+
+    if (networkState) {
+      yield put(networkActions.onNetworkStateChange(networkState.isConnected));
+    }
+  }
+}
+
+export function* tryAgain(): SagaGenerator<void> {
```
desfero, line 28: i would name function and action `forceNetworkStateCheck`

packages/wallet/app/modules/network/sagas.ts
```diff
@@ -0,0 +1,38 @@
...
+import NetInfo from "@react-native-community/netinfo";
+import { networkActions } from "modules/network/actions";
+import { createNetworkStateChannel } from "modules/network/utils";
+
+export function* watchNetworkStateSaga(): SagaGenerator<void> {
+  const networkStateChannel = yield* call(createNetworkStateChannel);
+
+  while (true) {
+    const { stop, networkState } = yield* race({
+      stop: take(networkActions.stopWatchNetworkState),
```
desfero, line 11: ok, we need to check two cases:

- Can we use `takeLatestUntil` to remove the need for `race` and still have the netinfo channel properly unsubscribed.
- We need to check what in case of error thrown inside while loop. Will the error unsubscribe from channel? If not we need to wrap while loop into a try catch block and close the channel manually.

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -102,6 +114,8 @@ function* handleAppStateChanges(): SagaGenerator<void> {
 }
 
 function* initAuthSigned(): SagaGenerator<void> {
+  yield put(networkModuleApi.actions.startWatchNetworkState());
```
desfero, line 117: i think that should not be correlated with auth state. How you can do authentication if you are not able to generate jwt token?

packages/wallet/ios/wallet.xcodeproj/project.pbxproj
```diff
@@ -573,7 +640,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 5F4B4B4B38B6CC79E9FB1328 /* Pods-wallet.debug.xcconfig */;
 			buildSettings = {
-				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = "$(inherited)";
```
arjunu, line 643: See https://github.com/CocoaPods/CocoaPods/issues/5981

packages/wallet/app/modules/network/sagas.ts
arjunu: changed

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -103,9 +116,7 @@ function* handleAppStateChanges(): SagaGenerator<void> {
 
 function* initAuthSigned(): SagaGenerator<void> {
   yield* call(walletConnectModuleApi.sagas.tryToConnectExistingSession);
-
```
arjunu, line 106: remove blank lines to make it easier to read

packages/wallet/app/modules/network/sagas.ts
```diff
@@ -0,0 +1,44 @@
...
+  });
+  logger.info("Setting up network state watcher");
+
+  let currentState: TNetworkState | null = null;
+  const networkStateChannel = yield* call(createNetworkStateChannel);
+
+  try {
+    while (true) {
+      const nextState = yield* take(networkStateChannel);
+      if (currentState === null || currentState.isConnected !== nextState.isConnected) {
```
arjunu, line 21: there are change events from the listener if `isConnected` didn't change but other variables in the network state change. So I gotta compare with previous state.

PR #4603, opened 2020-10-01
Don't load ETO if user not verified

arjunu:


packages/web/app/modules/eto-flow/sagas.ts
```diff
@@ -313,16 +312,18 @@ export function* publishEtoData({ logger }: TGlobalDependencies): Generator<any,
   }
 }
 
-export function* loadIssuerStep(): Generator<any, any, any> {
-  yield neuCall(loadIssuerEto);
```
arjunu, line 317: I couldn't figure out what caused this issue/how it used to work before (if it used to work). No recent changes to this code in commit history.

Issue #2934, opened 2019-05-15
Fix character limiting in form widgets


AlexeyTuboltsev:
**Background**
1. character limiting is buggy
2. character limiting is  handled differently across different form widgets, using 2 different, sometimes overlapping props. We should unify this
2. CampaigningActivatedInvestorApprovedWidgetLayout.tsx should use moneyInput

**Designs**
na
**Acceptance criteria**
* [ ] character limiting works on text area, textinput, numbers showing under the input are correct, especially when user tries to type in more chars then allowed
* [ ] (TECH, QA can't test it) character limiting props and typing are consistent across widgets
* [ ] CampaigningActivatedInvestorApprovedWidgetLayout commitment input works as expected, numbers appear in right format.

**Technical notes**

**Testing notes**
na

glupiawymowka:
imo we need a design/product input here to decide how big the limit should be for different field types. 

We definitely need to fix it to avoid the following cases: 
- [x] form not saving due to the long Allocation label, no information what's wrong (happens for the other inputs, too) 
![image](https://user-images.githubusercontent.com/18315589/67682148-002b7b80-f98f-11e9-84e4-3820bf8997fe.png)
- [x] super long email address that got accepted (and verified!)
![image](https://user-images.githubusercontent.com/18315589/67682281-484a9e00-f98f-11e9-92a4-0fa3e0389173.png)

andywalton:
Regarding the handling of super long email addresses... from all my "internet" reading it's summarised best as follows:

> Character Limitations in an Email Address
> The total length of an email address in the original standard was 320 characters. The standard limited the username to 64 characters and the domain name to 255 characters. (If you're doing the math, that adds up to 319. The final character is the @ symbol in between.)

The challenge is therefore not to limit what the user enters as 320 characters would be very long in anycase, but how we display this in the frontend. I'd normally recommend using an ellipse (...) and we shorten the text when it reaches a specific character length. 

The problem is we should not do this for email addresses as it has it's difficulties, how many characters is enough, do we do it pre @ symbol or after and/or both. We could simply do it and cut off the domain part.

It's hard because we need to handle this responsively as well. We should also show the user their email address and make it easy for them to copy and paste and for general recognition.

I'd recommend we wrap the text onto a second (third, fourth ...) line for this. Please use css applied to the email as per here: https://developer.mozilla.org/de/docs/Web/CSS/word-wrap

This is very well supported.

[...]

PR #4606, opened 2020-10-04
#2934 Fix characters limits

desfero:


packages/design-system/src/utils/yupUtils.ts
```diff
@@ -20,13 +21,28 @@ export const getSchemaField = <T>(
 
 export const isRequired = compose(includes("required"), getSchemaTests);
 
+// TODO: Refactor to use Yup's `describe` method
+export const getSchemaConstraints = (schema: any) =>
```
desfero, line 25: typed as `any`. Should be typed properly when we migrate to newest YUP

patches/formik+2.1.7.patch
```diff
@@ -1,8 +1,8 @@
 diff --git a/node_modules/formik/dist/formik.cjs.development.js b/node_modules/formik/dist/formik.cjs.development.js
-index a40e31e..6c3e58e 100644
+index 3a53834..c1b380f 100644
```
desfero, line 2: Bug is still now fixed :(

Issue #3775, opened 2019-11-25
Migrate inputs to the new design


desfero:
**Background**
As a part of this ticket, we need to migrate all our inputs to the new designs (labels where already migrated in #3748).

**Designs**
See input styleguide here: https://zpl.io/VY1M3nJ and components styles may be found here: https://zpl.io/bzXwm07

Note, this doesn't cover all our input styles yet. These should work fairly well across the platform but they were specifically looked at for the recent KYC changes and incoming signup, login and onboarding changes.

For example, they might mess with our currency input fields, I'd therefore recommend we leave those input fields as per the current designs and we address them when we get to those views as we consolidate the visual UI changes across the platform.

Please let @andywalton know if you have any questions.

**Acceptance criteria**
- [ ] All inputs use consistent new inputs style for the styleguide.

**Technical notes**
Please check the rich text editor whether it matches a new style guide as it's disconnected from the `input` tag and therefore generic input styles. Also, make sure all variants are in style guide (for e.g. input with a prefix, checkbox tag, etc..., to check all available styles go through storybook for `fields`). Would be good also to get rid of `reactstrap` inputs from the codebase 

**Testing notes**

andywalton:
Updated with input field design links to Zeplin.

Moejoe90:
Lets split this ticket into a couple of tickets as the migration process will be very painful and slow.

PR #4713, opened 2020-11-26
Migrate inputs to design system

desfero:
A top-level PR with all input changes.

This PR closes all tickets assigned to #3775 epic.

**Testing notes**
- [x] Verified Neufund Investor (https://github.com/Neufund/platform-frontend/issues/4764) was fixed
- [x] `FormFieldColorful` removed from the codebase. That means ICBM migration tool will use a normal input rather then a custom one


glupiawymowka (changes requested):
- [ ] any `token send` amount is marked as optional, while not true (the input correctly won't allow to continue without it)  
image

glupiawymowka (dismissed):

glupiawymowka (approved):

PR #4779, opened 2021-02-04
#3775 Migrate final inputs 

desfero:
Last PR related to input migration with a bunch of refactoring.

Run #p3


packages/web/app/utils/react-connected-components/AsyncIntervalScheduler.ts
```diff
@@ -1,5 +1,4 @@
-import { ILogger } from "@neufund/shared-modules";
-import { isPromise } from "@neufund/shared-utils";
```
arjunu, line 2: accidentai change?

packages/web/app/utils/react-connected-components/AsyncIntervalScheduler.spec.ts
```diff
@@ -1,5 +1,4 @@
-import { noopLogger } from "@neufund/shared-modules";
-import { safeDelay } from "@neufund/shared-utils";
```
arjunu, line 2: accidentai change?

arjunu (changes requested):
Noticed few issues

1. In Profile -> ICBM MIGRATION TOOL widget, type in the input and hit backspace it crashes.
2. There's a gap between the Additional Info input
 and "co/ apartment" 
 image
3. Spacing between fields has increased in this form
image
4. The units here looks kind of weird. Maybe ask @andywalton to give it a different color?
image
5. Should be single line I guess
image


Storybook issues:
- can't type in forms -> bare fields -> inputs stories, I think it's missing a value prop
- is this supposed to be in two lines?
image
- forms -> fields -> text area fields story doesn't load

arjunu (approved):

PR #4783, opened 2021-02-11
Update critical error page design

arjunu:
To test, go to `/test/critical-error`

Run #p3


packages/web/app/components/layouts/CriticalError.module.scss
```diff
@@ -1,32 +1,27 @@
+.wrapper {
```
desfero, line 1: not needed because the layout component is misconfigured. All layouts should be wrapped in a layout container because it has a proper `grid-template-areas`. We should maybe refactor `LayoutContainer` to don't have anything related to the authorization state so it can be reused for critical error layout.

packages/web/app/components/layouts/CriticalError.tsx
```diff
@@ -1,29 +1,30 @@
...
-      <p className={styles.errorDetail} data-test-id="critical-error-page-error-detail">
-        <FormattedMessage id="critical-error.error-detail" values={{ message }} />
-      </p>
-    )}
-    <p className={styles.description}>
-      <FormattedHTMLMessage tagName="span" id="critical-error.paragraph" />
-    </p>
-  </section>
+        <p className={styles.description}>
+          <FormattedHTMLMessage tagName="span" id="critical-error.paragraph" />
```
desfero, line 23: FormattedMessage?

desfero (changes requested):
Some small change to layout

packages/web/app/components/layouts/CriticalError.tsx
arjunu: updated

packages/web/app/components/layouts/CriticalError.module.scss
arjunu: `LayoutContainer` is used in a lot places directly and indirectly. So I implemented a quick solution by renaming it to `LayoutContainerWithModals` so whenever we see it somewhere in code, we can replace it with `LayoutContainer` which is now just a div with `.layout` class

desfero (approved):

glupiawymowka (approved):

PR #4788, opened 2021-02-15
Issuer dashboard copy changes

arjunu:


packages/web/app/components/modals/wallet-access/AccessWallet.tsx
```diff
@@ -61,6 +59,8 @@ export const AccessWalletLayout: React.FunctionComponent<TComponentProps> = ({
   walletSubType,
   inputLabel,
 }) => {
+  invariant(walletType, "Wallet type should be defined");
```
desfero, line 62: is there a need for that? I see we have `assertNever` on line 139

packages/web/app/components/settings/kyc-states/KycStatusWidget.tsx
```diff
@@ -30,18 +30,31 @@ export type IKycStatusWidgetProps = THocProps<typeof connectKycStatusWidget> & I
```
desfero, line 30: I would say we should split that by 2 separate components right now we have a ton of ifs here. 
Would be good to have KycInvestorStatusWidget, KycIssuerStatusWidget. Also what about Nominee?

packages/web/intl/locales/en-en.json
```diff
@@ -1270,20 +1269,31 @@
...
+  "settings.kyc-status-widget.status.pending.business.issuer": "Thank you for submitting your account verification request! We will review it within 7-14 business days and contact you if we require further information. Otherwise, we will notify you by email once your account has been successfully verified. To contact Customer Care, please click <a href=\"{url}\" target=\"_blank\">here</a>.",
+  "settings.kyc-status-widget.status.pending.individual": "Thank you for submitting your idensettings.kyc-status-widget.status.ignoredtity verification request! We will review it within 2-3 business days and contact you if we require further information. Otherwise, we will notify you by email once your account has been successfully verified. To contact Customer Care, please click <a href=\"{url}\" target=\"_blank\">here</a>.",
```
desfero, line 1289: `Thank you for submitting your idensettings.kyc-status-widget.status.ignoredtity verification request` some typo

packages/web/intl/locales/en-en.json
```diff
@@ -1270,20 +1269,31 @@
...
+  "settings.kyc-status-widget.status.pending.individual.issuer": "Thank you for submitting your account verification request! We will review it within 2-3 business days and contact you if we require further information. Otherwise, we will notify you by email once your account has been successfully verified. To contact Customer Care, please click <a href=\"{url}\" target=\"_blank\">here</a>.",
   "settings.kyc-status-widget.status.rejected": "We regret to inform you that after careful review, your identity verification request has been rejected. To contact Customer Care, please click <a href=\"{url}\" target=\"_blank\">here</a>.",
+  "settings.kyc-status-widget.status.rejected.issuer": "We regret to inform you that after careful review, your account verification request has been rejected. To contact Customer Care, please click <a href=\"{url}\" target=\"_blank\">here</a>.",
```
desfero, line 1292: Is there a reason we duplicate all strings here even when they are equal?

desfero (changes requested):

packages/web/app/components/modals/wallet-access/AccessWallet.tsx
arjunu: replaced with `nonNullable`

packages/web/app/components/settings/kyc-states/KycStatusWidget.tsx
arjunu: Nominee stays same as before, we will change it later it seems.

arjunu: Used `isIssuer` condition inside strings so most of them are removed from here now

packages/web/intl/locales/en-en.json
arjunu: Oops fixed

desfero (dismissed):

glupiawymowka (approved):

PR #4789, opened 2021-02-15
Small fixes to logout flow

desfero:


.circleci/config.yml
```diff
@@ -296,6 +300,22 @@ workflows:
   web:
     jobs:
       - web_install-lint-test
+        check-artifacts: true
```
desfero, line 303: Not sure if there is a better way to run job with a conditional parameter that's based on git branch without such duplication. Tried to find something in the docs but no such usages there

packages/web/app/modules/auth/user/sagas.ts
```diff
@@ -163,6 +163,8 @@ export function* handleLogOutUserInternal(
 ): Generator<any, void, any> {
   yield neuCall(logoutUser);
 
+  yield put(actions.init.start(EInitType.APP_INIT));
```
desfero, line 166: if the navigation happens-before reinit then
- `AppComponent` is mounted
- `actions.init.start` sets `isInProgress` to true therefore `AppComponent` is unmounted and `LoadingIndicator` mounted
- `actions.init.start` sets `isInProgress` to false therefore `AppComponent` is mounted

As we can see the first mount is a waste mount and is causing some actions to be sent twice (for e.g. `LOAD_ETOS`)

arjunu (approved):

PR #4790, opened 2021-02-16
Use shared set-env

arjunu:
Use set-env for wallet to create .env since it was dependant on Circle CI env. Also moved the script to root of the repo to be shared with web package.


packages/web/package.json
```diff
@@ -29,7 +29,10 @@
     "lint:fix:staged": "concurrently -p \"[{name}]\" -n \"TSC,LINT-STAGED\" -c \"black.bold.bgYellow,black.bold.bgMagenta\" \"yarn tsc\" \"lint-staged\"",
     "modernizr": "modernizr -c modernizr-config.json -d app/external/unsupported-browser/modernizr-custom.min.js",
     "prepare": "yarn contract-artifacts:generate",
-    "set-env": "node scripts/set-env.js",
+    "set-env:commit": "node ../../scripts/set-env.js packages/web ./app/config/.env.template ./app/config/.env.e2e.commit",
+    "set-env:e2e": "node ../../scripts/set-env.js packages/web ./app/config/.env.template ./app/config/.env.e2e",
+    "set-env:io": "node ../../scripts/set-env.js packages/web ./app/config/.env.template ./app/config/.env.e2e.io",
+    "set-env:storybook": "node ../../scripts/set-env.js packages/web ./app/config/.env.template ./app/config/.env.e2e.storybook",
```
arjunu, line 35: Unfortunately this makes package json bigger but at least circle CI config is free from package specific information

arjunu: It would be nice to not have to pass package path `packages/web` to `set-env` but I couldn't figure out an easy way.

desfero (approved):

PR #4791, opened 2021-02-16
CI run jobs only for changed packages

arjunu:


packages/design-system/src/components/buttons/Button.tsx
```diff
@@ -63,9 +63,9 @@ const ButtonLayout = React.forwardRef<
     {isLoading ? (
       <>
         {/*
-                &nbsp; makes button the same in height as normal button
-                (avoids height jumping after switching to loading state)
-              */}
```
arjunu, line 68: dummy change for testing

.circleci/monorepo.json
```diff
@@ -0,0 +1,70 @@
+{
```
arjunu, line 1: @desfero I didn't include all files from the root. Just ".circleci", "yarn.lock", and  "package.json" because we have README, LICENSE etc so whitelisting is easier than blacklisting

.circleci/monorepo.json
```diff
@@ -0,0 +1,70 @@
...
+  "packages": {
+    "design-system": [
+      ".circleci",
+      "package.json",
+      "yarn.lock",
+      "packages/eslint-config",
+      "packages/shared-utils",
+      "packages/design-system",
+      ":!packages/**/*.md"
```
desfero, line 10: is that excluding md files? What about running md files though linter?

packages/design-system/src/components/buttons/Button.tsx
desfero: can be reverted now?

.circleci/config.yml
```diff
@@ -68,267 +97,384 @@ commands:
...
+                at: /home/circleci/app
+            # Add swap to machine
+            - run: ./.circleci/add-swap.sh
+            - run: ./.circleci/install-yarn.sh
+            - restore-yarn-cache
+            - restore-cypress-cache
+            - run:
+                command: source /opt/circleci/.nvm/nvm.sh && yarn workspace @neufund/web show-tags
+            - run:
+                command: source /opt/circleci/.nvm/nvm.sh && yarn workspace @neufund/web show-tags
```
desfero, line 323: show tags duplicated?

.circleci/config.yml
```diff
@@ -337,37 +483,45 @@ workflows:
               only:
                 - master
     jobs:
-      - web_install-lint-test_nightly
-
-      - web_build-docker-test-e2e:
+      # this is same as pr-and-master-workflow minus preflight_checks & deploy-storybook
+      - repo_install
+      - shared-utils:
           requires:
```
desfero, line 489: Is it possible to move that as a comman/job that will have a parameter `preflight_checks` and `deploy-storybook`? So we kinda don't duplicate the same code twice?

desfero (changes requested):

.circleci/config.yml
arjunu: removed

packages/design-system/src/components/buttons/Button.tsx
arjunu: Prettier didn't complain so I left, reverted it now

.circleci/monorepo.json
arjunu: removed

.circleci/config.yml
arjunu: Setting a cron trigger makes the workflow not run without it. I moved artifacts check into install job like you did with condition. Nightly is still duplicated with trigger and storybook job

desfero (approved):

PR #4797, opened 2021-02-18
Improve insufficient ETH for gas error message

arjunu:
Run #p3


packages/web/app/modules/tx/validator/sagas.ts
```diff
@@ -102,6 +103,8 @@ export function* validateGas(
   const costUlps = multiplyBigNumbers([txDetails.gasPrice, txDetails.gas]);
   const valueUlps = subtractBigNumbers([maxEtherUlps, costUlps]);
 
+  yield put(actions.txValidator.setGasValidationInfo(convertFromUlps(costUlps).toString()));
```
desfero, line 106: i see that we need some refactoring to be done here in the future. 
For e.g we can refactor validation `notifications` to have also a payload and then don't throw an error but rather put a `NOT_ENOUGH_GAS` as a validation notification.

packages/web/app/modules/tx/validator/selectors.ts
```diff
@@ -14,3 +20,15 @@ export const selectTxValidationNotifications = (
 export const selectInvestmentFLow = (
   state: TAppGlobalState,
 ): DeepReadonlyObject<IInvestmentFlowState> => state.investmentFlow;
+
+export const selectTxValidatorGasCostEth = (state: TAppGlobalState) =>
+  state.txValidator.gasValidationInfo.gasCostEth;
+
+export const selectTxValidatorGasCostEur = (state: TAppGlobalState) => {
```
desfero, line 27: that's a good example for `createSelector` to memorize and don't rerun `convertFromUlps,multiplyBigNumbers` multiple times

packages/web/app/test-e2e/investor/wallet/withdraw/Withdraw.spec.e2e.io.ts
```diff
@@ -142,6 +143,19 @@ describe("Wallet Withdraw", () => {
       assertWithdrawButtonIsDisabled();
     });
 
+    it("should show error when ETH isn't enough to pay for tx costs #wallet #withdraw #p3", () => {
+      cy.get(tid("wallet.eth.withdraw.button")).awaitedClick();
+
+      cy.get(tid("modals.tx-sender.withdraw-flow.withdraw-component.value")).type(
+        (ethBalance + 1).toString(),
```
desfero, line 150: let's move `ethBalance + 1` as a variable `ethBalanceAboveMax`

desfero (changes requested):
Some small changes

packages/web/app/modules/tx/validator/selectors.ts
arjunu: updated

packages/web/app/test-e2e/investor/wallet/withdraw/Withdraw.spec.e2e.io.ts
arjunu: updated

packages/web/app/modules/tx/validator/sagas.ts
arjunu: Ok I added a TODO. It is a big change since we both sender saga and validator saga expects it to throw and we are showing the error in the UI

desfero (dismissed):

glupiawymowka:
hey, looks like there's something wrong with displaying the value in ETH. 


sorry for Polish :D But it basically shows that 5.81€ is 0.0044 ETH and not <0.0001, while 0.0001 is actually 0.13€

desfero (approved):

glupiawymowka (approved):

Issue #4795, opened 2021-02-17
Highlight that transaction costs are high due to Ethereum network congestion


andywalton:
This ticket is all about trying to better prevent a user accepting a negative net payout from NEU due to the gas price being higher than the actual payout they receive.

**Example from a real customer:**
![customer-payout-expensive-gas](https://user-images.githubusercontent.com/609657/108835313-972e3500-75cf-11eb-969c-34909df423f7.jpg)

**Improve it as follows:**
![Negative payout prevention](https://user-images.githubusercontent.com/609657/108836621-58997a00-75d1-11eb-9d62-38e367b6b2a2.png)

1. Add a new checkbox below the existing high transaction cost checkbox that highlights a negative net payout will result if they proceed with the transaction.

'**READ THIS.** I’d like to proceed and accept this transaction even though it will result in a total net payout of **≈ -77.48 EUR**.'

2. Update the copy of the first checkbox to the below.

`**IMPORTANT.** I understand that due to the current high Ethereum network gas prices, the transaction cost I pay may be higher than the actual payout I receive.`

3. Update the intro copy. You can also remove the unnecessary link to the **NEU token holder agreement** as it currently simply links to commit.neufund.org which is incorrect.

`Please find below a summary of the ETH payout you will receive as a NEU token holder. Once you accept this payout, it will automatically become available in your wallet balance.`

4. Please align the crypto column so that the values align up to the right below each other without offset spacing.

5. Please add a space between the approximation symbol (`≈`) and the value. Ie.  `≈-77.48 EUR`. should be `≈ -77.48 EUR`. We want to make sure there is at least enough of a visual separation so that negative values are more clearly displayed.

Lastly please increase the TOTAL NET PAYOUT font size from 14px to 16px.

**Technical notes:**
The two checkboxes, for now, should be displayed together, and both need to be checked in order to enable the disabled Accept button. So however we currently determine when to show the current checkbox will possibly be what should be used. 

Please however make sure we display these checkboxes always when there is a negative net total payout. 

Thanks

glupiawymowka:
tested on stage for low payout in ETH and nEUR

PR #4800, opened 2021-02-24
#4795 highlight costly payout transactions

desfero:
**Testing notes**
Besides fixes mentioned in the ticket:
 - we standardized the use of ≈ to be always `<space>≈<space>`. 
 - we used `~` on the portfolio page. That symbol can be confusing for negative values as they became `~-100 EUR`. Was replaced by `≈` for consistency

To test high gas prices you have to set some test values in the localStorage. Ping me on slack for details I will then do a short walkthough.

Run #p3


packages/web/app/components/shared/MoneySuiteWidget/MoneySuiteWidget.tsx
```diff
@@ -197,15 +197,26 @@ const MoneySuiteBase: React.FunctionComponent<
...
+          <Money
+            value={smallNumber}
+            inputFormat={bottomInputFormat || inputFormat}
+            outputFormat={outputFormat}
+            valueType={currencyTotal}
+          />
+        </AlmostEqualTo>
+      ) : (
+        <>
+          {" = "}
```
arjunu, line 211: Missed `AlmostEqualTo` here

arjunu (approved):

glupiawymowka (changes requested):
needs a fix to only allow user to continue if both checkboxes are selected.
Currently, user can continue if either of two is.

glupiawymowka (approved):

PR #4813, opened 2021-03-03
Added some dev scripts

arjunu:
Timer saver scripts to automate few things that were done locally
- script to build Android and run tests on Browserstack from local 
- script to install pods automatically when running yarn
- script to handle Cocoapods version mismatch
- script to delete and reinstall node modules


package.json
```diff
@@ -7,15 +7,16 @@
     "node": ">=12.0.0"
   },
   "scripts": {
-    "clean:node_modules": "find . -name \"node_modules\" -type d -prune -exec rm -rf '{}' +",
+    "node_modules:clean": "find . -name \"node_modules\" -type d -prune -exec rm -rf '{}' +",
+    "node_modules:reinstall": "./dev-scripts/reinstall-node-modules.sh",
```
arjunu, line 11: I'm adding a convention with `dev-scripts` directory as a directory that contains only scripts used by developers locally and never on CI

packages/wallet/package.json
```diff
@@ -47,7 +47,7 @@
     "storybook:android:release": "ENVFILE=./storybook/.env.storybook react-native run-android --variant release",
     "storybook:prepare": "rnstl",
     "contract-artifacts:generate": "node ./scripts/generate-contracts.js",
-    "postinstall": "patch-package",
```
arjunu, line 50: If I'm not wrong this `patch-package` isn't used because this `postinstall` is called only if we do `yarn` inside `packages/wallet` and not the root

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
```
arjunu, line 1: I will add an iOS script in future. iOS build takes too long. Since Android is faster I'd prefer to test this

.circleci/config.yml
```diff
@@ -624,6 +624,28 @@ workflows:
...
+          requires:
+            - wallet
+            - wallet-e2e
+      - mobile_ios_build-success:
+          type: approval
+          requires:
+            - mobile_ios_build
+      - mobile_ios_test-e2e:
+          requires:
+            - mobile_ios_build-success
```
arjunu, line 648: I forgot to add mobile e2e jobs for nightly

packages/wallet/dev-scripts/install-pods.sh
```diff
@@ -0,0 +1,37 @@
...
+compare_pods() {
+  CUR_COCOAPODS_VER=$(sed -n -e 's/^COCOAPODS: \([0-9.]*\)/\1/p' ios/Podfile.lock)
+  ENV_COCOAPODS_VER=$(pod --version)
+
+  echo "Current CocoaPods version: $CUR_COCOAPODS_VER"
+  echo "Project CocoaPods version: $ENV_COCOAPODS_VER"
+
+  if [ "$CUR_COCOAPODS_VER" != "$ENV_COCOAPODS_VER" ]; then
+    echo "Environment's CocoaPods version doesn't match with that of the project's."
+    read -r -p "Reinstall correct version?? (y/n): " confirm && [[ $confirm == [yY] || $confirm == [yY][eE][sS] ]] || exit 1
```
desfero, line 19: hm, can we branch the logic here because during `yarn install` developer would not expect any questions to be asked by command line prompt? Maybe in the case of `preinstall` we can fail the job if it's not the same and in the case of `install pods command ask that question?

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
```diff
@@ -0,0 +1,31 @@
...
+
+upload_apk_to_browserstack() {
+  curl -u "${BROWSERSTACK_USER}:${BROWSERSTACK_KEY}" -X POST "https://api-cloud.browserstack.com/app-automate/upload" -F "file=@./app-release.apk"
+}
+
+main() {
+  source ./scripts/.env
```
desfero, line 8: maybe for consistency we can keep `.env` in the root of `wallet-e2e` package?

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
```diff
@@ -0,0 +1,31 @@
...
+
+  if [ -z "$BRANCH_NAME" ] || [ -z "$BROWSERSTACK_USER" ] || [ -z "$BROWSERSTACK_KEY" ] || [ -z "$BROWSERSTACK_URI" ]; then
+    echo "Missing one or more of BRANCH_NAME, BROWSERSTACK_USER, BROWSERSTACK_KEY, BROWSERSTACK_URI in packages/wallet-e2e/scripts/.env"
+    exit 1
+  fi
+
+  cd ../wallet || exit 1
```
desfero, line 15: Why do we need `|| exit 1`?

desfero: Also maybe we can use somehow over `cd` `yarn workspace @neufund/wallet build` or something like this?

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+upload_apk_to_browserstack() {
```
desfero, line 3: let make it in a way that this function receives path to apk as a parameter

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
```diff
@@ -0,0 +1,31 @@
...
+  fi
+
+  cd ../wallet || exit 1
+  cd android || exit 1
+  echo "Building APK ... "
+  ./gradlew assembleDebug
+  cd app/build/outputs/apk/release || exit 1
+
+  echo "Uploading APK to Browserstack ... "
+  upload_apk_response=$(upload_apk_to_browserstack)
```
desfero, line 22: and to not `cd` we can pass `upload_apk_response=$(upload_apk_to_browserstack(app/build/outputs/apk/release/app-release.apk))`

dev-scripts/reinstall-node-modules.sh
```diff
@@ -0,0 +1,4 @@
+yarn workspace @neufund/wallet clean &&
```
desfero, line 1: would be cool if we can do `wsrun clean` as a new script called `clean` in root `package.json` as it can be handy in the future and here we can just call that. Then we don't have any strict dependency to any package here

desfero (changes requested):
Looks cool just some questions

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
arjunu: if build failed or if `cd app/build/outputs/apk/release` doesn't exist then cd will exit with code 1

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
arjunu: `|| exit 1` because if target directory doesn't exist it then script won't crash and continues to execute next line.

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
arjunu: done

dev-scripts/reinstall-node-modules.sh
arjunu: done

packages/wallet/dev-scripts/install-pods.sh
arjunu: updated

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
arjunu: replaced with `e2e:build:android-release`

packages/wallet-e2e/dev-scripts/local-browserstack-run.sh
arjunu: replaced with full path

packages/design-system/package.json
```diff
@@ -14,12 +14,14 @@
   "scripts": {
     "build": "NODE_ENV=production webpack --config ./webpack/webpack.config.js",
     "build:analyze": "NODE_ENV=production webpack --config webpack/webpack.config.js --analyze=static",
+    "clean": "echo Nothing to clean",
```
arjunu, line 17: in future we can add dist cleanup here

desfero: yeah exactly, that's why i wanted to have a root clean job already in place

desfero (approved):

Issue #4792, opened 2021-02-17
Input changes 


desfero:
**Background**
This ticket should contain all input-related changes that are not bugs as per se and can be done as a separate PR.

- [x] Input prefix/suffix with a lighter grey (#556877) at our Helper text font size (12px Roboto regular)
<img src='https://user-images.githubusercontent.com/7031578/108176058-cfca9c00-7101-11eb-9950-c45ead45161f.png' width='500'>
(@andywalton will upload new design to zeplin soon)

- [x] password eye icon becomes huuuge IF the page zoom changed. Changing the page size back to 100% or less doesn't fix it, only the module closing/page refresh does. True for all the password inputs.
<img src='https://user-images.githubusercontent.com/18315589/108234467-1d1c2d00-7145-11eb-898d-6813f885cb35.png' height='350'>

- [x] a left padding in the recovery phrase inputs is missing : 
dev
<img src='https://user-images.githubusercontent.com/18315589/108239561-52774980-714a-11eb-96e6-ca94fa6d8de2.png' width='500'>
vs stage 
<img src ='https://user-images.githubusercontent.com/18315589/108239954-ac780f00-714a-11eb-873a-cd78436f9745.png' width='350'>

- [x] the focus on the date input stopped to move to next one automatically once the field is filled 
<img src='https://user-images.githubusercontent.com/18315589/108355372-05977f80-71eb-11eb-96a4-0455ccb0ffa0.png' width='350'>

- [ ] use backup seed select for all select components (with search functionality / good multiselect support)
This basically means that now all select components are with superpowers so for e.g.
- KYC country select can be quickly searched for a given country 
![image](https://user-images.githubusercontent.com/7031578/110539639-cc856780-8125-11eb-841b-555f306ad9eb.png)

- Company information `categories` label now is in sync with the design system and is using a select with multiple allowed options 
![image](https://user-images.githubusercontent.com/7031578/110539310-5bde4b00-8125-11eb-83de-d54222e40d7a.png)

PR #4816, opened 2021-03-09
#4792 inputs fixes

desfero:
**Testing notes**
Run #p3


arjunu (commented):
the font size in select seems to have decreased, is that intentional?
image

IO:
image

packages/design-system/src/components/buttons/Button.tsx
```diff
@@ -6,7 +6,7 @@ import { InlineIcon } from "../icons/InlineIcon";
 import { LoadingIndicator } from "../loading-indicator";
 import { ButtonBase } from "./ButtonBase";
 
-import * as styles from "./Button.module.scss";
+import styles from "./Button.module.scss";
```
arjunu, line 9: why the change from * as to this?

arjunu:
There is a blank item in the confirm your recovery phrase dropdowns, intentional?
image

packages/design-system/src/components/buttons/Button.tsx
desfero: the newest `style-loader` outputs ES module format by default https://github.com/webpack-contrib/style-loader/blob/master/CHANGELOG.md#-breaking-changes

arjunu (approved):

PR #4828, opened 2021-03-21
[Do not merge] Accommodate voting

arjunu:
Testing notes:
- Investor with access to voting INV_ICBM_ETH_M_HAS_KYC_DUP_HAS_NEURO
- Investor without access INV_HAS_EUR_HAS_KYC
- Issuer without access ISSUER_PREVIEW
- Issuer with access to voting ISSUER_PAYOUT


packages/web/package.json
```diff
@@ -47,10 +47,10 @@
     "test:e2e": "./e2e/setup-e2e.sh",
     "test:e2e:chrome": "./e2e/setup-docker-e2e.sh",
     "test:e2e:vis-reg": "./e2e/setup-vis-reg-e2e.sh",
-    "test:e2e:cypress:vm": "NODE_ENV=production cypress run --browser chrome --headless --env NF_TAGS=\"$(yarn run --silent show-tags)\" --config testFiles=\"**/*.spec.e2e.ts\" --record --parallel --group localNodeTests --ci-build-id $CIRCLE_BUILD_NUM$(echo node)",
```
arjunu, line 50: only tests with `voting,e2e.ts` and `spec.voting.e2e.io.ts` are run

.circleci/config.yml
```diff
@@ -393,8 +393,8 @@ jobs:
             - run:
                 command: yarn workspace @neufund/web build:prod
                 no_output_timeout: 10m
-            - run:
-                command: yarn workspace @neufund/web test:e2e:vis-reg
+#            - run:
+#                command: yarn workspace @neufund/web test:e2e:vis-reg
```
arjunu, line 397: disable vis reg as it will fail

packages/web/app/modules/accommodate-voting/reducer.ts
```diff
@@ -0,0 +1,28 @@
...
+const initialState: TAccommodateVotingState = {
+  showCannotLoginModal: false,
+};
+
+export const accommodateVotingReducer: AppReducer<
+  TAccommodateVotingState,
+  typeof accommodateVotingActions
+> = (state = initialState, action): DeepReadonly<TAccommodateVotingState> => {
+  switch (action.type) {
+    case accommodateVotingActions.showCannotLoginModal.getType():
```
Moejoe90, line 19: Does it ever stop showing? 
As here we only set the state to true

Moejoe90 (commented):
One hanging question

packages/web/app/modules/accommodate-voting/reducer.ts
arjunu: User can't hide it, there's a logout button which the user if forced to click

Moejoe90 (dismissed):

desfero (dismissed):
Looks good. But we definitely should not merge that into master IMO and just keep that on a separate branch

Moejoe90 (approved):

packages/web/app/modules/accommodate-voting/reducer.ts
andywalton: @arjunu -> in this case, can we hide the email modal when showing the `showCannotLoginModal` when a user successfully recovers their account but we can't allow them to log in?

desfero:
Voting would be done on a normal platform. Closing this one but let's keep the brach alive for a while

PR #4835, opened 2021-03-29
Auto unlock wallet app

arjunu:


packages/wallet/app/App.tsx
```diff
@@ -116,7 +124,10 @@ const AppLayout: React.FunctionComponent<TLayoutProps> = ({
       });
     }
 
-    if (previousInitStatus === EInitStatus.OFFLINE && initStatus === EInitStatus.IN_PROGRESS) {
+    if (
+      [EInitStatus.NOT_STARTED, EInitStatus.OFFLINE].includes(previousInitStatus) &&
+      initStatus === EInitStatus.IN_PROGRESS
```
arjunu, line 129: there is still sometimes a white screen because of the delay between app opening and this code executing.

arjunu: ok I updated `react-native-bootsplash` and the white screen seems to be gone now

packages/wallet/app/App.tsx
```diff
@@ -104,21 +106,33 @@ const AppLayout: React.FunctionComponent<TLayoutProps> = ({
...
+    };
+  }, [start, reset]);
 
   React.useEffect(() => {
     if (previousInitStatus === EInitStatus.IN_PROGRESS) {
       // wait for all interactions to finish before showing app
       // otherwise for a moment a white screen will appear for a user
       // eslint-disable-next-line @typescript-eslint/no-floating-promises
       InteractionManager.runAfterInteractions(() => {
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
arjunu, line 123: because now `show` and `hide` return promises

packages/wallet/app/App.tsx
```diff
@@ -104,21 +106,33 @@ const AppLayout: React.FunctionComponent<TLayoutProps> = ({
 
   React.useEffect(() => {
     start();
-  }, [start]);
+
+    return () => {
+      // on Android when app is reopened after back button minimize, App remounts and starts init
+      // so on unmount reset init status flag for sagas
+      reset();
```
desfero, line 113: cool

packages/wallet/app/modules/auth/sagas.ts
```diff
@@ -1,26 +1,25 @@
...
-  assertNever,
+  toEthereumPrivateKey,
 } from "@neufund/shared-utils";
 // eslint-disable-next-line react-native/split-platform-components
 import { ToastAndroid } from "react-native";
-import Config from "react-native-config";
 
 import { biometricsModuleApi } from "modules/biometrics/module";
-import { walletEthModuleApi, EWalletExistenceStatus } from "modules/eth/module";
+import { EWalletExistenceStatus, walletEthModuleApi } from "modules/eth/module";
```
desfero, line 14: hm, i'm curious why it sorted that now and not before?

packages/wallet/app/modules/auth/sagas.ts
```diff
@@ -42,23 +41,11 @@ export function* trySignInExistingAccount(): SagaGenerator<void> {
...
-          break;
-
-        case "auto":
-        default:
-          yield* call(signInExistingAccount);
-          break;
-      }
-
+      yield* call(allowToUnlockExistingAccount);
+      yield put(authActions.unlockAccount());
```
desfero, line 45: let's add a comment why we have both `yield* call(allowToUnlockExistingAccount);` and `yield put(authActions.unlockAccount());` as i was not sure initially why that was needed

packages/wallet/storybook/index.tsx
```diff
@@ -30,6 +30,7 @@ const Storybook: React.FunctionComponent = () => {
     InteractionManager.runAfterInteractions(() => {
       // we do have custom logic when to hide splash screen for the normal flow
       // for storybook we can hide just when we have UI ready
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
       RNBootSplash.hide();
```
desfero, line 34: let's also use `async/await` here

desfero (changes requested):
Some small comments

packages/wallet/storybook/index.tsx
arjunu: done

packages/wallet/app/modules/auth/sagas.ts
arjunu: strange indeed

packages/wallet/app/App.tsx
arjunu: done

packages/wallet/app/modules/auth/sagas.ts
arjunu: done

desfero (approved):

PR #4843, opened 2021-04-06
Add messages to nonNullable

desfero:
**Tech notes**
I found that it's pretty annoying to have errors with just `nonNullable value expected` as then it's hard to understand what exactly is missing


packages/web/app/modules/shareholder-resolutions-voting-view/selectors.ts
```diff
@@ -27,7 +27,10 @@ export const selectProposalId = (state: TAppGlobalState) =>
 
 export const selectProposalShareholder = (state: TAppGlobalState) => {
   const { proposalId } = state.shareholderResolutionsVotingView;
-  const shareholderAddress = nonNullable(selectUserId(state));
+  const shareholderAddress = nonNullable(
+    selectUserId(state),
+    "`selectProposalShareholder` requires `shareholderAddress`",
```
arjunu, line 32:
```suggestion
    "`selectProposalShareholder` requires `userId`",
```

packages/web/app/modules/nominee-flow/sagas.ts
```diff
@@ -585,7 +585,10 @@ export function* setActiveNomineeEto({ logger }: TGlobalDependencies): Generator
           ),
         );
       } else {
-        const firstEto = nonNullable(Object.values(etos)[0]);
+        const firstEto = nonNullable(
+          Object.values(etos)[0],
+          "`setActiveNomineeEto` requires `firstEto`",
```
arjunu, line 590:
```suggestion
          "`setActiveNomineeEto` requires `etos[0]`", 
```
since firstEto is not a variable that exists

arjunu (approved):

arjunu (approved):

PR #4844, opened 2021-04-06
Wallet connect wallet e2e

arjunu:
**Testing notes**
There is one e2e test failing in web app but that's a separate known issue


packages/wallet-e2e/tests/specs/assertUtils.specUtils.ts
```diff
@@ -17,3 +19,9 @@ export const assertImportAccount = (recoveryPhraseOrPrivateKey: string) => {
 
   homeScreen.waitForIsShown();
 };
+
+export const assertLogout = () => {
```
arjunu, line 23: thought of using this, but it is unused now. Didn't remove in case we wanna use it in future

packages/wallet-e2e/tests/specs/wallet-connect/disconnect.e2e.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+      disconnected = true;
+    });
+
+    appAuthRouterTabs.goToProfile();
+    profileScreen.openWalletConnectSession();
+
+    walletConnectSessionScreen.waitForIsShown();
+    walletConnectSessionScreen.cancelSession();
+
+    browser.waitUntil(() => disconnected, {
```
arjunu, line 24: A bit stupid but it's simple and works reliably. Easy to read code also. Webdriver polls until this condition is true or until timeout

packages/wallet-e2e/tests/specs/wallet-connect/killSession.e2e.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+import { appAuthRouterTabs } from "../../screenObjects/tabs.specUtils";
+import { walletConnectSessionScreen } from "../../screenObjects/walletConnectSession.screen.specUtils";
+import { assertConnectionApprove, assertSignMessage } from "./walletConnectAssertUtils.specUtils";
+
+describe("WalletConnect kill session", () => {
+  it("should disconnect when session is killed by client", () => {
+    let connector = assertConnectionApprove();
+    connector = assertSignMessage(connector);
+
+    browser.call(() => connector.killSession());
```
arjunu, line 12: `browser.call` waits for promise to resolve

KonradFilipiak (dismissed):

packages/wallet-e2e/tests/specs/wallet-connect/walletConnectAssertUtils.specUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+  signMessageSigner.approve();
+
+  browser.waitUntil(() => signMessageApproved, {
+    timeout: DEFAULT_TIMEOUT,
+    timeoutMsg: "Expect `signMessageApproved` to be true",
+  });
+
+  return connector;
```
desfero, line 103: not sure we have to return connector and then overwrite it inside each test given it has not changed as it's a singleton instance. What do you think?

packages/wallet-e2e/tests/specs/wallet-connect/killSession.e2e.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { nativeAlert } from "../../nativeObjects/alert";
+import { profileScreen } from "../../screenObjects/profile.screen.specUtils";
+import { appAuthRouterTabs } from "../../screenObjects/tabs.specUtils";
+import { walletConnectSessionScreen } from "../../screenObjects/walletConnectSession.screen.specUtils";
+import { assertConnectionApprove, assertSignMessage } from "./walletConnectAssertUtils.specUtils";
+
+describe("WalletConnect kill session", () => {
+  it("should disconnect when session is killed by client", () => {
+    let connector = assertConnectionApprove();
+    connector = assertSignMessage(connector);
```
desfero, line 10: I would also hide these two under one call given in the future we have to refactor wallet-connect implementation to show modal just once

packages/wallet-e2e/tests/specs/wallet-connect/transaction.e2e.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      timeoutMsg: "Expect `txApproved` to be true",
+    });
+
+    connector
+      .sendCustomRequest(request)
+      .then(() => {
+        txApproved = true;
+      })
+      .catch(() => {
+        txApproved = false;
```
desfero, line 54: do we need that catch here? There is no way from the code above for `txApprove` to be `true`

desfero (changes requested):
Can we also assert 

1. Allert that appears after session was disconnected from the user side (web app side)
2. Wallet-connect modal changes in the wallet app (profile page)

packages/wallet-e2e/tests/specs/wallet-connect/transaction.e2e.spec.ts
arjunu: eslint/typescript was complaining :D

arjunu: Removed and suppressed the error

packages/wallet-e2e/tests/specs/wallet-connect/killSession.e2e.spec.ts
arjunu: combined

packages/wallet-e2e/tests/specs/wallet-connect/walletConnectAssertUtils.specUtils.ts
arjunu: ok now it looks like this

``` 
const assertWCConnection = () => assertSignMessage(assertConnectionApprove());

// test code
const connector = assertWCConnection();
```

desfero (approved):

PR #4853, opened 2021-04-10
Fix keyboard avoiding view

arjunu:
Handles both https://github.com/Neufund/platform-frontend/issues/4722 and https://github.com/Neufund/platform-frontend/issues/4718


desfero (changes requested):
Some changes request on slack

packages/wallet/app/components/shared/Screen.tsx
```diff
@@ -23,6 +26,9 @@ type TSafeAreaScreenExternalProps = {
 } & TExternalCommonProps &
   React.ComponentProps<typeof Animated.ScrollView>;
 
+const getKeyboardVerticalOffset = (headerHeight: number, insetBottom: number) =>
+  headerHeight + (isIOS ? insetBottom + spacingStyles.p4.padding : spacingStyles.p4.padding * 2);
```
desfero, line 30: not sure about the math here but from what i see.
if stack `mode` is `modal` then we calculate `keyboardVerticalOffset` with `headerHeight + (isIOS ? insetTop + 6 : 0)`
if stack mode is normal we just use `headerHeight` as an inset (i have no idea how to get `mode` type from `react-navigation` so for now we can support `modal` mode with some comment

desfero: image

desfero: and 6 px is for that space in between inset and header

packages/wallet/app/components/shared/Screen.tsx
```diff
@@ -81,15 +88,39 @@ const Screen: React.FunctionComponent<TScreenExternalProps> = ({
...
+
+  const [keyboardAvoidingViewEnabled, setKeyboardAvoidingViewEnabled] = React.useState(true);
+
+  /**
+   * On iOS KeyboardAvoidingView pushes all stacked screen to top when keyboard opens
+   * https://github.com/facebook/react-native/issues/29614
+   * https://github.com/Neufund/platform-frontend/issues/4718
+   */
+  React.useEffect(() => {
+    const unsubscribeFocus = navigation.addListener("focus", () => {
```
desfero, line 101: Wny not `https://reactnavigation.org/docs/use-focus-effect` effect?

packages/wallet/app/components/shared/Screen.tsx
```diff
@@ -101,6 +132,16 @@ const Screen: React.FunctionComponent<TScreenExternalProps> = ({
         {children}
       </Animated.ScrollView>
     </KeyboardAvoidingView>
+  ) : (
+    <Animated.ScrollView
```
desfero, line 136: to not duplicate this can be kept as a variable

packages/wallet/app/components/shared/Screen.tsx
```diff
@@ -81,15 +88,39 @@ const Screen: React.FunctionComponent<TScreenExternalProps> = ({
...
+    return () => {
+      unsubscribeFocus();
+      unsubscribeBlur();
+    };
+  }, [navigation]);
 
   const scrollMode = overScrollMode ?? (bounces ? "never" : undefined);
 
-  return (
+  return keyboardAvoidingViewEnabled ? (
```
desfero, line 117: we need to duplicate the same in normal `Screen` component, right?

desfero (changes requested):

packages/wallet/app/components/shared/Screen.tsx
arjunu: done

packages/wallet/app/components/shared/Screen.tsx
arjunu: yes but I can't test it since we don't have a screen that uses `SafeAreaScreen` and inputs. Ok I will add a TODO

packages/wallet/app/components/shared/Screen.tsx
arjunu: didn't know about it, used it

packages/wallet/app/components/shared/Screen.tsx
arjunu: updated

desfero (approved):

PR #4854, opened 2021-04-11
#4845 Light wallet export

desfero:
**Testing notes**
I had to refactor our header menu components so we need to make sure they still operate as normal.
Given QRCode design has changed a little bit (background/margin and qr color) QR code during ETH `Recieve` flow (accessible from `Wallet` page) has changed to match the newest design.

**Before merge**
- [ ] Create a PR to devops with the following envs (remove NF_PORTFOLIO_PAGE_VISIBLE, add NF_LIGHT_WALLET_IMPORT_ENABLED)


packages/design-system/.storybook/main.js
```diff
@@ -53,6 +52,9 @@ module.exports = {
                 loader: "css-loader",
                 options: {
                   importLoaders: 2,
+                  modules: {
+                    compileType: "icss",
```
desfero, line 56: This will allow exporting scss variables to js

packages/design-system/package.json
```diff
@@ -118,6 +119,7 @@
     "react-virtualized": "^9.22.3",
     "react-virtualized-select": "^3.1.3",
     "sanitize-html": "^2.1.0",
-    "recompose": "^0.30.0"
+    "recompose": "^0.30.0",
+    "qrcode": "^1.4.4"
```
desfero, line 123: QRCode can be a part of design-system to reduce web complexity

packages/design-system/src/components/shared/Image.tsx
```diff
@@ -29,13 +30,22 @@ const srcSetToString = (srcSet: ISrcSet) =>
...
+    () =>
+      // We can omit `1x` given it's set as `src` attribute
+      srcSetToString(omit(srcSet, "1x")),
+    [srcSet],
+  );
+
+  return (
+    <img
+      src={srcSet["1x"]}
+      srcSet={srcSetString}
```
desfero, line 44: Otherwise we duplicate src and srcSet 1x and given we often use base64 html is growing in size

packages/design-system/.storybook/main.js
```diff
@@ -125,13 +127,8 @@ module.exports = {
         __DEV__: false,
       }),
     );
-    config.resolve.extensions.push(".ts", ".tsx");
 
-    config.plugins.push(
-      new DefinePlugin({
-        __DEV__: false,
```
arjunu, line 132: lol duplicate

packages/design-system/src/components/shared/QRCodeImage.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const QRCodeImage: React.FunctionComponent<TExternalProps & TAllowedImageProps> = ({
+  uri,
+  ...rest
+}) => {
+  const [svgBase64, setSvg] = React.useState<string>();
+
+  React.useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    QRCode.toString(uri, {
```
arjunu, line 38: memo?

packages/design-system/.storybook/main.js
arjunu: cooool

packages/web/app/components/layouts/menus/menuItems.tsx
```diff
@@ -13,30 +13,48 @@ export const menuSeparatorData = (key: string): TMenuEntry => ({
...
+      key: "logout",
+      svgString: logoutIcon,
+      menuName: <FormattedMessage id="menu.logout" />,
+      onClick: () => logout(),
+      "data-test-id": "menu-logout-button",
+    },
+  ];
+
+  if (walletType === EWalletType.LIGHT && process.env.NF_LIGHT_WALLET_IMPORT_ENABLED === "1") {
+    return data.concat({
```
arjunu, line 47: I think we should put this before logout as logout is usually the last option in a dropdown

packages/web/intl/locales/en-en.json
```diff
@@ -1007,10 +1021,11 @@
   "menu.eto-page": "ETO Page",
   "menu.governance-page": "Governance",
   "menu.help": "Help",
+  "menu.light-wallet-export": "Join mobile app beta",
```
arjunu, line 1024: The page title is "Join _the_ mobile app beta"

arjunu: I think it's ok

packages/web/intl/locales/en-en.json
```diff
@@ -997,6 +1002,15 @@
   "kyc.success.go-to-additional-documents": "Or add additional documents",
   "kyc.success.go-to-dashboard": "Go to Dashboard",
   "kyc.success.title": "Congratulations!",
+  "light-wallet-export.description": "Please follow the steps below to connect you Neufund account with our beta mobile apps.",
+  "light-wallet-export.qr-code-alt": "Light wallet export QR code",
+  "light-wallet-export.step": "Step",
+  "light-wallet-export.step1": "Make sure you’ve subscribed to the mobile app beta {subscriptionLink}.",
+  "light-wallet-export.step2": "Thereafter you will receive an email titled “Welcome to the Neufund Beta”. This email contains instructions to download and install our iOS and Android Beta apps download.",
```
arjunu, line 1009: I think Beta should be small be

packages/design-system/src/components/shared/QRCodeImage.tsx
```diff
@@ -0,0 +1,51 @@
...
+      type: "svg",
+      color: { dark: variables.neuSpruce, light: variables.neuPaleGreyTwo },
+    }).then((svg: string) => {
+      setSvg(btoa(svg));
+    });
```
arjunu, line 43:
```
.catch(err => {
    ...
  })
  ```

File: packages/shared-modules/src/lib/customURIs.ts [hunk 10]
```diff
@@ -0,0 +1,6 @@
+/**
+ * Generates EIP-831 URI Format for Ethereum transfer transactions
+ */
+import {EthereumAddressWithChecksum} from "@neufund/shared-utils";
```
arjunu: looks like this file isn't formatted? there would be spaces between the brackets right?

packages/design-system/src/components/shared/QRCodeImage.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type TExternalProps = {
+  alt: string;
+  uri: string;
+};
+
+/**
+ * An empty qr code to show as an initial placeholder
+ */
+const QRCodePlaceholder: React.FunctionComponent<CommonHtmlProps> = props => (
```
arjunu, line 19: what is this svg supposed to be? I see blank

packages/design-system/src/components/shared/QRCodeImage.tsx
```diff
@@ -0,0 +1,51 @@
+import * as QRCode from "qrcode";
```
arjunu, line 1: this is for the future, the shared directory here doesn't make sense. All components in design-system are shared :P

desfero: yeah, i have the same feelings and was looking for a better place too :D

packages/design-system/src/components/shared/QRCodeImage.tsx
desfero: it's supposed to be blank. While qr code is calculated we have to show something and it should also take whole space as qr code to avoid content jumping

packages/design-system/src/components/shared/QRCodeImage.tsx
desfero: not sure why given `useEffect` will only rerun on uri change. Maybe i'm missing something?

packages/web/app/components/layouts/menus/menuItems.tsx
desfero: Good point

packages/design-system/src/components/shared/QRCodeImage.tsx
desfero: done, now we log. Not sure whether we catch anything on UI as it's a really edge case that should never happen

packages/web/intl/locales/en-en.json
desfero: hm, i would say it should be `Beta` given it's a heading (https://www.webucator.com/how-to/how-capitalize-headings-titles.cfm)

packages/web/app/components/layouts/menus/menuItems.tsx
```diff
@@ -17,8 +18,8 @@ export const accountMenuData = (
   actionRequired: boolean,
   logout: Function,
   walletType: EWalletType,
-): TMenuEntry[] => {
-  const data: TMenuEntry[] = [
+): TMenuEntry[] =>
+  compact([
```
arjunu, line 22: cool

packages/design-system/src/components/shared/QRCodeImage.tsx
arjunu: right

arjunu (dismissed):

arjunu (approved):

PR #4858, opened 2021-04-12
Ignore wallet app lock when permission dialog opens (Android)

arjunu:


packages/wallet/app/modules/init/sagas.ts
```diff
@@ -85,18 +88,48 @@ function* initStartSaga(): SagaGenerator<void> {
   }
 }
 
+/**
+ * Special cases where we shouldn't lock app on minimize
+ * temporary situation to handle Android permissions
```
arjunu, line 93: App also locks when we share address from profile. I assume this is fine since  to share we need to open another app anyways

desfero: yeah, we can leave it as it is

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -85,18 +88,48 @@ function* initStartSaga(): SagaGenerator<void> {
...
+    permissions: permissionsModuleApi.symbols.permissions,
+    logger: coreModuleApi.symbols.logger,
+  });
+
+  const currentRoute = navigationRef.current?.getCurrentRoute()?.name;
+
+  /**
+   * On Android, when permission dialog opens the app goes to background so don't lock
+   */
+  if (isAndroid && currentRoute === EAppRoutes.qrCode) {
```
desfero, line 106: i would add a constant `ALLOWED_UNLOCKED_SCREENS` as an array with screen names

packages/wallet/app/modules/init/sagas.ts
```diff
@@ -85,18 +88,48 @@ function* initStartSaga(): SagaGenerator<void> {
...
+    const cameraPermissionStatus = yield* call([permissions, "check"], PERMISSIONS.ANDROID.CAMERA);
```
desfero, line 107: let's simplify that by just returning `true` if that's a QR code screen? What do you think?

desfero (changes requested):
Some questions

desfero (approved):
Some questions

KonradFilipiak (approved):

PR #4863, opened 2021-04-13
Fix tab switch crash

arjunu:


packages/wallet/app/components/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -50,7 +50,17 @@ export const HomeScreen = compose<TStateProps, UnknownObject>(
...
   onLifecycle({
-    onMount: dispatch => dispatch(homeScreenModuleApi.actions.loadHomeScreen()),
-    onFocus: dispatch => dispatch(homeScreenModuleApi.actions.refreshHomeScreen()),
+    onMount: (dispatch, { screenState }) =>
+      dispatch(
+        screenState === EScreenState.READY
+          ? homeScreenModuleApi.actions.refreshHomeScreen()
+          : homeScreenModuleApi.actions.loadHomeScreen(),
+      ),
+    onFocus: (dispatch, { screenState }) =>
```
arjunu, line 59: this fixes a crash when you switch tab while it is still loading and when you come back on Focus the `screenState` would be  `EScreenState.REFRESHING` and selector will crash

packages/wallet/app/components/screens/HomeScreen/HomeScreen.tsx
```diff
@@ -50,7 +50,17 @@ export const HomeScreen = compose<TStateProps, UnknownObject>(
     stateToProps: state => homeScreenModuleApi.selectors.selectHomeScreenData(state),
   }),
   onLifecycle({
-    onMount: dispatch => dispatch(homeScreenModuleApi.actions.loadHomeScreen()),
-    onFocus: dispatch => dispatch(homeScreenModuleApi.actions.refreshHomeScreen()),
+    onMount: (dispatch, { screenState }) =>
```
arjunu, line 53: this changes showing skeleton animation every time we switch tabs, the data is already in the reducer so it is a bit annoying to see skeleton

desfero (changes requested):
Discussed on slack, we will fix that on sagas level without doing `if`s in components

desfero (approved):

KonradFilipiak (approved):

PR #4869, opened 2021-04-14
Remove unnecessary waits from e2e tests

KonradFilipiak:


packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
```diff
@@ -68,9 +67,13 @@ describe("Eto Company Information Rich Text Editor", () => {
...
-  it("should allow to add images #eto #p3 #flaky", () => {
-    cy.get(tid("companyPreviewCardBanner.uploaded.remove-confirm")).click();
-    cy.get(tid("companyPreviewCardBanner.uploaded.remove")).click();
+  it("should allow to add images #eto #p3", () => {
+    cy.get("body").then($body => {
+      if ($body.find(tid("companyPreviewCardBanner.uploaded.remove-confirm")).length > 0) {
+        cy.get(tid("companyPreviewCardBanner.uploaded.remove-confirm")).click();
+        cy.get(tid("companyPreviewCardBanner.uploaded.remove")).click();
+      }
+    });
```
KonradFilipiak, line 76: It was one of the reasons this test was flaky - it expected another test to upload an image and needed to remove it. If the image was not uploaded for some reason (the other test didn't run, it failed, BE didn't save the image etc.) then this test would fail with "Element not found"

PR #4871, opened 2021-04-16
Remove waits e2e

desfero:


packages/web/app/test-e2e/investor/invest/InvestWithChange.spec.e2e.io.ts
```diff
@@ -18,15 +18,13 @@ describe("Invest with change", () => {
     cy.get(tid("investment-type.selector.ICBM_NEUR")).check({ force: true });
     cy.get(tid("invest-modal-eur-field")).clear().type("500");
 
-    cy.wait(1000);
-
-    cy.get(tid("invest-modal-invest-now-button")).click();
+    cy.get(tid("invest-modal-invest-now-button")).should("not.be.disabled").click();
```
desfero, line 21: but the `enabled` state should be awaited by `cypress`?

packages/web/app/test-e2e/issuer/eto/Campaigning.spec.e2e.io.ts
```diff
@@ -48,17 +48,15 @@ describe("Eto campaigning state", () => {
 
         cy.visit(etoPublicViewByIdLinkLegacy(ETO_ID));
 
-        cy.wait(5000); //let the store get the ETO data, otherwise this place is flaky
-
-        cy.get(tid("open-pledge-modal-join")).should("exist");
+        cy.get(tid("open-pledge-modal-join")).should("be.visible");
```
desfero, line 51: so what's the problem here given we await `visibility` now?

packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
```diff
@@ -86,16 +89,14 @@ describe("Eto Company Information Rich Text Editor", () => {
...
 
     assertIssuerDashboard();
 
     goToEtoPreview();
 
-    cy.get(`${tid("eto-view-company-description")} img`).should($field => {
-      expect($field.attr("src")).to.contain("https://documents.neufund.io/");
-    });
+    cy.get(`${tid("eto-view-company-description")} img`)
+      .should("have.attr", "src")
```
desfero, line 99: why just not `should("have.attr", "src", "https://documents.neufund.io/")`?

packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
```diff
@@ -68,9 +67,13 @@ describe("Eto Company Information Rich Text Editor", () => {
     });
   });
 
-  it("should allow to add images #eto #p3 #flaky", () => {
-    cy.get(tid("companyPreviewCardBanner.uploaded.remove-confirm")).click();
-    cy.get(tid("companyPreviewCardBanner.uploaded.remove")).click();
+  it("should allow to add images #eto #p3", () => {
+    cy.get("body").then($body => {
+      if ($body.find(tid("companyPreviewCardBanner.uploaded.remove-confirm")).length > 0) {
```
desfero, line 72: i'm really against having `if` inside e2e tests as then the test is not pure/idempotent. 
The case here is that some tests modify fixtures/shared accounts so locally they can be run just once without local backend. Would be good to have some fixture cloning mechanism but that's not very easy from the backend/eth side unfortunatly.

KonradFilipiak: This was one of the reasons the test was flaky - it expects another test to upload an image and then it starts by removing this image. But if the other test fails, doesn't start, or the image is not uploaded for any other reason then this test will fail with `Element not found`
I changed it, so it first checks if the image is part of the page. If it is then it removes it, and if it isn't then it just continues with the test

packages/web/app/test-e2e/issuer/eto/Campaigning.spec.e2e.io.ts
KonradFilipiak: `exist` only waits for the element to be part of the DOM. This button is part of the DOM almost immediately but without ETO data it's not visible to the user - that's why the test was flaky and `wait` was needed, cypress couldn't click it
`be.visible` waits for the element to be visible for the user- so in this case it waits for ETO data automatically

packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
KonradFilipiak: This way it would assert that `src` value equals "https://documents.neufund.io/" instead of asserting that it contains given string
Actual value of `src` is something like `https://documents.neufund.io/some-random-uuid.png` and I can't know the uuid

desfero: ah, i see. But i see that maybe we can get rid of that assertion on the domain url? What if we deploy `document` or use some other domain to run tests on then this test will fail given everything is still working properly just images are uploaded to a different domain. What's the proper approach in that case?

packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
desfero: ok, i see the problem now. So then it's a bug in tests because they should not be dependent on each other. I would say we can modify the test a little bit to create an empty account for every test

```
Index: packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts b/packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
--- a/packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts	(revision ea62ae4155c9a73e59af9288bc4c45836a245a4f)
+++ b/packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts	(date 1618904489441)
@@ -7,21 +7,14 @@
 import { aboutFormRequired, aboutFormSubmit } from "../../fixtures";
 
 describe("Eto Company Information Rich Text Editor", () => {
-  before(() => {
-    createAndLoginNewUser({ type: "issuer", kyc: "business" });
-    cy.saveLocalStorage();
-  });
-
   beforeEach(() => {
-    cy.restoreLocalStorage();
+    createAndLoginNewUser({ type: "issuer", kyc: "business" });
     goToCompanyInformation();
   });
 
   it("should render field as rich-text #eto #p3", () => {
     fillForm(
-      {
-        ...aboutFormRequired,
-      },
+      aboutFormRequired,
       { submit: false },
     );
 
@@ -76,9 +69,7 @@
     });
 
     fillForm(
-      {
-        ...aboutFormRequired,
-      },
+      aboutFormRequired,
       { submit: false },
     );
 ````

KonradFilipiak: I am just going to delete the before block and move all the setup to beforeEach - I assumed it's important to have only one user created, so I didn't change it

packages/web/app/test-e2e/issuer/eto/forms/components/RichTextEditor.spec.e2e.io.ts
KonradFilipiak: We could keep the domain as an env, but in general I agree that asserting it is not necessary - we should have a unit test keeping track of where we upload stuff. Here we just need to check that the image we uploaded before is now visible to the user, so `cy.get('${tid("eto-view-company-description")} img').should('be.visible');`

Moejoe90 (approved):

PR #4872, opened 2021-04-16
Migrate to `fast` gas price

desfero:


packages/web/app/modules/tx/sender/selectors.ts
```diff
@@ -31,5 +32,7 @@ export const selectTxGasCostEthUlps = (state: TAppGlobalState): string => {
 export const selectTxGasCostEurUlps = (state: TAppGlobalState): string =>
   multiplyBigNumbers([selectEtherPriceEur(state), selectTxGasCostEthUlps(state)]);
 
-export const selectStandardGasPriceWithOverHead = (state: TAppGlobalState): string =>
-  calculateGasPriceWithOverhead(gasApi.selectors.selectStandardGasPrice(state));
+export const selectCurrentGasPriceWithOverHead = createSelector(
```
desfero, line 35: using `reselect` here is something we should do from the begging given `calculateGasPriceWithOverhead` is doing some math operations

arjunu (approved):

PR #4876, opened 2021-04-20
Change tab bar's icons to new version

Schabaani:


packages/wallet/app/router/AppAuthRouterTabs.tsx
```diff
@@ -15,35 +15,39 @@ import { EAppRoutes } from "./appRoutes";
...
-type TBarProps = { color?: string };
+type TBarProps = { color?: string; focused: boolean };
 
 const tabConfig = [
   {
     name: "Home",
     route: EAppRoutes.home,
     component: HomeScreen,
     icon: EIconType.HOME,
+    fillIcon: EIconType.HOME_FILLED,
```
desfero, line 26: i would say `fillIcon` is not a proper name given we can in the future replace filled icons with some other variant. 
We should rename that for e.g. to `focusedIcon` as that's describing the logic over ui style

packages/wallet/app/router/AppAuthRouterTabs.tsx
```diff
@@ -58,14 +62,16 @@ const AppAuthRouterTabs: React.FunctionComponent = () => {
...
       }}
     >
-      {tabConfig.map(({ route, name, component, icon, testID }) => (
+      {tabConfig.map(({ route, name, component, icon, fillIcon, testID }) => (
         <Tab.Screen
           key={route}
           name={name}
           component={component}
           options={{
             tabBarLabel: ({ color }: TBarProps) => <MenuLabel style={{ color }}>{name}</MenuLabel>,
```
desfero, line 71: We should also sync label color with the icon color from what i see in designs 
https://app.zeplin.io/project/5e6f48a43076c61012fd3793/screen/6077e8a5d24f574453fb8f53

desfero (changes requested):
image
Can you please sync with @andywalton the icon on the profile screen. We use the same icon as in tabs but now after change, it looks to me like a Pokeball

desfero (approved):

ozanmanav-ng (approved):

Issue #4878, opened 2021-04-20
Change Country Banned List


Moejoe90:
```Afghanistan
Bosnia and Herzegovina
Democratic People’s Republic of Korea
Ethiopia
Guyana
Iran
Iraq
Laos
Pakistan
Sri Lanka
Syria
Trinidad and Tobago
Tunisia
Uganda
Yemen
Vanuatu
```

**Testing notes**
All the countries above should be blocked through the KYC process. 
An e2e test to understand the flow:
https://github.com/Neufund/platform-frontend/blob/49dbdaab03c22a3863e2b36f64bfffc987a5192c/packages/web/app/test-e2e/investor/kyc/KYC.spec.e2e.io.ts#L21-L30

PR #4892, opened 2021-04-22
#4878 Change KYC restricted countries

desfero:
Run #p3


arjunu (approved):

packages/web/app/test-e2e/investor/kyc/IDnow.spec.e2e.io.ts
```diff
@@ -66,7 +66,9 @@ const fillAndAssert = (personalData: TFormFixture, addressData: TFormFixture, is
   assertOutsourcedKycWidgetStatus();
 };
 
-describe("IDNow", () => {
+// IDNow was disabled on the backend side
+// We are awaiting business decision whether we remove IDNOW support for codebase
+describe.skip("IDNow", () => {
```
arjunu, line 71: should we not change the test to check ID doesn't show up?

desfero: let's push this one as it is given it's a high priority. To test whether idnow doesn't show up we have to modify code to read env variables from some global variable. Will add that to my todo list :D

Schabaani (approved):

ozanmanav (approved):

Issue #4891, opened 2021-04-22
Migrate `tx` module to `shared-modules`


desfero:
**Background**
Given tx module is a crucial part of adding actions to our wallet app we have to migrate that to `shared-modules`

**Technical notes**

**Testing notes**

PR #4899, opened 2021-04-23
#4891 Monitor TX migration to `shared-modules`

desfero:
Run #p3


packages/shared-modules/src/modules/eto/sagas.spec.ts
```diff
@@ -139,6 +145,18 @@ class DummyEthManager implements IEthManager {
   getWalletAddress(): Promise<EthereumAddressWithChecksum> {
     return Promise.resolve(toEthereumChecksumAddress("0x7824e49353BD72E20B61717cf82a06a4EEE209e8"));
   }
+
+  getBlockNumber(): Promise<number> {
+    throw new Error("Not implemented");
```
desfero, line 150: Should be done as a part of #4414

packages/shared-modules/src/modules/tx/monitor/sagas.ts
```diff
@@ -0,0 +1,220 @@
...
+    try {
+      yield* call(updatePendingTxs);
+    } catch (e) {
+      logger.error(e, "Error getting pending txs");
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+    // TODO: Add block watcher
+    yield delay(10000);
+    // yield take(actions.web3.newBlockArrived);
```
desfero, line 202: will do a separate issue to add block tracker to `shared-modules`

packages/shared-modules/src/modules/tx/monitor/sagas.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  const isValid = schema?.isValidSync(pendingTransaction.transactionAdditionalData);
+
+  if (!isValid) {
+    throw new SchemaMismatchError(pendingTransaction.transactionType);
+  }
+}
+
+export function* txMonitorSagas(): SagaGenerator<void> {
+  yield* takeLatestUntil(authModuleAPI.actions.setUser, authModuleAPI.actions.reset, txMonitor);
```
arjunu, line 219: move to web app init module

arjunu (changes requested):
one comment

packages/shared-modules/src/modules/tx/core/sagas/getTransactionOrThrow.ts
```diff
@@ -0,0 +1,50 @@
+import { call, SagaGenerator } from "@neufund/sagas";
+import { TEthereumTxHash } from "@neufund/shared-utils";
+
+import { neuGetBindings } from "../../../../utils";
```
ozanmanav-ng, line 4: Shouldn't we use alias for this ?

desfero: good point, unfortunately we don't have yet setup for aliases in `shared-modules`

packages/shared-modules/src/modules/tx/monitor/sagas.ts
desfero: done

arjunu (approved):

PR #4902, opened 2021-04-26
Upgrade cypress to 7.1.0

KonradFilipiak:
Upgrades Cypress version to 7.1.0


packages/web/package.json
```diff
@@ -128,7 +128,7 @@
     "concurrently": "^4.1.0",
     "copy-webpack-plugin": "^5.1.0",
     "css-loader": "^3.3.0",
-    "cypress": "5.6.0",
+    "cypress": "7.1.0",
```
desfero, line 131: let's maybe update also `@cypress/webpack-preprocessor` and `cypres-terminal-report` given they both are cypress related

desfero (changes requested):
Also let's make sure whole pipeline is green with #p3 tests running too

desfero (approved):

PR #4941, opened 2021-05-07
Fix onboarding spec file (flaky test)

KonradFilipiak:


packages/web/app/test-e2e/utils/forms.ts
```diff
@@ -61,8 +61,19 @@ export const fillSelectField = (key: string, value: string, parent: string = "bo
       // we need to force clear/type as at the begging select placeholder covers input
       .clear({ force: true })
       .type(value, { force: true })
-      .wait(100)
-      .type("{enter}");
+      .wait(100);
```
desfero, line 64: Seems we can remove `wait` here, right?

packages/web/app/test-e2e/utils/forms.ts
```diff
@@ -61,8 +61,19 @@ export const fillSelectField = (key: string, value: string, parent: string = "bo
...
+
+    cy.get('[class~="VirtualizedSelectOption"]')
```
desfero, line 66: Why just not?
```
 cy.contains('.VirtualizedSelectOption', new RegExp(`^${value}$`))
      .click()
```

desfero (changes requested):

PR #4951, opened 2021-05-12
Exit claim e2e

KonradFilipiak:


packages/web/app/test-e2e/investor/portfolio/payout/ExitClaimPayout.spec.e2e.ts
```diff
@@ -0,0 +1,47 @@
...
+      cy.get(tid(`modals.portfolio.portfolio-assets.claim-payout-${ETO_ID}`)).click();
+
+      cy.get(tid("exit-claim-flow.summary.confirm")).click();
+
+      confirmAccessModal();
+
+      cy.get(tid("modals.shared.tx-success.modal")).should("exist");
+
+      closeModal();
+      cy.reload();
```
desfero, line 26: Do we need to reload? We should never reload a SPA page as that's an anti-pattern and means there is a bug on the frontend.

desfero (changes requested):

packages/web/app/test-e2e/investor/portfolio/payout/ExitClaimPayout.spec.e2e.ts
KonradFilipiak: You're right I've added that and didn't give a second thought
It seems that there's a bug, I've added a comment and created issue on zenhub - https://app.zenhub.com/workspaces/nf-platform-5a620eae4604395352a06fdb/board?filterLogic=any&repos=116978635,120639497

desfero: Cool, can you please also add a tech note to the above ticket to remove `cy.reload` here once that's fixed?

desfero (approved):

Moejoe90 (approved):

Issue #4913, opened 2021-04-29
Align tab bar icons with design


arjunu:
**Description**
There are a couple of difference between the implementation and design of new tab bar icons: 
- the opacity/color of inactive icons and text is different, the design is at full black
- the spacing between icon and text is more in the implementation
- text font weight is 500 in design

Design:
![image](https://user-images.githubusercontent.com/4038217/116537728-e30a9c80-a8e6-11eb-838c-842a54b23afb.png)
Implementation:
iOS:
![image](https://user-images.githubusercontent.com/4038217/116538010-4694ca00-a8e7-11eb-9ef5-33acb032df59.png)
Android:
![image](https://user-images.githubusercontent.com/4038217/116537749-e9007d80-a8e6-11eb-98a2-047f0b305f20.png)

**Environment**
iOS and Android

**Steps To Reproduce**
Go to home screen

**Comments, context, logs**
Original ticket https://github.com/neufund/platform-frontend/issues/4870
PR: https://github.com/Neufund/platform-frontend/pull/4876

PR #4961, opened 2021-05-17
#4913 Align tab bar icons with design

ozanmanav-ng:


packages/wallet/app/router/AppAuthRouterTabs.tsx
```diff
@@ -52,14 +54,25 @@ const tabConfig = [
...
   const { navigationTheme } = useTheme();
+  const insets = useSafeAreaInsets();
+  const tabHeight =
+    insets.bottom === 0 ? BASE_TAB_HEIGHT : BASE_TAB_HEIGHT + insets.bottom - TAB_PADDING;
 
   return (
     <Tab.Navigator
       initialRouteName={EAppRoutes.home}
       tabBarOptions={{
         inactiveTintColor: navigationTheme.colors.text,
```
desfero, line 70: inactive tint color should be a base gray one. We should also specify active tint color to base gray (by default it's base gray but we never know when react-navigation updates styling)

packages/wallet/app/router/AppAuthRouterTabs.tsx
```diff
@@ -52,14 +54,25 @@ const tabConfig = [
...
+        style: {
+          paddingBottom: insets.bottom + TAB_PADDING,
```
desfero, line 72: what that change is fixing exactly? I'm trying to spot the difference but it seems the same

packages/wallet/app/router/AppAuthRouterTabs.tsx
```diff
@@ -68,7 +81,9 @@ const AppAuthRouterTabs: React.FunctionComponent = () => {
           name={name}
           component={component}
           options={{
-            tabBarLabel: ({ color }: TBarProps) => <MenuLabel style={{ color }}>{name}</MenuLabel>,
+            tabBarLabel: ({ color }: TBarProps) => (
+              <MenuLabel style={[styles.label, { color }]}>{name}</MenuLabel>
```
desfero, line 85: Let's use `MenuLabelBold`

packages/wallet/app/router/AppAuthRouterTabs.tsx
```diff
@@ -81,4 +96,11 @@ const AppAuthRouterTabs: React.FunctionComponent = () => {
   );
 };
 
+const styles = StyleSheet.create({
+  label: {
+    fontWeight: "500",
+    letterSpacing: 0.1,
```
desfero, line 102: btw do we need that later spacing? Why we don't use the one that was set by default in typography? If there is some issue with dithering we can try to reduce `letterSpacing to 0.1` but that should be also done on a typography level.

Also we don't need `fontWeight` probably if we use `menuLabelBold`

desfero (changes requested):

packages/wallet/app/router/AppAuthRouterTabs.tsx
ozanmanav-ng: Active tint color looks the same, according to Design. 

image

https://app.zeplin.io/project/5e6f48a43076c61012fd3793/screen/606c646c2493089832b1e969

packages/wallet/app/router/AppAuthRouterTabs.tsx
ozanmanav-ng: You should check with iphone 11 and iphone 7. I used it to bring the text and icon closer together.

packages/wallet/app/router/AppAuthRouterTabs.tsx
ozanmanav-ng: done

packages/wallet/app/router/AppAuthRouterTabs.tsx
```diff
@@ -52,14 +54,25 @@ const tabConfig = [
...
 
   return (
     <Tab.Navigator
       initialRouteName={EAppRoutes.home}
       tabBarOptions={{
         inactiveTintColor: navigationTheme.colors.text,
+        style: {
+          paddingBottom: insets.bottom + TAB_PADDING,
+          paddingTop: TAB_PADDING,
+          height: tabHeight,
```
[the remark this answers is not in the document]
ozanmanav-ng: When I talked to Andrew,  he asked me to use 56px on non-insets devices. That's why we have to use the height. I have refactored as follows. I used the padding to bring the icon and text closer together.
 
image

packages/wallet/app/router/AppAuthRouterTabs.tsx
ozanmanav-ng: I refactored to this:
image

packages/wallet/app/router/AppAuthRouterTabs.tsx
ozanmanav-ng: done

desfero (approved):

KonradFilipiak (approved):

PR #4969, opened 2021-05-19
Wallet clear store on logout

arjunu:


packages/shared-utils/package.json
```diff
@@ -65,7 +65,8 @@
     "react-intl-phraseapp": "^0.2.4",
     "recompose": "^0.30.0",
     "tslib": "^1.10.0",
-    "yup": "^0.32.8"
+    "yup": "^0.32.8",
+    "redux": "^4.0.1"
```
arjunu, line 69: peer dep

packages/shared-utils/src/utils/middlewares/reduxLogoutReset.ts
```diff
@@ -0,0 +1,31 @@
...
+) => (next: StoreEnhancerStoreCreator) => <S, A extends Action = AnyAction>(
+  reducer: Reducer<S, A>,
+  preloadedState: PreloadedState<S> | undefined,
+) => {
+  const enhanceReducer = (state: S | undefined, action: A) => {
+    switch (action.type) {
+      case logoutActionType: {
+        // Need to force cast as `any` due to the way `StoreEnhancer` type is defined
+        // (there is not access to provide specific store type)
+        return reducer(logoutReducer(state as any) as any, action);
```
arjunu, line 20: couldn't get rid of this any easily, we need to tell Typescript that T and S and related but TAppGlobalState is return type of `setupInitModule` and both are inside wallet and web

packages/shared-utils/src/utils/middlewares/reduxLogoutReset.ts
```diff
@@ -0,0 +1,31 @@
+import { Action, AnyAction, PreloadedState, Reducer, StoreEnhancerStoreCreator } from "redux";
```
arjunu, line 1: moved from web with slight modification

desfero (dismissed):

KonradFilipiak (approved):

PR #4979, opened 2021-05-24
Make buttons rounded

ozanmanav-ng:
If we set it as 24px, it is automatically set to half the height. The maximum value is 2 times the height. So it is enough to set nds-border-radius to 24px.


desfero (dismissed):

ozanmanav-ng:
@pavelneu  I've applied this fix here, it looks good for now: https://github.com/facebook/react-native/issues/6480#issuecomment-764311130 Can you review again ?

packages/wallet/app/components/shared/Touchable.tsx
```diff
@@ -25,13 +28,16 @@ const Touchable = React.forwardRef<TouchableHighlight, TExternalProps>(
     if (isRippleSupported) {
       // TODO: There is a bug in react-native were ripple effect do not follow the borderRadius
       // see https://github.com/facebook/react-native/issues/6480
+      // This fix was applied https://github.com/facebook/react-native/issues/6480#issuecomment-764311130
```
ozanmanav-ng, line 31: here

arjunu (approved):

KonradFilipiak (approved):

PR #4997, opened 2021-06-07
Ledger use web USB

arjunu:
Tested on 
- Chrome Version 91.0.4472.106 (Official Build) (x86_64)
- Chrome Version 93.0.4535.0 (Official Build) canary (x86_64)


packages/web/app/lib/web3/ledger-wallet/LedgerConnector.ts
```diff
@@ -83,33 +85,40 @@ export class LedgerWalletConnector {
...
 
   public getMultipleAccounts = async (
     derivationPaths: string[],
   ): Promise<IDerivationPathToAddress> => {
-    const accounts = await Promise.all(
-      derivationPaths.map(path => this.getMultipleAccountsFromHdPath(path, 0, 1)),
-    );
+    const accounts = [];
+    for (let i = 0; i < derivationPaths.length; i += 1) {
+      accounts.push(await this.getMultipleAccountsFromHdPath(derivationPaths[i], 0, 1));
```
arjunu, line 120: it can only perform one operation at a time so we should do it in serial

ozanmanav-ng (approved):
LGTM

packages/web/app/lib/web3/ledger-wallet/LedgerConnector.ts
desfero: let's put a comment about that so we know the reason

desfero (changes requested):

packages/web/app/lib/web3/ledger-wallet/LedgerConnector.ts
arjunu: added

desfero (approved):

